PR #14, opened 2015-12-08
Mods

tjguk:
Small fix-ups. Only two really necessary: __index__ should presumably have been __init__? And keyword was not imported.

The others are slightly cosmetic: ignoring __pycache__ files recursively; and making the package -m runnable (which is at least my preferred approach these days).


ntoll:
Thanks Tim... I had to resolve the .gitignore file because I'd realised I'd got it wrong and decided to user GitHub's standard version for Python. ;-)

ntoll:
Also, if you want... can you add yourself to the AUTHORS list now that you've contributed things! ;-)

tjguk:
Done -- on the same branch

PR #18, opened 2015-12-09
Fix arrow keys and backspace in REPL

### LINKED ISSUE #17: Arrow keys and backspace don't work in REPL
Pull request incoming!

inglesp:
Also add support for the control key.  (Although see note about
portability!)

I've deleted some code that appears to handle the escape key, because I
couldn't work out what it was supposed to do, and when I tried out the
escape key via the microrepl, nothing seemed to happen.

Fixes #17.


ntoll:
Cheers!

Issue #9, opened 2015-12-08
Add the ability to create a new Python script


ntoll:
As a user, when I click the "New" button I expect to be asked for a name for my new script.

If such a script already exists I should be challenged about this and asked to select to back out or overwrite the existing script.

Upon completion I'm presented with an empty editor ready for me to code in. When I click "Save" the newly created script's name is correctly used.

carlosperate:
What is the result when a file editor is already open? Will tabbed files be used?

ntoll:
Fixed by #24

PR #24, opened 2015-12-11
Implement "New" and "Save" behaviour

### LINKED ISSUE #7: Add ability to save the single Python script
As a user, when I click save I expect my code to be saved as a Python script to the file-system.

A decision we need to make is if we just save Python scripts in a sensible default directory (the platform equivalent of `~/MicroPython`) or allow the user to choose the file's location. Currently I prefer the latter since it's less thinking on the part of the user, could be automagically turned into a SCM repository and/or is a sensible convention.

inglesp:
Fixes #7 and #9

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,2 +1,3 @@
 Nicholas H.Tollervey (ntoll@ntoll.org)
 Tim Golden (mail@timgolden.me.uk)
+Peter Inglesby
--- a/mu/editor.py
+++ b/mu/editor.py
@@ -106,6 +106,7 @@ class EditorPane(QsciScintilla):
         super().__init__()
         self.path = path
         self.setText(text)
+        self.setModified(False)
         self.configure()
 
     def configure(self):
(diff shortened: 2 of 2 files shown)


alistair-broomhead:
New works, and load opens up a dialogue, but on hitting save I get:
```
  File "/home/al/git/mu/mu/editor.py", line 328, in save
    if ed.path is None:
AttributeError: 'EditorPane' object has no attribute 'path'
```

alistair-broomhead:
OK so looking into this, I'm not entirely sure we should be relying so heavily on the QScintilla models for this.

It may be better to go for somewhat of an MVW approach, and have the UI update a tab model, and the tab model update the relevant pieces of UI - I feel like this sort of architecture would not only be easier to test and more resilient to change, but is also a really good fit for the Qt event model, so wouldn't be too much of a pain. Using attributes of QObject derived classes on the other hand may feel more direct, but has, in my limited and somewhat dated experience, always lead to a certain amount of pain.

ntoll:
@alistair-broomhead happy for you to refactor what you think would be a better architecture on another branch (this is my first Qt application so I'm just guessing). Now is the time to make such a suggestion/change given the lack of technical debt.

alistair-broomhead:
@ntoll cool, I might just do that, I'd be happy to have a discussion as to what architecture we go with, but I think we should definitely decouple view code from the business logic, or it will be somewhat impossible to test anything individually. Will you be at the project nights meetup this Wednesday by any chance? It would be good to go through some of this face to face. If not I'm sure I'll catch you on IRC.

alistair-broomhead:
I've added issue #30 for when I start on this, we can discuss it further there. As for this PR, I'm not sure that save works, as the EditorPane instance has no 'path' attribute unless it has been loaded. 

While the simple solution is to add a classattribute of `path = None`, it might be better to include the path as either event information in a signal, or as the argument to a direct method call, rather than attempting to get and set it as a property on the editor tab.

ntoll:
I'll try to get to this Wednesday's project night. No guarantees though because it depends on family things. I'm about 90% sure I'll be there though.

WRT path=None. Prompt user for a filename and save it in the default directory (~/MicroPython)

ntoll:
@inglesp can you please fix this a la the problems @alistair-broomhead points out..? Would be good to have this ready for a Wednesday evening refactor session at PyProjectNights. ;-)

ntoll:
Also, @inglesp please add yourself to the AUTHORS file. ;-)

inglesp:
Have fixed that problem and have hopefully not introduced any new ones!

Issue #30, opened 2015-12-14
Decouple view code from business logic


alistair-broomhead:
This should make other tickets easier to work on, especially #4 - adding tests

Up for discussion:
1. Is it worth trying to make this a more functional codebase?
2. Should we be using QtQuick/QML for view code where possible?
3. What architecture? MVC, MTV, MVVM, some N-tier architecture?

I'm tempted on 1. and 2. but I'm aware that this would require buy-in from other contributors, and this is not my project. As for 3. I think choosing something and sticking to it is probably more important than getting this decision perfect. I'm tempted for this reason to go MVC as it's common and I think just about everyone has used it in some form.

alistair-broomhead:
On Wednesday we decided that 1. is worth a try, as is 2., and for 3. we're going with MVC for now. I then promptly tries to do everything at once and ended up with a mess,

Today I've gone back to the drawing board, and I'm trying to do one thing at a time on https://github.com/alistair-broomhead/mu/tree/feature/30-decouple-view-code-from-business-logic, starting with 3. - I'm currently re-factoring 62e193f0636f2678a3c68af16fce821e1a01cc7c to MVC, then I'll add in what's been done since then to get an MVC version of the code-base.

After this I shall attempt to make the models more functional, and then look at porting the view code to QML.

inglesp:
Hi @alistair-broomhead -- when do you think you might have something ready to be reviewed and merged?  I ask, because it's clear you're making some quite big changes on your branch, and the more changes there are on master, the more work you'll have to do to keep up.  As such, if you were going to produce a reviewable pull request by, say, the end of tomorrow, we could hold off merging other things, but we obviously don't want to be doing that forever.  What do you think?

alistair-broomhead:
@inglesp It's possible, but I wouldn't say definitely. There are going to be merge conflicts, someone is going to solve them for each set of changes, and it probably makes sense for those all to be resolved by one person, to avoid anything being missed. What else is likely to need merging?

ntoll:
@alistair-broomhead if you get the PR ready, I'll review and resolve. Would like to get this change in ASAP so I (we) can hack on the app over Xmas hols. ;-)

ntoll:
The initial work for this has been completed.

PR #42, opened 2015-12-21
Start refactoring to separate models from views

alistair-broomhead:
- Everything in the `models` module should be independent of Qt, and is pure business logic
 - Everything in the `views` module should purely consist of Qt view logic, with no business logic
 - The `controller` module should connect the views and models
 - the `hybrid` module should not exist - this is where there is active, working functionality that does not yet obey this pattern


ntoll:
@alistair-broomhead how's it going on this... ready for a review? ;-)

alistair-broomhead:
@ntoll so this doesn't come anywhere near finishing #30, but it's a start - probably worth reviewing, and we can continue re-factoring small bits as we go

ntoll:
OK... I can see what you're doing here. If you don't mind I'll take a look at this tomorrow and will probably finish and/or rename various things so it's obvious what they're for from the PoV of an 11yo.

This is a great start @alistair-broomhead...!

I'll add tests too! :-)

Yay. :christmas_tree: :santa: :snowman: :smiley:

alistair-broomhead:
I'm unlikely to get much time at a laptop today, but I should be available by text or email until this evening, if you want to discuss anything.

mu/views/chrome.py
```diff
@@ -0,0 +1,213 @@
+from PyQt5.QtCore import QSize, Qt, pyqtSignal
```
inglesp: `chrome.py` doesn't feel like a great name to me -- it's too jargonny.  Could it be `window.py` instead?

mu/views/chrome.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+        self.addAction(
+            name="zoom-out",
+            tool_text="Zoom out (to make the text smaller).",
+        )
+        self.addSeparator()
+        self.addAction(name="quit", tool_text="Quit the application.")
+
+
+def tab_widget():
```
inglesp: For consistency with eg the `ButtonBar`, could this be a class?

mu/views/chrome.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+
+        self.setWindowIcon(load_icon(self.icon))
+        self.update_title()
+        self.setMinimumSize(800, 600)
+
+        splash = QSplashScreen(load_pixmap(self.icon))
+        splash.show()
+
+        self.widget = QWidget()
```
inglesp: Is `widget` a good name for this?  I don't know what kind of thing it is or what it does.

mu/views/chrome.py
```diff
@@ -0,0 +1,213 @@
...
+        splitter = QSplitter(Qt.Vertical)
```
inglesp: Should this be a `Splitter`?

mu/views/chrome.py
```diff
@@ -0,0 +1,213 @@
...
+
+        splitter.addWidget(self.tabs)
+
+        self.addWidget(self.widget)
+        self.setCurrentWidget(self.widget)
+
+        self.show()
+        self.autosize_window()
+
+        return (lambda: splash.finish(self)), splitter
```
inglesp: Only the `lambda` is used by the calling code.  Returning the lambda also feels a bit odd -- could we add an `end_splash` method to `Window, which is then called when `run()` is ready:

    def end_splash(self):
        self.splash.finish(self)

mu/views/chrome.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class Splitter(QSplitter):
```
inglesp: This isn't used anywhere.

mu/views/chrome.py
alistair-broomhead: Maybe, or splitting this up closer to 1 file per class

mu/views/chrome.py
alistair-broomhead: I'd actually prefer to move to using factories for Qt classes, rather than subclassing where possible, but haven't had much time to work on this, hence the scatter-gun approach in getting something mergable and reviewable

mu/views/chrome.py
alistair-broomhead: Container maybe? It doesn't directly do anything, it's just the ui hierarchy root node. So I guess root would do too

mu/views/chrome.py
alistair-broomhead: Yes, good catch

mu/views/chrome.py
alistair-broomhead: I'm not sure I'd like this to be part of window's public interface, I'd like to move towards factories, although I guess the lambda could easily become a locally scoped function

mu/views/chrome.py
alistair-broomhead: Part of the drive for moving to factories, is that it makes a good intermediate step for introducing qml

mu/views/chrome.py
alistair-broomhead: As mentioned above, it should be ;)

ntoll:
Hi @alistair-broomhead, I've had a look through the code and the comments above. I agree with @inglesp about the naming. 

Basically, I want kids to understand how the editor works. It's important that we don't get too abstract. I realise this *may* mean we're not doing the most efficient thing, or perhaps the conventional thing, but we must certainly do the *obvious* thing from the perspective of a new programmer.

Does this make sense?

I will personally buy everyone who has contributed to this editor a beer at PyCon UK if we get a pull request from a fellow coder who is 16 or under (whilst obviously excluding said 16yo from beer offer).

:-)

alistair-broomhead:
I think that's all quite sensible, although I'm not sure what you're hinting at in terms of making things less efficient? Is this to do with classes versus factories? Whether ending the splash screen should be a method? Splitting up the chrome module or just renaming it? What the root widget is called? 

To me these are all issues of clarity independent of efficiency; I agree that it's important to get this right, and in my hurry to prepare something reviewable I've probably fallen well short of the standard we'd want to set for the project. I'm happy for all of these to be fluid, although I'd question complicating the view's interface, and also feel that moving from subclasses to factories would probably aid comprehension too.

ntoll:
Merry Christmas @alistair-broomhead :-)

My bad for not being clear. By efficient I meant in how we write/structure the code... sometimes it'd better to just be explicit in, say 5 lines of code, rather than a single (say) list comprehension or other opaque trick that requires more advanced knowledge. I don't mean efficiency in terms of, say big O notation. ;-)

Put simply clarity and obviousness should trump funky code. ;-)

ntoll:
Fixed by my branch stacked upon this on.

ntoll:
For reference: https://github.com/ntoll/mu/pull/48

PR #43, opened 2015-12-23
Add pyinstaller build for travis and appveyor

carlosperate:
Squahed PR for https://github.com/ntoll/mu/pull/29


ntoll:
That's better. ;-)

package/install_osx.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -ev
+brew update
+brew install python3
+sudo pip3 install --upgrade pip setuptools
```
inglesp: Hi @carlosperate, quick question -- what's this for, and why does it need sudo?  (Okay, two quick questions!)

carlosperate: It ensures pip for python 3 and setuptools are installed and updated to the latest version. I had to add sudo to a few commands as they would otherwise fail in the CI server, however there is a chance that this might have been left out from a previous test. I'll make a note to test without sudo on the next update to this shell file.

Issue #47, opened 2016-01-09
Save raises error on close


NDevox:
If you close out of the save dialog (click the x, don't actually save) an error is raised:

    Traceback (most recent call last):
      File "/home/path/to/mu/mu/editor.py", line 356, in save
        with open(ed.path, 'w') as f:
    FileNotFoundError: [Errno 2] No such file or directory: ''

On Ubuntu this just gets handled and the window survives. However this kills mu on windows 10 which will likely hold most of the userbase.

Considering where it is, this could be a very frustrating bug (crashing the program just as you are thinking about saving).

Opening doesn't seem to the same effect on either OS so seems specific to saving.

NDevox:
Built a patch for this. Will commit and push tomorrow.

NDevox:
This was closed by #48

Issue #34, opened 2015-12-16
Saved script should add extension if not given


carlosperate:
Clicking the save button prompts the user to select a location and filename for the Python script. If not extension is included in the filename, mu should automatically attach it.

NDevox:
Built a fix for this (went with #47).

Will push tomorrow. Need to branch a few things....

NDevox:
Pull #54 does this for me on win10 and on ubuntu.

NDevox:
Made updates to #54. Current state of affairs would force files to be a .py (by always adding it if the filename does not end with .py)

PR #52, opened 2016-01-10
Fix for #47 and #34

NDevox:
…as path is empty).

Also adds a '.py' extension if no extension is given.


NDevox:
Didn't realize that Alistairs reqwuest had been merged.

I'll rebase and make another request.

PR #54, opened 2016-01-10
Added default extension of .py when saving

### LINKED ISSUE #34: Saved script should add extension if not given

NDevox:
Fixes #34.

Added so when there is no extension given, it appends .py.

Most commits below are just me rebasing to master.


mu/editor.py
```diff
@@ -189,6 +189,11 @@ def save(self):
             tab.path = self._view.get_save_path(MICROPYTHON_DIRECTORY)
         if tab.path:
             # The user specified a path to a file.
+
+            if '.' not in tab.path.split('/')[-1]:
```
ntoll: Could you do this instead?

```
if not os.path.basename(tab.path).endswith('.py'):
```

Why? Because we'd be using built-in functions designed for this sort of check rather than rolling our own platform specific check (what you have won't work on Windows ~ it uses the path separator `\`).

Will merge once fixed. Great stuff! :smile:

NDevox:
I was quite tempted to do something along those lines, but I'm wondering if anyone would want to save a file as something other than a `.py`.

If someone for some reason wanted to save a .csv or something, they'd end up with `something.csv.py` which is probably not what they want!

Appreciate that this IDE is specific to MicroPython on the MicroBit though so it may be worth forcing it.

In the case that we do force it I think we might want to make this more explicit as well. You can set the save dialog to only look at .py files which is probably worth doing.

Let me know your opinion and I'll work it in tonight.

NDevox:
Added the suggested edits above.

Still a bit wary about forcing things to save as a `.py`

Also added myself to the AUTHORS file.

ntoll:
@NDevox totally understand wariness - in fact I'd like to be able to allow it to work with other file extensions. But for the time being it's a sane default and we can create an issue when it becomes an issue. ;-)

PR #59, opened 2016-01-12
Try to understand why OSX package doesn't run

inglesp:
Don't merge me!!!

### DIFF
--- a/package/extras/qscintilla2.rb
+++ b/package/extras/qscintilla2.rb
@@ -5,7 +5,7 @@ class Qscintilla2 < Formula
   sha256 "9b7b2d7440cc39736bbe937b853506b3bd218af3b79095d4f710cccb0fabe80f"
 
   depends_on :python3
-  depends_on "pyqt5" => "with-python3"
+  depends_on "pyqt5"
 
   def install
     cd "Qt4Qt5" do
--- a/package/install_osx.sh
+++ b/package/install_osx.sh
@@ -2,11 +2,6 @@
 set -ev
 brew update >/dev/null 2>&1  # This produces a lot of output that's not very interesting
 brew install python3
-sudo pip install git+git://github.com/inglesp/pyinstaller.git@e299fcb1d630c65e91019f6e6bff6fa3b87e8bc1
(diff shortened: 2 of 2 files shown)


inglesp:
The problem I'm trying to understand is that when I run the OSX executable that Travis builds, we get the following error:

```
$ ./mu-2016-01-12_09_22_21 
  File "<string>", line 2, in <module>
  File "/Library/Python/2.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 389, in load_module
  File "mu/app.py", line 23, in <module>
ImportError: No module named PyQt5.QtWidgets
run returned -1
```

The debugging output that I've added demonstrates that after homebrew has installed the necessary Qt libraries, Travis is able to import `PyQt5.QtWidgets`, which suggests that PyInstaller's not able to find it -- so I've just tweaked the PyInstaller code we're using to log verbosely, and I'll look at the output tomorrow.

inglesp:
Also, the reason for the failure of b4e96c3 is that the build timed out -- the slowness is caused by homebrew thinking it has to install pyqt5 twice.  :-1:

inglesp:
So the build now runs much more quickly (hurrah!) but the verbose logging produces too many logs, so the build job gets terminated (boo!).

ntoll:
That's progress... I'm currently building a test suite..! 45% test coverage so far this evening. It'll be wonderful when it's done! ;-)

carlosperate:
Don't forget to change the travis deployment to all branches, or the built file will not be uploaded.
Based on the latest run it looks like pyinstaller has added more modules to the file, as it is about 17Mbs now: https://travis-ci.org/ntoll/mu/jobs/102198080#L276
That's a similar size to the first osx test, so there is a chance that it could still have the QSCI module missing.

inglesp:
Hey @carlosperate -- how would I change the travis deployment to all branches?

carlosperate:
I believe you'd have to change line 55 from `branch: [master, travis]` to `all_branches: true`. More info in: https://docs.travis-ci.com/user/deployment/

inglesp:
Closing in favour of https://github.com/ntoll/mu/pull/60.

inglesp:
Thanks @carlosperate -- I think in this case, I'd rather just get #60 merged.  Would you like to review it?

PR #64, opened 2016-01-15
Udev rule for any user to access the device.

kushaldas:
Just the udev rule for now. Will put it in the right place by using the installer.


ntoll:
Great stuff... as I understand it this is a Linux only thing (so this applies equally well to Debian & Fedora derived distros).

Having said that, I didn't need to update udev rules on a newly installed Raspbian in order to make the REPL work. I don't have that much experience with such things so, down the line, I'd like to understand exactly what's going on. ;-)

PR #65, opened 2016-01-15
Initial setup.py, still requires lot of work

kushaldas:


ntoll:
So the reason I didn't do this to begin with is that there's no way to specify the Qt based requirements for this project. Any guidance on this would be most welcome.

PR #66, opened 2016-01-15
Adds icon, & desktop file, and updates setup and MANIFEST for the same

kushaldas:


setup.py
```diff
@@ -23,4 +23,7 @@
             "mu = mu.app.run",
         ],
     },
+    data_files = [('/etc/udev/rules.d', ['conf/90-usb-microbit.rule',]),
+                  ('/usr/share/pixmaps', ['conf/mu.png',]),
```
carlosperate: instead of adding mu.png, can we link to the file in /mu/resources/images/icon.png ?

ntoll: I was wondering too... ;-)

kushaldas: All the applications I know, have their icons as normal files than links. That is why I did the copy.

ntoll: OK... I've chatted to @kushaldas on IRC and it appears that the desktop entries for packaging purposes expect files to be in a certain place and such files may not be links to existing files - hence the need for a completely new file (even though it is a copy of what's found at /mu/resources/images/icon.png). It's a KDE/Gnome related thing afaict.

As a result, I'll merge.

PR #67, opened 2016-01-18
Add all tests to travis OS X CI server

carlosperate:
Adding the test to the OS X build.

Currently the linux version fails with pytest, for some reason it looks like it fails to collect the correct files, and falls when trying to parse unrelated files, so the linux `make check` has been temporarily excluded.
https://travis-ci.org/carlosperate/mu/jobs/103223019#L520-L524
https://travis-ci.org/carlosperate/mu/jobs/103223019#L538-L543

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,14 +31,15 @@ before_install:
 
 # Build Mu and time stamp it
 install:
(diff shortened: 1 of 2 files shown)


inglesp:
Looks sensible, thanks!

PR #69, opened 2016-01-20
Fix new location for get-pip.py script in linux ci install.

carlosperate:
Issue reported in https://github.com/ntoll/mu/issues/4#issuecomment-173309520, thanks @ntoll for catching that right away.


carlosperate:
The continuous-integration/travis-ci/push failure was basically caused by my manual cancellation of the OS X CI server run.

PR #80, opened 2016-02-09
Correct serial configuration order

### LINKED ISSUE #75: REPL doesn't work with automatically built binary on Linux.
What it says. When run from source locally, it works fine.

I'm using vanilla Debian Jessie with the expected QT libraries installed.

carlosperate:
Fixes #75.

Even though the baud rate should be able to be set before opening the port (http://doc.qt.io/qt-5/qserialport.html#baudRate-prop), a few Stack Overflow qserialport answers indicated that it should be set afterwards. Doing this simple change fixes the REPL in the Linux CI-built binaries.

This binary was created on a test branch when debugging this issue, which should now work: https://s3-us-west-2.amazonaws.com/ardublockly-builds/microbit/linux/mu-2016-02-08_23_49_10

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -544,8 +544,8 @@ class REPLPane(QTextEdit):
         # open the serial port
         self.serial = QSerialPort(self)
         self.serial.setPortName(port)
-        self.serial.setBaudRate(115200)
         if self.serial.open(QIODevice.ReadWrite):
+            self.serial.setBaudRate(115200)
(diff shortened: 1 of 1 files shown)


ntoll:
Tested... it works! Awesome sauce! :-)

carlosperate:
@ntoll I've deleted the "repl_fix" branch, which has been lingering around from this PR.

ntoll:
+1

PR #84, opened 2016-02-11
Checks for device serial number to identify micro:bit

kushaldas:
In Fedora we need to check the serial number to identify the
micro:bit device.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -30,6 +30,8 @@ from mu.contrib import uflash, appdirs
 MICROBIT_PID = 516
 #: USB vendor ID.
 MICROBIT_VID = 3368
+#: USB serial number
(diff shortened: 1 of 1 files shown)


ntoll:
Just a thought... is this a serial number for a specific instance of a device (i.e. your micro:bit) or is it the same for *all* micro:bits..?

kushaldas:
Correct, that may be a problem. Can we get the serial numbers from few devices? So that we can create/find the common part.

ntoll:
Unfortunately, it looks like each device has a different serial number. Viz. the two numbers below:

```
9900023431864e450003100e0000003b00000000cc4d28bd
9900023431864e45005910050000005f00000000cc4d28bd
```

The top one is the one you test for, the bottom one is one of my own micro:bits. :-(

Back to the drawing board.

Issue #83, opened 2016-02-10
Make use of a consistent font across all platforms and windows


ntoll:
See: https://www.google.com/fonts/specimen/Inconsolata (It's free)

carlosperate:
Sounds like #83 has implemented this feature and this issue can be closed?

ntoll:
You mean #89 ..?

+1

PR #89, opened 2016-02-15
#83: Use Source Code Pro font for editors and REPL

stestagg:
#83: Include and use the Source Code pro font for the editor and REPL window.

I've included some variants of Source Code Pro (It's SIL licensed so no issues there)
and made the interface load and use that.  Because of some changes in the way the
fonts are specified, I took advantage of the existing Font class to include
some of the loading logic there.

Tests updated etc to match, I've run a build using travis (on my fork) and the OSX binary loads the font fine.
screenshot 2016-02-15 18 12 02


tests/test_interface.py
```diff
@@ -74,12 +55,24 @@ def test_theme_apply_to():
     lexer.setEolFill = mock.MagicMock(return_value=None)
     lexer.setPaper = mock.MagicMock(return_value=None)
     theme.apply_to(lexer)
-    assert lexer.setFont.call_count == 2
+    assert lexer.setFont.call_count == 17
```
stestagg: This used to be 2 because a new font was only added to the editor if/when it was actually different from the base font.  

I've made the assumption here that calling setFont with the same font won't be a measurable performance impact, so just made it call setFont each time.  

Re-implementing the old behaviour wouldn't be too tricky here, if that's better

stestagg:
Windows build not working

ntoll:
I realise this is closed, but just a quick comment. Can we not introduce a dependency on `fin` (no idea what it is)? It only appears to be used to cache fonts, something that could be easily implemented ourselves. Also, since Mu is supposed to be *as simple as possible* and *completely obvious* hiding such functionality isn't in line with this philosophy.

Any moderately skilled programmer should be able to see how we do everything. Caching fonts with a magic module that's only used for this specific reason is opaque.

Otherwise, this looks great! ;-) (Shame it doesn't work for Windows - yet).

stestagg:
Yeah, that's fair enough, I wrote fin so it's my go-to for attribute caches :).

I've worked out the windows problem (funnily enough, requirements weren't being installed, so fin was missing), and have a version without fin (just testing it), singletons/class attributes are always a bit weird in Python, so I woudl welcome suggestions about a pragmatic alternative if you don't like the approach.

stestagg:
Windows now builds fine:

muwin

stestagg:
And linux:
screenshot 2016-02-15 23 16 19

ntoll:
Great stuff, and many thanks! Apologies it's taken so long to review the PR.

PR #99, opened 2016-03-20
Not to be merged! Adding linux and windows make check tests to CI servers.

carlosperate:
Not to be merged, this branch has been set up to investigate the current issues when running `make check` on other platforms than OS X.


carlosperate:
Tests are already working on Appveyor for windows, and tested to work on Travis Linux when Python 3.5 is used (so will be implemented when the Qscintilla wheel issue is solved). Therefore this PR and branch are no longer needed.

PR #100, opened 2016-03-21
Mu quests

Avaren:
Initial ideas about quests


ntoll:
Hi @Avaren, thank you so much for this contribution. The code base for Mu has moved on such a lot since you forked the repos to make this PR and, as a result, this PR would be impossible to merge without significant effort. Perhaps post the 1.0 release (coming soon) we can revisit the possibility of quests for new users..?

PR #102, opened 2016-03-25
add Debian packaging

asb:
This is enough to build a working mu deb, which has been tested on Debian
jessie (Raspbian). To build a .deb yourself, `git-buildpackage
--git-upstream-tree=master -uc -us` should do the trick.

The debian directory was generated using stdeb.
https://wiki.debian.org/Python/Packaging is some help, but it turns out almost
everything it mentions is unnecessary due to the magic of pybuild.


ntoll:
Cheers m'dears... :-)

debian/changelog
```diff
@@ -0,0 +1,5 @@
+mu (0.1-1) unstable; urgency=low
```
carlosperate: Since the version has not been added to the control file, I get this impression that it could be picking it up from here?

Is there any better way to manage versioning and are there any rules about the numbering? I'm thinking of generating automatic builds and would need to manually edit whichever file defines the version in order to append an incremental number (so if mu is at version 0.2 right now, I'd like to generate builds with something like `0.2-<incremental number>a` or anything else to denote it's not "production" code.)

asb: Yes, the version is picked up from debian/changelog. You can look at tools like `dch` for programmatically manipulating this. The rules on version numbers are described [here](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version).

carlosperate: Looks like we'll have to synchronise this with https://github.com/ntoll/mu/blob/master/mu/__init__.py. Looking at https://github.com/ntoll/mu/issues/58 I guess mu is currently being installed as 0.1-1 as listed here.
I'll see if I can have a look tonight at the link you posted.

carlosperate: Thanks for the info @asb. I was able to successfully use git-dch to update the changelog in `debian/changelog` (with and without the dch snapshot flag), but when the deb package gets built, the changelog packed into `\usr\share\doc\mu\changelog.Debian.gz` still contains the original content without the extra version added. So, at packing it prints:

```
dpkg-buildpackage: source package mu
dpkg-buildpackage: source version 0.1-1
...
```

And when installing over a previous version the console also shows the same version:

```
Unpacking mu (0.1-1) over (0.1-1) ...
Setting up mu (0.1-1) ...
...
```

And of course `dpkg-query -l 'mu'` also shows 0.1-1.

These are the commands I am currently using, am I missing something?:
```
git-dch --ignore-branch --since="$PREV_COMMIT"
git-buildpackage --git-upstream-tree="$GIT_COMMIT" --git-ignore-branch --git-ignore-new --git-export-dir="$DEB_BUILD_DIR" --git-force-create -uc -us
```

asb: It's not immediately clear what is going wrong here. Nothing you're doing seems obviously wrong. What does the newly added entry to the changelog look like?

carlosperate: With the snapshot flag I get:

```
mu (0.1-2~1.gbp609e30) UNRELEASED; urgency=medium

  ** SNAPSHOT build @609e30ea818e5e45cede2b7394b6bd8f64b367ad **

  * Add setuptools to package dependencies.

 -- carlosperate <carlosperate@users.noreply.github.com>  Thu, 21 Apr 2016 07:25                          :55 +0000

mu (0.1-1) unstable; urgency=low
...
```

Without the snapshot flag the message is very similar (without any mentions to it being a snapshot), and the version is a simple `mu (0.1-2) UNRELEASED; urgency=medium`.

asb: Sorry, nothing is coming to mind.

carlosperate: No worries, thanks for your help @asb, I'll see if I can find some time to have a look again and try to figure it out.

PR #103, opened 2016-04-17
Add setuptools to .deb package dependencies.

carlosperate:
The `python3-setuptools` needs to be added to the package dependencies as well, otherwise we get:

```
  File "/usr/bin/mu", line 5, in <module>
    from pkg_resources import load_entry_point
```

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,6 +7,6 @@ Standards-Version: 3.9.1
 
 Package: mu
 Architecture: all
-Depends: ${misc:Depends}, ${python3:Depends}, python3-pyqt5, python3-pyqt5.qsci,
-         python3-pyqt5.qtserialport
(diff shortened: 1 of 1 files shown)


ntoll:
Thanks!

ntoll:
@carlosperate hmm... this just broke the build. I'm just about to step into a meeting. Care to take a look? ;-)

carlosperate:
make coverage is giving a `make: *** [coverage] Segmentation fault: 11`, which is quite odd. So I retested an older commit that passed when it was originally pushed (f72bdc7fb9fb21af7ac204f09e76d25b6fdd5582), and it now fails as well: https://travis-ci.org/ntoll/mu/builds/122734337
So it's either a problem with travis, the travis environment, or pulling the newest versions of the dependencies. Will need to investigate further.

carlosperate:
Never mind, looks like we are all green now.

Issue #106, opened 2016-06-02
Clicking 'x' next to sketch name deletes the user's program with no warning


matt-land:
mu is a little too happy to discard the active program without saving. It could say 'are you sure you want to close this file without saving' before discarding the user's program.

matt-land:
PR open to fix this
https://github.com/ntoll/mu/pull/109

ntoll:
Closed by #109 getting merged. Many thanks! :-)

PR #109, opened 2016-06-03
fix for github issue #106, the unsafe tab 'close' button

matt-land:
I had a go at the QT api. This should resolve files being closed without prompting the user to save. It does work correctly with multiple files too.

The styling changes are to fix the dark theme usage to still be high contrast.
screen shot 2016-06-03 at 3 36 41 pm

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -310,6 +310,29 @@ class ButtonBar(QToolBar):
                       self.parentWidget()).activated.connect(handler)
 
 
+class FileTabs(QTabWidget):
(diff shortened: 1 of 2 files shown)


carlosperate:
Don't forget to run `make check` before submission to ensure all tests (including linter) pass :)
Ideally it'd be great if you could include unit test with the PRs, that way the project could maintain 100% coverage.

ntoll:
@carlosperate I'll write the tests... mocking and tests with Qt is painful and I've enough scars to have developed a thick skin. :-)

PR #111, opened 2016-06-13
Configure for pip/setup.py install on Windows

lordmauve:
pip install is broken because no install_requires are given.

setup.py install is broken on Windows because of data_files including Linux-only paths.

Also, the existing install creates a console script (python.exe) not a GUI script (pythonw.exe).

This change fixes all three problems.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,13 @@
+import sys
 from setuptools import setup
 from mu import __version__
 
+
+data_files = []
+if sys.platform == 'linux':
+    data_files = [('/etc/udev/rules.d', ['conf/90-usb-microbit.rules', ]),
+                  ('/usr/share/pixmaps', ['conf/mu.png', ]),
+                  ('/usr/share/applications', ['conf/mu.desktop', ])]
 
 setup(
     name='mu',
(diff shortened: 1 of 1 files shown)


ZanderBrown:
looks good I've been haveing this issue as well but I don't know much about setup tools

ntoll:
I'll test and merge just as soon as I can get hold of a fresh Windows install. ;-)

tjguk:
Annoyingly, this works fine for pip install, but python setup.py install fails to find the PyQt5/QScintilla wheels, although pip install has no problem with them. This is within a Python 3.5 venv. I've cast around a bit but can't see a well-known difference between the two, given that we're importing setup from setuptools.

If I can't see anything I'll ping distutils-sig to see if anyone there can help

tjguk:
Question asked over at distutils-sig:

https://mail.python.org/pipermail/distutils-sig/2017-March/030227.html

tjguk:
Feedback over at distutils-sig points out that setuptools (which is what's being used for setup.py install) won't install wheels, and PyQt5 doesn't publish sdists. (Even if it did, building PyQt is a non-trivial exercise). So basically with the approach here, pip install will work; setup.py install won't.

ZanderBrown:
Looking at master @ntoll has independently incorporated these changes at some point in the last 2 years

ntoll:
Closed since this has been fixed elsewhere. Thanks for the contribution!

PR #114, opened 2016-06-22
Specify pyinstaller 3.1.1 to avoid VC++ redistributable dependency.

### LINKED ISSUE #110: 'missing' msvcr100.dll in Windows 64-bit
Running the binary **mu.exe** under Windows 8.1 64-bit there's an error message saying _"program cannot be started because msvcr100.dll is missing on the computer"_
Solution: there is **Microsoft Visual C++ 2010 Redistributable x86** needed to be installed -- yes: the 32-bit version = x86
DL: [https://www.microsoft.com/en-in/download/details.aspx?id=5555](https://www.microsoft.com/en-in/download/details.aspx?id=5555)

carlosperate:
Tested on a clean Windows 7 32-bit VM. Builds previous to this commit throw the missing dll error, and reverting back to an older version of PyInstaller fixes the issue.

Fixes #110

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,8 +38,8 @@ install:
   - cmd: pyqt5_installer.exe /S
   - cmd: python -c "import PyQt5"
 
-  # Install python dependencies
-  - cmd: pip install pyinstaller
+  # Install python dependencies (v3.2 has VC++ redistributable dependencies)
+  - cmd: pip install pyinstaller==3.1.1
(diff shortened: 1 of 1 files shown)


carlosperate:
Looks like the fix has been confirmed in #110, and at the time of writing version 3.1.1 is still the latest released version without this issue.
@ntoll I'd say this can be safely merged.

PR #130, opened 2016-08-06
Close unused file descriptor in pycodestyle check.

### LINKED ISSUE #117: Code check permission error on Windows
When pressing the check button:

```
  File "C:\workspace\mu_\mu\logic.py", line 532, in check_code
    pep8 = check_pycodestyle(tab.text())
  File "C:\workspace\mu_\mu\logic.py", line 123, in check_pycodestyle
    os.remove(code_filename)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\IEUser\\AppData\\Local\\Temp\\tmpdeeh1sux'

```

Using the resource monitor, it indicates the file is locked by the python process.

Tested on Win10x64 (the built .exe file, fails silently) and Win7x32 (using python 3.4.4 32bit, with the above trace).

carlosperate:
Fixes #117.

In essence, on Windows the temporary file created for the pycodestyle check was not removable as the Python process was keeping it locked.

I've only tested the fix on Windows 10 x64, but it seems like an innocuous enough change to not cause issues in the other platforms (famous last works..).

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -102,7 +102,8 @@ def check_pycodestyle(code):
     """
     # PyCodeStyle reads input from files, so make a temporary file containing
     # the code.
-    _, code_filename = tempfile.mkstemp()
+    code_fd, code_filename = tempfile.mkstemp()
+    os.close(code_fd)
     with open(code_filename, 'w') as code_file:
         code_file.write(code)
     # Configure which PEP8 rules to ignore.


carlosperate:
Btw Nicholas, don't you even dare looking at the PRs before your well deserved holidays!

ntoll:
LGTM :-)

If things go "bang" we know we've knackered something.

PR #134, opened 2016-08-12
Editor.reset_annotations() line_no correction.

carlosperate:
The QsciScintilla Editor member `self.indicators` is set in `Editor.annotate_code()` as 

```python
self.indicators[marker_id] = message
```

In this case `message` is a list of dictionaries, each containing the `line_no` (line count starts at value 1), `column`, and `message`.

On `Editor.reset_annotations()` `self.indicators` dictionary is read as if the keys were the line numbers instead of the marker IDs. This PR corrects the `reset_annotation()` implementation.

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -282,9 +282,11 @@ class EditorPane(QsciScintilla):
         """
         self.clearAnnotations()
         self.markerDeleteAll()
-        for line_no in self.indicators:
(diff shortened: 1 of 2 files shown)


ntoll:
Thanks again! Works for me and my bad for forgetting to update this in the first place once I'd figured out using marker IDs. :-)

PR #135, opened 2016-08-12
Feature/multi-board

eduvik:
Support for serial-only devices, and webrepl devices - implemented by abstracting interactions to the device in upython_device.py. File management not yet implemented. Config is set in config.py. Some conversions between bytes/text need attention, and junk data on the serial line (such as during startup of esp8266) needs to be filtered.

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -19,6 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import keyword
 import os
+import platform
 import logging
 from PyQt5.QtCore import QSize, Qt, pyqtSignal, QIODevice
 from PyQt5.QtWidgets import (QToolBar, QAction, QStackedWidget, QDesktopWidget,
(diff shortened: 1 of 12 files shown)


ntoll:
Lots of good stuff in here... but something appears to have upset Travis. :-/

Any ideas?

eduvik:
Might be because I haven't looked at tests (yet) :-/ 
[5 minutes later]
Have had a look at tests; how does the test runner handle the hardware? Are there tests for different platforms (i.e. macOS, Windows, Linux)?

carlosperate:
At the moment the tests are only run on macOS, on windows we need to configure the environment to correctly run the make file and on linux it throws a weird `*** [coverage] Aborted (core dumped)` when pytests starts collecting items. Should probably look into both these issues sometime soon.

ZanderBrown:
This PR seems to have been polluted with stuff from the unmerged #165

mu/interface.py
```diff
@@ -780,9 +780,10 @@ def keyPressEvent(self, data):
             msg = b'\x1B[C'
         elif key == Qt.Key_Left:
             msg = b'\x1B[D'
-        elif data.modifiers() == Qt.ControlModifier or \
-                (platform.system() == 'Darwin' and
-                    data.modifiers() == Qt.MetaModifier):
+        elif (platform.system() == 'Darwin' and
+                data.modifiers() == Qt.MetaModifier) or \
+             (platform.system() != 'Darwin' and
+                data.modifiers() == Qt.ControlModifier):
```
ZanderBrown: That's starting to get complicated. Is there no way to simplify this?

ZanderBrown:
I think V1 makes this somewhat defunct

Perhaps some of this could be reimplemented on the new design

ntoll:
Hi @eduvik, I'm going to close this PR since the architecture of Mu is completely different from the version you used to fork from. :-/ However, please don't think this means such contributions are unwelcome... they're not! If you have time, perhaps once 1.0 is out the door (soon) we can revisit this code..?

Issue #128, opened 2016-08-03
files/REPL dialog box needs to be a modal dialog


peterhinch:
The dialog box which pops up if you click on files when the REPL is open (and vice versa) does not behave like a proper modal dialog: if you click on the main window it disappears behind the main window and the application appears to be unresponsive. For a beginner-focussed app this is confusing: the dialog box should remain in the foreground.

peterhinch:
The problem is that a modal dialog needs a parent window otherwise it is "owned" by Qt rather thn by a specific window, So here's how to fix it (docstrings omitted for clarity)
```python
    def show_message(self, message, information=None, icon=None, parent=None):
        message_box = QMessageBox(parent)
```
and
```python
    def show_confirmation(self, message, information=None, icon=None, parent=None):
        message_box = QMessageBox(parent)
```
Then in each call provide
```python
            self._view.show_message(message, information, parent = self._view)
```

ntoll:
Hi @peterhinch 
Great stuff!
Care to turn this into a PR? Then your contribution will be immortalised forever! :-)

peterhinch:
I would if I weren't well out of my depth with GitHub. I'm actually working on a fork targeted at the ESP8266 running MicroPython. My [fork](https://github.com/peterhinch/mu) is forked from [eduvik's](https://github.com/eduvik/mu) which is in turn forked from your code. While I can raise PR's against eduvik I'm lost as to how to do it against your code.

carlosperate:
When you create a pull request from your repository page it will most likely show you a comparison between your selected branch and eduvik's master. If you click on that drop down menu you should be able to select any other repository, including ntoll's mu master.

peterhinch:
OK, I've raised a PR. I hope I've got it right - I find Python a whole lot easier than Git ;-)

ZanderBrown:
travis isn't happy but otherwise everything appears to be in order. It does help when you reference the pull (#138) and you appear to have renamed `find_microbit()` to `find_upython_device()` and your pull doesn't reference this

peterhinch:
This came from eduvik's fork. I clearly failed to disentangle my changes from his.

ZanderBrown:
ah. Well would you mind if I was to setup a pull containing only your changes?

peterhinch:
No problem.

carlosperate:
This issue has been fixed in #141. Thanks for the contributions!

PR #138, opened 2016-08-16
Dialog boxes are now modal

peterhinch:
Formerly if the user clicked on the main window when a dialog box was present, the dialog became hidden behind the main window. To a naive user it appeared as if the application had crashed. This patch ensures dialogs remain visible.

### DIFF
--- a/mu/contrib/microfs.py
+++ b/mu/contrib/microfs.py
@@ -41,13 +41,13 @@ For example, 'ufs ls' will list the files on a connected BBC micro:bit.
 """
 
 
-def find_microbit():
+def find_upython_device():
     """
     Finds the port to which the device is connected.
     """
     ports = list_serial_ports()
     for port in ports:
-        if "VID:PID=0D28:0204" in port[2].upper():
+        #if "VID:PID=0D28:0204" in port[2].upper():
             return port[0]
     return None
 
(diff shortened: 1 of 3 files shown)


ZanderBrown:
this is for issue #128

mu/contrib/microfs.py
```diff
@@ -41,13 +41,13 @@
 """
 
 
-def find_microbit():
+def find_upython_device():
```
carlosperate: Looks like this might be a left over from the esp8266 fork.

carlosperate:
I was thinking, `self._view` is in instance of the `Window` class and the `QMessageBox`
is created within the method `Window.show_message()`. Instead of having to indicate a parent widget in the logic portion of the application, could we instead just pass `self` as the parent for `QMessageBox`? As we want the main window to always be the parent for anything opened by `Window.show_message()`.

mu/contrib/microfs.py
peterhinch: You're right. I'm afraid I've got hopelessly confused with Git/Github.

ZanderBrown:
i don't see why not

ntoll:
Hi!

Thanks for the contribution. I notice Travis has had a fit of the collywobbles. Care to revise? This is a valuable contribution!

:-)

N.

ZanderBrown:
As mentioned in #128 I was going to be putting together a cleaner PR on @peterhinch 's behalf although I might modify taking into account @carlosperate 's comment. I will admit I forgot but it's top of my TODO for tomorrow morning.

ntoll:
@ZanderBrown cheers. Make sure you reference this PR as the thing being replaced. ;-)

carlosperate:
As this fix has been already merged in #141 we can close this PR.

PR #139, opened 2016-08-22
Separate code analysers to have different editor indicators

carlosperate:
This PR separates pyflakes and pep8 indicators to have different editor decorations, in this case using different underlining colours.

virtualbox_xubuntu14 04-64-installed_22_08_2016_00_57_23

Did this a while ago based on some feedback from the mailing list. It looks like I didn't really have time to finish the unit test, which have been updated to pass, but not to test the new stuff properly. I kind of forgot about this until I saw it lurking around in my branches, so I though before I do any more work on it I should submit it to ask for some feedback, as I am not sure if this is still a desired feature or if the approach is appropriate.

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -140,7 +140,8 @@ class DayTheme(Theme):
     Paper = QColor('white')
     Caret = QColor('black')
     Margin = QColor('#EEE')
-    Indicator = QColor('red')
+    IndicatorError = QColor('red')
+    IndicatorStyle = QColor('blue')
 
 
 class NightTheme(Theme):
@@ -166,7 +167,8 @@ class NightTheme(Theme):
     Paper = QColor('black')
     Caret = QColor('white')
     Margin = QColor('#333')
-    Indicator = QColor('white')
+    IndicatorError = QColor('white')
+    IndicatorStyle = QColor('cyan')
 
 
 class PythonLexer(QsciLexerPython):
@@ -201,8 +203,10 @@ class EditorPane(QsciScintilla):
         super().__init__()
         self.path = path
         self.setText(text)
-        self.indicators = {}
-        self.INDICATOR_NUMBER = 19  # arbitrary
+        self.indicators = {  # IDs are arbitrary
+            'error': {'id': 19, 'markers': {}},
+            'style': {'id': 20, 'markers': {}}
+        }
         self.MARKER_NUMBER = 22  # also arbitrary
         self.api = api if api else []
         self.setModified(False)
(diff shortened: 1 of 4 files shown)


ZanderBrown:
any chance of using tooltips to display messages?

carlosperate:
you mean tooltips instead of the clickable markers on the left?

ZanderBrown:
yes because underline makes it appear as if a misspelling in a Word Processor where you would normally right-click or hover (some form of contextual action). The markers are useful to identify lines with issues but it took me a while to realise they were clickable.

carlosperate:
Yes, that would be possible as well, I believe only on click though, but I am not too familiar with the Scintilla built-in capabilities.

carlosperate:
Okay, so I was briefly playing with this feature to add the annotations when the underlined text is clicked and there is one small problem.

Some of the annotations are general to the line instead of just to a specific column, and the side marker groups all of them into a single annotation, while the text click would only display that specific issue. This could lead to missed information for lines that contain a "general line" and "specific column" annotations.

So in this example:

```python
not code:

bad_style={}

# Not ending with last line
```

Clicking on the equal sign produces:

text_clicked

And clicking on the side marker:

marker_clicked

So taking in consideration the tendency to click the underlined word, it is quite easy to miss the "Expected and indented block" message.

ntoll:
Very pertty! :-)

carlosperate:
Any thoughts on implementing the "click on underlined code" to display the annotation? I agree with @ZanderBrown that I didn't instinctively click on the arrows right away, it feels more natural to do so on the underlined text (although my first instinct was to right click, which I don't think is supported out-of-the-box), but then it might obfuscate the clickable nature of the side markers even more.

ntoll:
So the way I originally implemented it was to follow the marking style many teachers use for prose: they'll underline something (often in different colours depending on the sort of annotation it is - which is why this is such a cool PR) and then write something more detailed in the margin. As a result, I assume this will be familiar to kids and teachers.

However, I guess we won't know until people start using the feature, provide feedback and we can judge for ourselves.

To be clear: I'm not precious about my code or implementation. If evidence or further enhancements lead to a solution that's obviously better for our target users (beginner programmers with only basic experience of computer use) then we should use it.

carlosperate:
That sounds like a perfectly good expectation, after all we are not the target audience for Mu and they could very well react to this implementation as it is currently assumed.

I think that for me personally, the main reason why it did not occur to me to click on the marker right away was because an arrow usually points to something, rather than provide additional content. Perhaps if there was something like a :heavy_plus_sign: sign I would have discovered its functionality sooner? Scintilla has the `SC_MARK_PLUS` / `self.Plus` ([available options](http://www.scintilla.org/Markers.png)), but to be honest I am not completely convinced it's better. As you've mentioned, this is the type of thing were ideally we can get a better idea if we could collect some feedback.

ZanderBrown:
I can see where you got the idea from just not sure how well it translates to a code editor.

PR #140, opened 2016-08-22
(WIP) Search for a settings file in the app directory.

carlosperate:
Taking in consideration the discussions from this week about the default location for the "mu_code" folder, and based on the fact that we'll probably add an entry in the settings to change it (
https://github.com/ntoll/mu/issues/126, https://github.com/ntoll/mu/issues/86, https://github.com/ntoll/mu/issues/87) I was thinking that having to find the settings file it's not all that straightforward to begin with.

So, this PR updates Mu to first search for the settings file in the same directory where the Mu executable or run.py file are located.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -46,8 +46,6 @@ PYTHON_DIRECTORY = os.path.join(HOME_DIRECTORY, 'python')
 DATA_DIR = appdirs.user_data_dir('mu', 'python')
 #: The default directory for application logs.
 LOG_DIR = appdirs.user_log_dir('mu', 'python')
-#: The path to the JSON file containing application settings.
-SETTINGS_FILE = os.path.join(DATA_DIR, 'settings.json')
 #: The path to the log file for the application.
 LOG_FILE = os.path.join(LOG_DIR, 'mu.log')
 #: Regex to match pycodestyle (PEP8) output.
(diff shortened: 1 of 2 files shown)


carlosperate:
Forgot to add a test for a malformed JSON to exercise the exception from `Editor.restore_session()`:

```python
try:
    old_session = json.load(f)
except ValueError:
    logger.error('Settings file {} could not be parsed.'.format(
                 settings_path))
```

Apart from that, the pyinstaller builds should probably be tested a little bit more thoroughly. So after I add that test I'll merge it to a different branch in this repository to kick the build servers and create downloadable test builds.

ntoll:
Hmm... could this not just be resolved with appropriate use of appdirs..? (asking naively).

Viz.:

https://github.com/ActiveState/appdirs

ZanderBrown:
How useful it will be in this context I don't know but that is a useful library. definitely going to use in my projects

carlosperate:
Depends on what you mean by "resolving this".

This PR wasn't made to change the default directory to save Mu code files, as discussed in the linked issues, or where to **_save_** the `setting.json` file.

Instead I was thinking about the use-case discussed specifically in https://github.com/ntoll/mu/issues/126. For that solution we had proposed to add a new entry in the `setting.json` file for the default "Mu code files" directory, therefore allowing users to change it to whatever they prefer, simply by editing this json file.
The only problem I was thinking about, is that the file itself is saved in the `DATA_DIR` directory, which is OS specific, so we would have to instruct users on how to find this specific directory (also taking in consideration the fact that we are dealing with hidden directories and locations that can be different in different versions of the same OS).

So what this PR is proposing is to allow Mu to first check if there is a `setting.json` present in the same directory where Mu is located, and use that instead of the default OS specific `DATA_DIR` location. If there is no file right next to the Mu executable or `run.py` file, then it would just fall back to the original behaviour.

This way we could instruct users to create a new file with this entry and just save it together with Mu.

carlosperate:
By the way, this definitely needs more testing with the Pyinstaller bundled app, specifically in macOS, which is why I was proposing to merge this PR into another branch from this repository, triggering the travis build server to produce the executable and see if it works. (I followed the documentation, but I am not 100% sure if it will return the mu.app directory instead of the `mu.app/Contents/MacOS/mu` one).

ntoll:
@carlosperate ok... I see what you mean and it sounds sensible enough.

carlosperate:
Please do let me know if you think there is no need to implement this specific feature or if perhaps we should stop for a minute to think about a different approach?
I will be on holidays for a couple of weeks from tomorrow, so I won't be able to work on this anyway, but happy to discuss it on the meantime.

Or maybe this is something we could look into during the PyCon UK sprint :)

ntoll:
I believe the PyCon UK sprint is going to be *very* productive. I'm going to draw up a "plan" that everyone can ignore, but it'll be a useful starting point so people can see what we may need to tackle. ;-)

carlosperate:
Okay, so in order to test the pyinstaller builds in the three OS I will merge this into a different branch in this repository, which should trigger the CI builds.

PR #141, opened 2016-08-24
Modal dialogs!

ZanderBrown:
Take 2 of #138

Simplest PR imaginable it appears

All seems to work fine however the 'OK' button in dialogues seems to be losing it's styling but they are certainly modal


ntoll:
Great stuff! :-)

PR #145, opened 2016-08-28
FIX: Highlight Code Errors Under Windows

jezdean:
The kids are getting indentation errors and it's quite difficult to debug. I wanted to add whitespace characters similar to Sublime:
sublime

The best I could do in mu is:
in mu

I can't seem to make the characters larger or visible only when highlighted. [I couldn't get tabs to show as arrows either.](http://pyqt.sourceforge.net/Docs/QScintilla2/classQsciScintilla.html#ab83469cc9550eadcd5d4e8ca3d20d07b)

If anyone has ideas :+1:

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -152,7 +152,7 @@ def check_pycodestyle(code):
     os.remove(code_filename)
     # Parse the output from the tool into a dictionary of structured data.
     style_feedback = {}
-    for result in results.split(os.linesep):
+    for result in results.split('\n'):
         matcher = STYLE_REGEX.match(result)
         if matcher:
             line_no, col, msg = matcher.groups()


mu/interface.py
```diff
@@ -236,6 +236,7 @@ def configure(self):
         self.setMarginSensitivity(1, True)
         self.marginClicked.connect(self.on_marker_clicked)
         self.setAnnotationDisplay(self.AnnotationBoxed)
+        self.setWhitespaceVisibility(self.WsVisible)
```
ntoll: where is `self.WsVisible` set..?

ntoll:
@jezdean! That's a really helpful contribution. It's based on actual evidence and includes a patch! :-)

I looked into this right at the very beginning (adding whitespace characters via SCiTE). You found the setting that I did and I believe what you see is the full extent of the feature that's available. There's also (I believe, if memory serves me correctly) indentation guides (i.e. lines showing the vertical alignment of code). I'll have a play around and post some screenies.

Also, I believe this feature should be optional. When I worked as a teacher I spent a lot of time with kids whose literacy skills were underdeveloped - one of the things my colleagues and I tried to avoid, to help such learners just get to the words, was to remove all extraneous typography from the page. Imagine, for example, how all the dots and lines in a text editor would interfere with a learner's ability to read the text.

jezdean:
I agree on the dots. But even as an option, I'm not sure how worthwhile they are because you can't see 'um! The character (interpunct?) is not in the .OTF file so I didn't know how to make it bigger.

The problem I was trying to resolve is debugging indentation errors. The kids have to go through their whole code trying to find missing/extra spaces. 

I think I worked it out. The solution is already in mu but it doesn't work on Windows. The indentation errors were not showing up on Windows when you click 'check'. It only finds the first PyCodeStyle error. Replacing `for result in results.split(os.linesep):` with `for result in results.split('\n')` in logic.py finds all the errors for me.

I've only tested on my Windows 10 computer.

changes-side-by-side2016-08-30 18_12_56-mu

ZanderBrown:
That makes sense because `for result in results.split(os.linesep):` assumes the file was created on the current os. Reading a 'Windows' file on Linux would work because `os.linesep` would have a value of `\n` so would match `\n` as well as `\r\n` where as the windows `\r\n` will only match windows.

In a side note Travis is unhappy about something

jezdean:
@ZanderBrown  That's right but the output of `checker.check_all()` isn't read from a file; it's read from `stdout`.

The Travis is unhappy about source forge. yay!

carlosperate:
Indeed, I've introduced this bug on the PR to fix the Windows line endings. As the output from the checker is not leaving the "python realm", line endings in this case will always stay as `\n`.
The current fix should be valid, I will restart the Travis worker to see what it might be complaining about.

carlosperate:
After the travis restart yesterday it seems to be happy now. I'll see if I can test this in linux and os x as well just to ensure it works as expected, but the change looks good to me.

carlosperate:
I've done some development in Linux and Windows with this fix applied and everything seems in order, haven't been able to do the same in OS X, but in theory it should behave the same. I'd say this is good to merge, although perhaps the title should be changed to reflect the final fix.

PR #156, opened 2016-09-19
Fix #126 Change name of default Python directory

### LINKED ISSUE #126: Mu creates directories in my home
Every time I start Mu, it creates a directory called "python" in my home directory. Not only is this extremely annoying, but it also took me some time to figure out which application is responsible for this, as the name is completely unrelated to Mu.

If it needs a configuration directory, I would suggest using ``~/.config/mu`` by default (or even better, follow the xdg standard for configuration directories).

tibs:
Change the name of the default Python (save/load) directory from ~/python to ~/mu_code, which fixes #126

Also:

- Add a changelog, CHANGES.rst
- Update the development documentation (in README.rst and docs/development.rst) to explain how to install PyQt5 and QScintilla on Mac OS
- Add me to the AUTHORS file

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,3 +4,4 @@ Peter Inglesby
 Carlos Pereira Atencio (carlosperate@embeddedlog.com)
 Nick Sarbicki (nick.a.sarbicki@gmail.com)
 Kushal Das (mail@kushaldas.in)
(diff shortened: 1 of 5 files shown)


carlosperate:
Travis was complaining about some segmentation fault during the tests (this is now becoming more common... if it keeps happening we'll have to investigate), but restarting the job has fixed it, it's all green now.

PR #157, opened 2016-09-19
Use a settings file local to the Mu executable.

carlosperate:
This is a continuation from #140, which was merged into a non-master branch to trigger the CI builds to create the executable files for all Operating Systems.

As a quick recap from #140:

> So what this PR is proposing is to allow Mu to first check if there is a setting.json present in the same directory where Mu is located, and use that instead of the default OS specific DATA_DIR location. If there is no file right next to the Mu executable or run.py file, then it would just fall back to the original behaviour.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ deploy:
   acl: public_read
   on:
     repo: ntoll/mu
-    branch: [master, travis]
+    branch: [master, travis, local_settings]
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ deploy:
   acl: public_read
   on:
     repo: mu-editor/mu
-    branch: [master, travis, local_settings]
+    branch: [master]
 
 notifications:
   email:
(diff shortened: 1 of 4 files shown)


carlosperate:
@tibs would you be so kind to test this OS X executable for me?
https://s3-us-west-2.amazonaws.com/ardublockly-builds/microbit/osx/mu-2016-09-19_14_25_06.zip

I'd be great if you could first run it normally to ensure it uses your default settings, and then create a `settings.json` file in the same directory as the .app application with the following content:

```
{
  "paths": [],
  "theme": "night"
}
```

So it should be quite easy to check if Mu changes the theme when this file is present.

carlosperate:
The executables have been tested successfully in Windows and Linux, but there seems to be some issues on OS X. I still need to investigate that before this could be considered for merge.

ntoll:
I can check on OSX (my wife's Macbook) if this is useful?

carlosperate:
If you could also post the the Mu log file then yes :-) I assume the issue is with the reported `sys.executable` by PyInstaller on OS X, and I've added that to the log.
I'll have access to a mac on Friday, which is when I will most likely be able to to work on this anyway, so it's not urgent either.

carlosperate:
Had to add an additional check for the .app package in macOS, as the executable path is different in that case. It all works now, but I still need to update the unit tests for the feature to be complete.

carlosperate:
I've updated the unit tests so we are back to full coverage, minus the temporary import for pep8.

I must admit unit tests are not my strong suit, so any comments on how to improve them will be greatly welcomed.

Issue #152, opened 2016-09-19
Enforce < 8k size for Python scripts


ntoll:
If a Python script it > 8k in size it corrupts things in the runtime. This is not good (tm). Ensure that Python scripts written in Mu cannot be > 8k in size.

carlosperate:
Implemented in #158.

PR #158, opened 2016-09-19
Check if script > 8k

ZanderBrown:
Okay so completely untested & quite possibly over simplified but here it is.

Check if the script is greater than 8k before flashing. Hopefully closing #152 

A workaround this 8k limit maybe to attempt to strip out comments in code before flashing?

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -314,6 +314,12 @@ class Editor:
         python_script = tab.text().encode('utf-8')
         logger.debug('Python script:')
         logger.debug(python_script)
+        if len(python_script) >= ‪8192:
+            message = 'Unable to flash "{}"'.format(tab.label)
+            information = ("Your script is too long!.")
+            self._view.show_message(message, information, 'Warning')
+            return
+        
         # Generate a hex file.
         python_hex = uflash.hexlify(python_script)
         logger.debug('Python hex:')
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -319,7 +319,6 @@ class Editor:
             information = ("Your script is too long!.")
             self._view.show_message(message, information, 'Warning')
             return
-        
         # Generate a hex file.
         python_hex = uflash.hexlify(python_script)
         logger.debug('Python hex:')
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -314,7 +314,7 @@ class Editor:
         python_script = tab.text().encode('utf-8')
         logger.debug('Python script:')
         logger.debug(python_script)
-        if len(python_script) >= ‪8192:
+        if len(python_script) >= 8192:
             message = 'Unable to flash "{}"'.format(tab.label)
             information = ("Your script is too long!.")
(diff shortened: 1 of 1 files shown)


ZanderBrown:
This should probably be implemented in uflash as well

carlosperate:
There is a style error:

    mu/logic.py:322:1: W293 blank line contains whitespace
    make: *** [pycodestyle] Error 1

At the moment only the OS X server is running the tests.

ZanderBrown:
hopefully that'll clear it up

ntoll:
Great stuff! I'd love to merge this but Travis is complaining and it appears to be some sort of configuration problem..? @carlosperate..?

carlosperate:
It like it's a "code issue" reported by pyflakes, although I cannot see the offending problem on the code itself, maybe an invisible characters?
https://travis-ci.org/mu-editor/mu/jobs/161134091#L313

@ZanderBrown could you try to run the `make check` on your machine to see if you can replicate locally?

carlosperate:
Actually, if I copy the offending line into an editor with ANSI or UTF-8 enconding it pastes the following:

```
if len(python_script) >= ?8192:
```

So there is a character there before the `8192`.

ZanderBrown:
hopefully all invisible characters are gone

ntoll:
Woop woop. :-)

carlosperate:
>Build mu 1.0.277 failed

>Commit 0ffbbbeaee by Zander on 9/20/2016 3:08 PM:
>Travis: i dare you to reject this

https://ci.appveyor.com/project/carlosperate/mu/build/1.0.277

Looks like you angered the other CI god  :scream:

No worries though, it a pyqt5_installer.exe download timeout, which happens sometimes, it should be fine the next time it tries.

ZanderBrown:
seems to be fine now.

Possibly investigate https://liftoff.github.io/pyminifier/ or similar to avoid anyone seeing this message of doom? Alternatively fix the runtime 😄

carlosperate:
Nothing to do with python or the application, just appveyor failing to fetch a file online. I'll increase the timeout a little bit anyway, that should give it a bit more breath room to download it.
If I have the time I'll look into caching the file to preserve it between builds.

ZanderBrown:
Looking at why >8k is a problem it would appear the cause would appear to be quite simple:

The Micro:Bit has 16k memory
The MicroPython runtime is now > 7k

7k + 8k is very close to 16k so it appears that the problem here is that we a running out of memory. Not a problem you get very often these days.

So I get the feeling we are going to need to revisit this with every update to the runtime.

ntoll:
Upstream MicroPython are addressing the limited amount of RAM. In this specific instance I believe the problem is that >8k causes the hex of the Python script to leak over other parts of the hex thus causing problems.

carlosperate:
This is something we could calculate in uFlash and expose the available flash memory to mu, which would be better than having that magic number .

ZanderBrown:
that would simplify. as I said this was only a quick fix

PR #164, opened 2016-10-02
Editor text selection searches and highlights matches. (WIP)

carlosperate:
So this week at work I had to work with an IDE without this feature, and out of sheer frustration I though I'd give this a go in Mu.

mu_select_find gif

I personally find it very useful, but I don't know if this might fall outside of the original goals of the project, so feel free to reject it if that's the case (I only really worked on this as a way to vent).

If you think it should go in then I should update the unit test to cover the new functionality.

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -19,6 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import keyword
 import os
+import re
 import logging
 from PyQt5.QtCore import QSize, Qt, pyqtSignal, QIODevice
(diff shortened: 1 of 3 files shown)


ZanderBrown:
Whether or not it should be in normal Mu a "MegaMu" should have this

ntoll:
@carlosperate I really like it. It's very simple, visually obvious and rather useful..! :-) That it doesn't involve yet more buttons is a bonus. :-)

ntoll:
Hi @carlosperate I've added some tests to get coverage back to 100% for the changes you've made. :-)

carlosperate:
Awesome, thanks @ntoll!

PR #166, opened 2016-10-06
File overwrite fix

eduvik:
Lets try this again

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -850,6 +850,19 @@ class MuFileList(QListWidget):
         self.setAcceptDrops(True)
         sibling.setAcceptDrops(True)
 
+    def show_confirm_overwrite_dialog(self):
+        msg = QMessageBox()
+        msg.setIcon(QMessageBox.Information)
+        msg.setText("File already exists; overwrite it?")
+        msg.setWindowTitle("File already exists")
+        msg.setStandardButtons(QMessageBox.Ok | QMessageBox.Cancel)
+        retval = msg.exec_()
+        if retval == QMessageBox.Ok:
+            return True
+        else:
+            return False
+
+
 
 class MicrobitFileList(MuFileList):
     """
@@ -865,17 +878,20 @@ class MicrobitFileList(MuFileList):
         source = event.source()
         self.disable(source)
         if isinstance(source, LocalFileList):
-            local_filename = os.path.join(self.home,
-                                          source.currentItem().text())
-            logger.info("Putting {}".format(local_filename))
-            try:
-                with microfs.get_serial() as serial:
-                    logger.info(serial.port)
-                    microfs.put(serial, local_filename)
-                super().dropEvent(event)
-            except Exception as ex:
-                logger.error(ex)
+            if self.findItems(source.currentItem().text(), Qt.MatchExactly) \
+                     and self.show_confirm_overwrite_dialog():
+                local_filename = os.path.join(self.home,
+                                              source.currentItem().text())
+                logger.info("Putting {}".format(local_filename))
+                try:
+                    with microfs.get_serial() as serial:
+                        logger.info(serial.port)
+                        microfs.put(serial, local_filename)
+                    super().dropEvent(event)
+                except Exception as ex:
+                    logger.error(ex)
         self.enable(source)
+        self.parent().ls()
 
     def contextMenuEvent(self, event):
         menu = QMenu(self)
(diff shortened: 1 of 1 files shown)


ZanderBrown:
Looks good. Let's see what the great Travis has to say...

mu/interface.py
```diff
@@ -865,17 +878,20 @@ def dropEvent(self, event):
...
-                                          source.currentItem().text())
-            logger.info("Putting {}".format(local_filename))
-            try:
-                with microfs.get_serial() as serial:
-                    logger.info(serial.port)
-                    microfs.put(serial, local_filename)
-                super().dropEvent(event)
-            except Exception as ex:
-                logger.error(ex)
+            if self.findItems(source.currentItem().text(), Qt.MatchExactly) \
```
carlosperate: I haven't tested this build, but if we are dragging a new file, woudn't then `self.findItems()` evaluate False and not transfer the file? (same applies to `LocalFileList.contextMenuEvent()`)

eduvik: Good catch; have fixed this.

mu/interface.py
```diff
@@ -850,6 +850,19 @@ def enable(self, sibling):
         self.setAcceptDrops(True)
         sibling.setAcceptDrops(True)
 
+    def show_confirm_overwrite_dialog(self):
+        msg = QMessageBox()
+        msg.setIcon(QMessageBox.Information)
+        msg.setText("File already exists; overwrite it?")
+        msg.setWindowTitle("File already exists")
+        msg.setStandardButtons(QMessageBox.Ok | QMessageBox.Cancel)
```
eduvik: I'd be inclined to leave OK as default; at least for computer to device transfers, as you're probably going to want to overwrite files on the device with a new one pretty often. Going the other way might be better to have cancel as default though

mu/interface.py
```diff
@@ -911,19 +928,24 @@ def dropEvent(self, event):
...
-                with microfs.get_serial() as serial:
-                    logger.info(serial.port)
-                    microfs.get(serial, microbit_filename, local_filename)
-                super().dropEvent(event)
-            except Exception as ex:
-                logger.error(ex)
+            file_exists = self.findItems(source.currentItem().text(),
+                                         Qt.MatchExactly)
+            if not file_exists or \
+                    file_exists and self.show_confirm_overwrite_dialog():
```
carlosperate: this would be the same as:
```python
if not file_exists or self.show_confirm_overwrite_dialog():
```

mu/interface.py
```diff
@@ -865,17 +877,22 @@ def dropEvent(self, event):
...
+                logger.info("Putting {}".format(local_filename))
+                try:
+                    with microfs.get_serial() as serial:
+                        logger.info(serial.port)
+                        microfs.put(serial, local_filename)
+                    super().dropEvent(event)
+                except Exception as ex:
+                    logger.error(ex)
         self.enable(source)
+        self.parent().ls()
```
eduvik: The ls() call is failing in the tests as the test object doesn't have a parent. There could be another (probably nicer) way to update the file lists, but this should just work. I think I'll just check if there is a parent, and ignore if there isn't

carlosperate: Yeah, I think this area probably deserves some consideration for future refactor, as the interface probably shouldn't be doing any of the microfs logic to `ls()` populate the `MicrobitFileList` list. And since `ls` is a method from `FileSystemPane`, now `LocalFileList` and `MicrobitFileList` would depend on being created inside a PyQt widget with an `ls()` method.

For the moment we could have a mandatory argument for `parent` in `MuFileList` (like `FileSystemPane` on any PyQt widget really) and then mock that for the tests. Seems a bit cleaner than checking if it has a parent here, although nothing stops a developer from creating the instance with a `parent=None` argument and having this crash here anyway, just like they could have an inherent parent without the `ls()` method and it would crash here as well.

mu/interface.py
carlosperate: I can see why that would be more productive, but it's probably better to stay constant on default behaviour and have all message boxes with the same, no?

mu/interface.py
```diff
@@ -850,6 +850,19 @@ def enable(self, sibling):
         self.setAcceptDrops(True)
         sibling.setAcceptDrops(True)
 
+    def show_confirm_overwrite_dialog(self):
+        msg = QMessageBox()
```
carlosperate: Not quite sure, to be honest I never tried to replicate the behaviour myself.

ZanderBrown: #128 was an odd one and I believe this issue would reemerge here. passing `self` to `QMessageBox()` should avoid it.

mu/interface.py
```diff
@@ -850,6 +850,19 @@ def enable(self, sibling):
...
+        msg.setIcon(QMessageBox.Information)
+        msg.setText("File already exists; overwrite it?")
+        msg.setWindowTitle("File already exists")
+        msg.setStandardButtons(QMessageBox.Ok | QMessageBox.Cancel)
+        retval = msg.exec_()
+        if retval == QMessageBox.Ok:
+            return True
+        else:
+            return False
```
ntoll: why not just do:

`return retval == QMessageBox.Ok`

???

ntoll:
Hi, 

I can't accept this as it stands since there are no tests with it. I'll write some tests right now, fix up some of the code to make it simpler and then merge. 

I want this in! It's a cool feature (thanks)! I'm also fed up with people asking when the next release of Mu will be so I want to get something out today.

Thanks for your patience..!

Issue #94, opened 2016-03-03
REPL text getting damaged on re-edit


whaleygeek:
If I type something into the REPL, press return to run it, then up arrow and edit it, and press return again to re-run the edited version - everything past where I press return gets deleted from the REPL window.

e.g.

> > > display.scroll("Hello nicholas")
> > > [up arrow and cursor to the n]
> > > display.scroll("Hello n^icholas")
> > > [change n to N]
> > > display.scroll("Hello N^icholas")
> > > [press return]
> > > display.scroll("Hello N
> > > [I see the above in the REPL window]

whaleygeek:
This looks to be the same as #82

ntoll:
Just checked... this is fixed in master. Closing.

ntoll:
Thanks for the bug report, sorry it has taken so long to reply! :-)

DavidWhaleMEF:
Thanks @ntoll !!

PR #167, opened 2016-10-09
Fix Ctrl keys to work on Windows and Linux; add VT100 cursor control in REPL

eduvik:
Not tested on microbit, but equivalent code from my esp8266 fork works fine.

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -33,7 +33,6 @@ from PyQt5.QtSerialPort import QSerialPort
 from mu.contrib import microfs
 from mu.resources import load_icon, load_stylesheet, load_font_data
 
-
 #: The default font size.
 DEFAULT_FONT_SIZE = 14
 #: All editor windows use the same font
@@ -46,7 +45,6 @@ FONT_VARIANTS = ("Bold", "BoldIt", "It", "Regular", "Semibold", "SemiboldIt")
 NIGHT_STYLE = load_stylesheet('night.css')
 #: DAY_STYLE is a light conventional theme.
 DAY_STYLE = load_stylesheet('day.css')
-
 
 logger = logging.getLogger(__name__)
 
@@ -742,6 +740,7 @@ class REPLPane(QTextEdit):
         self.setFont(Font().load())
         self.setAcceptRichText(False)
         self.setReadOnly(False)
+        self.setUndoRedoEnabled(False)
         self.setObjectName('replpane')
         # open the serial port
         self.serial = QSerialPort(self)
@@ -791,9 +790,10 @@ class REPLPane(QTextEdit):
             msg = b'\x1B[C'
         elif key == Qt.Key_Left:
             msg = b'\x1B[D'
-        elif data.modifiers() == Qt.ControlModifier or \
-                (platform.system() == 'Darwin' and
-                    data.modifiers() == Qt.MetaModifier):
+        elif (platform.system() == 'Darwin' and
+                data.modifiers() == Qt.MetaModifier) or \
+             (platform.system() != 'Darwin' and
+                data.modifiers() == Qt.ControlModifier):
             # Handle the Control key. On OSX/macOS/Darwin (python calls this
             # platform Darwin), this is handled by Qt.MetaModifier. Other
             # platforms (Linux, Windows) call this Qt.ControlModifier. Go
@@ -801,6 +801,15 @@ class REPLPane(QTextEdit):
             if Qt.Key_A <= key <= Qt.Key_Z:
                 # The microbit treats an input of \x01 as Ctrl+A, etc.
                 msg = bytes([1 + key - Qt.Key_A])
+        elif (data.modifiers == Qt.ControlModifier | Qt.ShiftModifier) or \
+                (platform.system() == 'Darwin' and
+                    data.modifiers() == Qt.ControlModifier):
+            # Command-key on Mac, Ctrl-Shift on Win/Lin
+            if key == Qt.Key_C:
+                self.copy()
+            elif key == Qt.Key_V:
+                self.paste()
+
         self.serial.write(msg)
 
     def process_bytes(self, data):
@@ -821,7 +830,7 @@ class REPLPane(QTextEdit):
                 self.setTextCursor(tc)
             elif data[i] == 13:  # \r
                 pass
-            elif data[i] == 27 and data[i+1] == 91:  # VT100 cursor: <Esc>[
+            elif data[i] == 27 and data[i + 1] == 91:  # VT100 cursor: <Esc>[
                 i += 2  # move index to after the [
                 m = re.search(r'(?P<count>[\d]*)(?P<action>[ABCDK])',
                               data[i:].decode('utf-8'))
@@ -875,6 +884,7 @@ class MuFileList(QListWidget):
     """
     Contains shared methods for the two types of file listing used in Mu.
     """
+
     def disable(self, sibling):
         """
         Stops interaction with the list widgets.
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -741,6 +741,7 @@ class REPLPane(QTextEdit):
         self.setAcceptRichText(False)
         self.setReadOnly(False)
         self.setUndoRedoEnabled(False)
+        self.setContextMenuPolicy(Qt.PreventContextMenu)
         self.setObjectName('replpane')
         # open the serial port
         self.serial = QSerialPort(self)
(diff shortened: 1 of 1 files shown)


ZanderBrown:
Travis: image

mu/interface.py
```diff
@@ -809,16 +813,55 @@ def process_bytes(self, bs):
...
+
+        i = 0
+        while i < len(data):
+            if data[i] == 8:  # \b
                 tc.movePosition(QTextCursor.Left)
                 self.setTextCursor(tc)
-            elif b == 13:  # \r
+            elif data[i] == 13:  # \r
                 pass
+            elif data[i] == 27 and data[i+1] == 91:  # VT100 cursor: <Esc>[
```
ZanderBrown: From the great Travis: `E226 missing whitespace around arithmetic operator`

I'm assuming it wants you to use `data[i + 1]`  in place of `data[i+1]`

eduvik: Presumably; even though it's not specified in PEP8 :-/ 

On a side note, it strikes me that PEP8 errors in travis are detracting from 'real' errors - I get that code style is important, but a 'warning' rather than 'error' approach (if that's possible) would be nicer IMHO. Also, there are occasions where breaking PEP8 makes code more readable (see https://www.python.org/dev/peps/pep-0008/#id15 )

mu/interface.py
```diff
@@ -821,7 +830,7 @@ def process_bytes(self, data):
                 self.setTextCursor(tc)
             elif data[i] == 13:  # \r
                 pass
-            elif data[i] == 27 and data[i+1] == 91:  # VT100 cursor: <Esc>[
+            elif data[i] == 27 and data[i + 1] == 91:  # VT100 cursor: <Esc>[
```
ZanderBrown: Tad ridiculous but Travis is happy now.

ZanderBrown:
looks like a better solution

carlosperate:
It would be easier to review and merge individual PRs for individual features. This one on itself is not a very large PR, so it's an issue, but it would keep things nice and tidy for other larger PRs if they have multiple features.

mu/interface.py
```diff
@@ -755,6 +760,30 @@ def __init__(self, port, theme='day', parent=None):
             raise IOError("Cannot connect to device on port {}".format(port))
         self.set_theme(theme)
 
+    def repl_paste(self):
+        """
+        Grabs clipboard contents then sends down the serial port
```
carlosperate: Apologies for I'm being really pedantic here, but it'd be nice to have these comments end with a perior `.`

mu/interface.py
```diff
@@ -755,6 +760,30 @@ def __init__(self, port, theme='day', parent=None):
...
+        """
+        clipboard = QApplication.clipboard()
+        self.serial.write(bytes(clipboard.text(), 'utf8'))
```
carlosperate: Will this throw an exception if the clipboard contains something other than text? If it returns something like `None`, or an empty string, does how does the micropython process that?

mu/interface.py
```diff
@@ -755,6 +760,30 @@ def __init__(self, port, theme='day', parent=None):
...
+    def repl_context_menu(self):
+        """"
+        Creates custom context menu with just copy and paste
+        """
+        menu = QMenu(self)
+        menu.addAction("Copy", self.copy)
+        menu.addAction("Paste", self.repl_paste)
+        menu.exec_(QCursor.pos())
+
+    def cursor_to_end(self):
```
carlosperate: Is this method called anywhere?

mu/interface.py
```diff
@@ -809,16 +850,55 @@ def process_bytes(self, bs):
...
                 pass
+            elif data[i] == 27 and data[i + 1] == 91:  # VT100 cursor: <Esc>[
+                i += 2  # move index to after the [
+                m = re.search(r'(?P<count>[\d]*)(?P<action>[ABCDK])',
+                              data[i:].decode('utf-8'))
+
+                # move to (almost) after control seq (will ++ at end of loop)
+                i += m.end() - 1
+
+                if m.group("count") == '':
```
carlosperate: If you prefer, this could be a bit more pythonic with

```python
count = 1 if not m.group("count") else int(m.group("count"))
```

mu/interface.py
carlosperate: If it gets called when the user clicks or changes the cursor, would this method move the cursor to the end of the selected line for all lines, would that be correct?
Does that affect text selection in case the user might want to copy some data from the repl?

mu/interface.py
```diff
@@ -809,16 +850,55 @@ def process_bytes(self, bs):
...
+                    count = 1
+                else:
+                    count = int(m.group("count"))
+
+                if m.group("action") == "A":  # up
```
carlosperate: I don't really know much about the python internals, but I was wondering if this `m.group("action")` result would get cached or if it performs the operation at every single `if` check in this code block?

mu/interface.py
```diff
@@ -809,16 +850,55 @@ def process_bytes(self, bs):
...
+                elif m.group("action") == "B":  # down
+                    tc.movePosition(QTextCursor.Down, n=count)
+                    self.setTextCursor(tc)
+                elif m.group("action") == "C":  # right
+                    tc.movePosition(QTextCursor.Right, n=count)
+                    self.setTextCursor(tc)
+                elif m.group("action") == "D":  # left
+                    tc.movePosition(QTextCursor.Left, n=count)
+                    self.setTextCursor(tc)
+                elif m.group("action") == "K":  # delete things
```
carlosperate: Is `K` only used to delete the entire line? or could `m.group("count")` be indicating the number of characters to delete?

eduvik: via http://www.termsys.demon.co.uk/vtansi.htm
- Erase End of Line	`<ESC>[K`  Erases from the current cursor position to the end of the current line. 
- Erase Start of Line	`<ESC>[1K` Erases from the current cursor position to the start of the current line.
- Erase Line		`<ESC>[2K` Erases the entire current line.

eduvik:
@carlosperate - I was originally going to put copy/paste fixes in a separate branch/PR, but then decided that it's closely tied to the control-keys so put it in here.

mu/interface.py
carlosperate: So, technically `m.group("count")` could be `""`, `1`, or `2`; so should they be added for completion?
Alternatively, is there a key to trigger this VT100 command with the keyboard? If not, is there any practicality to have this command?

mu/interface.py
eduvik: clipboard.text() casts the clipboard contents to text, so it works for non-text objects. Empty string is fine (just writes nothing); None will break, so fixed in latest commit

mu/interface.py
eduvik: Looks like the microbit port doesn't assume VT100 support (unlike stock micropython, which does): https://github.com/bbcmicrobit/micropython/blob/a20fbe82c919cea9519ce5a1140103f518b51caa/inc/microbit/mphalport.h#L38

eduvik: There's not really many implemented in micropython: on the receiving side (i.e. device receiving) it just implements A,B,C,D,H and F while device sending implements D and K (am unsure if it echoes back the ones it receives):

https://github.com/micropython/micropython/blob/b9672bcbe8dd29b61326af8eb026df4d10a8f0ce/lib/mp-readline/readline.c

ntoll:
@eduvik nice work. I see Travis has the collywobbles... ;-) I also agree with @carlosperate that having features separated into different PRs/branches would be good.

Any chance you could re-submit these as separate PRs etc...? :-)

ntoll:
Folks, these changes are valuable but the PR is in a borked state (Travis related collywobbles).

@eduvik if you don't mind, I'll take this PR and fix things up so it's good to merge and your contributions remain. I'll probably get round to this at the weekend.

ntoll:
Merged in b5550afc14368b0723a72467862c4f35c89e0355.

(Fixed merge conflicts, updated stylistic suggestions form reviews and added tests so we're back at 100% coverage.)

Many thanks.

whaleygeek:
@ntoll @carlosperate how do I find out from github if this merge actually made it into a real release of MicroPython and of Mu, and which version do I have to download to get this fix? I'd like to close down ticket #94 but I am unsure if this fix is deployed yet or not, and how widely deployed it is.

carlosperate:
Hi David,
The Mu releases are also marked using the GitHub releases feature, which add a git tag that can be used to identify which features have or haven't made it. There is also the [CHANGES.rst](https://github.com/mu-editor/mu/blob/master/CHANGES.rst) file, that should contain a general description of what's new.

Issue #178, opened 2016-10-29
Better compatibility for restricted environments in schools


rpiuser:
Would there be any possibility of setting a default save / load path under a windows environment ? Currently C: is restricted and My Documents is redirected using GPO. When loading a "This operation has been cancelled due to restrictions in effect on this computer" message is displayed (Due to C: being restricted) clicking okay takes you to "My Documents" The same behaviour happens with save. Ideally pointing to the users home folder (e.g H:\Mu) would be a lot cleaner and a much better experience / less confusing for the students using it.

ZanderBrown:
Take a look at #126 & #86

rpiuser:
Thank you, I've read through the threads and can't seem to determine if this has been resolved. I'm assuming I need the settings.json and I need to set a path in there, however I'm unsure of the format of the path  :- "paths": ["path/foo.py" so would I use "paths": ["path/H:/mu"] ? I'll try testing.

ZanderBrown:
Ah yes seems that didn't happen. I'm assuming your using the exe?

rpiuser:
Using this build mu-2016-10-23_16_42_53.exe on windows

ZanderBrown:
Okay. Having received some support for the feature in #179 hopefully it should make it into master over the next week to be included in the next release

rpiuser:
Thank you, much appreciated

ZanderBrown:
#179 has been merged. Does it resolve your issue?

carlosperate:
As a general note to remind us before we close this issue: We really need to add a page about the settings file to the documentation so that this feature can be discovered.

rpiuser:
Using mu-0.9.13.win works great thank you, however if you specify a wrong drive letter or the drive doesn't exist in settings.json mu exits with a run returned -1

ZanderBrown:
ah makes sense. I will put a check in asap

rpiuser:
Other than that, it's working great - kids are merrily coding away using it :)

ntoll:
Closing since fixed.

PR #179, opened 2016-10-29
Custom code directory

ZanderBrown:
I'm not especially expecting this to be merged and I've probably upset Travis but this should help out people like @rpiuser in #178 

Basically reads `python_dir` in `settings.json` to use as the save directory


ntoll:
This looks useful! I'd like to merge this and release:

```
./mu/logic.py:117:1: E302 expected 2 blank lines, found 1
./mu/logic.py:128:30: E126 continuation line over-indented for hanging indent
./mu/logic.py:131:1: E302 expected 2 blank lines, found 1
```

ntoll:
Also, I notice that it breaks the unit tests. :sob: 

Care to fix this up? It looks a useful feature for advanced users.

ZanderBrown:
I'll fix it up tomorrow sorry I didn't see you reply earlier!

ZanderBrown:
Anyone with a better understanding of the testing system have any pointers here?

ntoll:
@ZanderBrown Pointers in what sense..?

ZanderBrown:
What I've broken would be great :-)

I've cleared up most the errors now but the last 4 have me stuck,

For example as far as i can see I've done nothing to upset this.
```
>       assert mock_open.call_count == 1
E       assert 2 == 1
E        +  where 2 = <MagicMock id='1917013104'>.call_count

tests/test_logic.py:1130: AssertionError
```

ntoll:
@ZanderBrown OK... I can't look *right now* but will get to it by the end of this afternoon. I'd like to see this released ASAP.

ZanderBrown:
Ah i may have cracked it. `mock_open.call_count` counts how many files have been opened. As a result of my changes settings.json is actually opened twice during these tests

ZanderBrown:
Assuming Travis likes that this should be good to merge

Summery:

* Add `python_dir` key to `settings.json`
* Update `test_logic.py` to allow for the changes in logic
* Add me to `AUTHORS`

Notes

* No tests for this feature specifically (yet?)

ZanderBrown:
Odd the tests run fine on my machine.

carlosperate:
About the error, I'm on a train now, so I could only have a look later tonight, but if it fails on the server but not on your computer make sure you check the paths are correct for all operating systems, as there are going to be diffencies there (test currently only run on macOS).

ntoll:
@ZanderBrown use `make check` as the build chain does. It checks more than just the tests (pep8, flake etc).

Here's what Travis says:

```
E               FileNotFoundError: [Errno 2] No such file or directory: '/Users/travis/Library/Application Support/mu/settings.json'
```

This looks to me like your tests are checking state of the file system. The tests should run in isolation so the set up and tear down for each individual tests needs. Also, the tests shouldn't interfere with nor depend upon the user's own settings files.

Does this make sense?

ZanderBrown:
Yes it does make sense although i'm getting errors for test i haven't edited!

I've has a realisation though and switched
```
        if not os.path.exists(get_python_dir()):
            logger.debug('Creating directory: {}'.format(get_python_dir()))
            os.makedirs(get_python_dir())
        if not os.path.exists(DATA_DIR):
            logger.debug('Creating directory: {}'.format(DATA_DIR))
            os.makedirs(DATA_DIR)
```
for
```
        if not os.path.exists(DATA_DIR):
            logger.debug('Creating directory: {}'.format(DATA_DIR))
            os.makedirs(DATA_DIR)
        if not os.path.exists(get_python_dir()):
            logger.debug('Creating directory: {}'.format(get_python_dir()))
            os.makedirs(get_python_dir())
```
On the basis get_python_dir() reads from DATA_DIR so it must exist first

This is getting quite annoying but of course highlights why these tests exist

ZanderBrown:
**Yay** *green tick!*

I seem to have sorted all the issues raised by the great Travis

mu/logic.py
```diff
@@ -112,12 +111,45 @@ def get_settings_path():
...
+    settings_path = get_settings_path()
+    try:
+        with open(settings_path) as f:
+            try:
+                # Load up the JSON
+                sets = json.load(f)
+                if 'python_dir' in sets:
+                    # They have set a custom folder return that
+                    return sets['python_dir']
+                else:
```
carlosperate: This else statement is not really necessary, as it will just return at the end of the function.

mu/logic.py
```diff
@@ -631,7 +663,8 @@ def quit(self, *args, **kwargs):
                 paths.append(widget.path)
         session = {
             'theme': self.theme,
-            'paths': paths
+            'paths': paths,
+            'python_dir': get_python_dir()
```
carlosperate: Could we use a different name for this? if I am reading the settings file and I see "python_dir" I would assume it would need to point to the python interpreter.
If we do change this, then I guess `get_python_dir()` should also be renamed as well.

mu/logic.py
```diff
@@ -112,12 +111,45 @@ def get_settings_path():
...
+    settings file to be used to set a custom path
+    """
+    settings_path = get_settings_path()
+    try:
+        with open(settings_path) as f:
+            try:
+                # Load up the JSON
+                sets = json.load(f)
+                if 'python_dir' in sets:
+                    # They have set a custom folder return that
```
carlosperate: Sorry about being pedantic about a comment, but technically since we are writing `python_dir` into the settings file then finding this key would be expected in most cases, so it doesn't really indicate a custom path has been set.

mu/logic.py
```diff
@@ -45,9 +45,8 @@
 MICROBIT_VID = 3368
 #: The user's home directory.
 HOME_DIRECTORY = os.path.expanduser('~')
-#: The default directory for Python scripts. This needs to be in the user's
-#  home directory, and visible (so not a . directory)
-PYTHON_DIRECTORY = os.path.join(HOME_DIRECTORY, 'mu_code')
+# Name of the directory within the home folder to use by default
+DIRECTORY_NAME = 'mu_code'
```
ZanderBrown: suggestions?

mu/logic.py
ZanderBrown: very true

mu/logic.py
carlosperate: `MU_CODE_DIRECTORY`?

ZanderBrown:
Any suggestions for tests / how they would work?

mu/logic.py
```diff
@@ -112,12 +111,42 @@ def get_settings_path():
...
+                    logger.debug('Creating settings file: {}'.format(
+                                 settings_dir))
+                    json.dump({}, f)
+            except FileNotFoundError:
+                logger.error('Unable to create settings file: {}'.format(
+                             settings_dir))
     return settings_dir
 
 
+def get_workspace_dir():
```
ZanderBrown: Admittedly tests for this would be good.
Any suggestions of what & how?

ntoll: Briefly (I have dad duty things taking up my time this evening, will reply more fully tomorrow AM), yes: ensure you set up the test with a settings file somewhere temporary and Mock away get_settings_path to return a path to this new temp settings file. Then test to your hearts' content. Does this make sense..?

ZanderBrown: makes sense. unfortunately i may not be able to work on this until tomorrow PM

ZanderBrown:
That was one nail bitingly long Travis build :smile:

tests/test_logic.py
```diff
@@ -152,6 +152,19 @@ def test_get_settings_no_files():
     assert mock_json_dump.call_count == 1
 
 
+def test_get_workspace():
+    """
+    Normally return generated folder otherwise user value
+    """
+    should_be = os.path.join(mu.logic.HOME_DIRECTORY,
+                             mu.logic.WORKSPACE_NAME)
+    assert mu.logic.get_workspace_dir() == should_be
```
carlosperate: Keep in mind that as it is, executing `get_workspace_dir()` will execute `get_settings_path()` which will create the settings.json file on the default location. Even if we don't care about creating this file, if a different test has also done this (in which case it should probably be changed), then we might be reading settings data that might be different than expected.
So we should probably mock `get_settings_path` and the file open to read an empty json object.

ZanderBrown: okay changes incoming

ZanderBrown:
Anything else or is this ready now?

ntoll:
I'm at "real work" *right now* but hope to be able to spend some time just before tonight's @ldnpydojo to get this merged in. :-)

ntoll:
I don't get 100% test coverage. :-(

Don't worry, I've coded up a test. :-)

I'll check back later to ensure the builds have succeeded against new master and will update with a new release.

Issue #187, opened 2016-11-14
Windows packaging instruction issues


tdicola:
I tried to follow the instructions to build a packaged executable of Mu on Windows from here https://github.com/mu-editor/mu/blob/master/package/README.rst but ran into a lot of trouble.  To help folks that hit similar issues, here are the main problems:

-   You probably want to clarify that most people should use 32-bit python.  If you use 64-bit python you'll build a version of mu that only works on 64-bit Windows whereas 32-bit apps will run on both 32-bit and 64-bit windows.  From your appveyor config for making official mu packages it appears you're using 32-bit python.

-   Don't install python 3.5 or higher, you want python 3.4.x.  The reason for this is that the PyQt5 binary installer from Riverbank software doesn't exist anymore for the latest PyQT versions.  Instead Riverbank seems to want people to install with pip, but pip installed PyQT binaries seem to be incompatible with packaging releases using pyinstaller (I tried and it just fails to find all the PyQT DLL files to copy into the package).  The reason you want Python 3.4.x is that you want to use PyQt 5.4.2 binaries which are only available for Python 3.4.x builds.  Find this installer here (again this is what the appveyor config seems to use): https://sourceforge.net/projects/pyqt/files/PyQt5/PyQt-5.4.2/

- Don't pip install pyinstaller, instead install exactly version 3.1.1 using `pip3 install pyinstaller==3.1.1`.  Again the appveyor config mentions and does this, the issue seems to be a new requirement on VC++ redistributables in later pyinstaller versions (I ran into this issue too, I think it needs VC++ 2010 redistributables if anyone has issues).

Once you have all those pieces lined up building a Windows package seems to work fine.  In general anyone attempting to do this should really just ignore the packaging instructions and look at the appveyor config in the project root--this appears to be the source of truth for how to package mu on windows.  IMHO it's worth mentioning in the packaging readme for the Windows instructions.

ntoll:
@tdicola thanks for this!n @carlosperate (our packaging supremo) any thoughts on the above..? The lack of / incompleteness of PyQT binaries for Python 3.5 is concerning.

I take your point about the README.

carlosperate:
Hi @tdicola,

Thanks for pointing this out, you are correct in all your points, we've left the package README file a little bit behind. I'll look at #188 , but yeah we should include all this in the instructions.

[...]

PR #188, opened 2016-11-14
Update instructions in light of feedback.

ntoll:
@carlosperate can you check if this is OK..? :-)

@tdicola this should address #187

### DIFF
--- a/package/README.rst
+++ b/package/README.rst
@@ -7,11 +7,18 @@ Install dependencies
 Windows
 +++++++
 
-* Install Python 3 from the `official website <https://www.python.org/downloads/>`_.
-* Install the PyQt5 binary package from `riverbankcomputing <https://riverbankcomputing.com/software/pyqt/download>`_.
+Please ensure you use the specified versions in the instructions below. We've
+found the latest versions of some of these packages are not always backwards
+compatible with each other.
+
+* Install the 32bit version of Python 3.4 from the `official website <https://www.python.org/downloads/>`_.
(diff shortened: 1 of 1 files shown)


package/README.rst
```diff
@@ -7,11 +7,18 @@ Install dependencies
...
 +++++++
 
-* Install Python 3 from the `official website <https://www.python.org/downloads/>`_.
-* Install the PyQt5 binary package from `riverbankcomputing <https://riverbankcomputing.com/software/pyqt/download>`_.
+Please ensure you use the specified versions in the instructions below. We've
+found the latest versions of some of these packages are not always backwards
+compatible with each other.
+
+* Install the 32bit version of Python 3.4 from the `official website <https://www.python.org/downloads/>`_.
+* Install the PyQt5 binary package from `riverbankcomputing <https://riverbankcomputing.com/software/pyqt/download>`_ (specifically, `this one <https://sourceforge.net/projects/pyqt/files/PyQt5/PyQt-5.4.2/>`_).
```
carlosperate: We should probably link to this page instead: https://riverbankcomputing.com/software/pyqt/download5

carlosperate:
I have updated the Appveyor script to use Python 3.5.2 and the PyQt 5.6 installer and everything seems to be working all right (I've only opened it and played around with it, still need to test it with a microbit, so I'll do that tonight at home):
https://s3-us-west-2.amazonaws.com/ardublockly-builds/microbit/windows/mu-2016-11-16_12_53_41.exe

So we can update the README to use these versions, and perhaps add a note that older versions of the PyQt installer should work with their respective targeted Python versions (the installers are named something along the lines of `PyQt5-x.x-gpl-Pyx.x-Qt5.x.x-x32.exe`).

I still need to check if we can solve the dll packaging issues when installed using pip wheels.

carlosperate:
This is a good reminder that `pip install pyinstaller==3.1.1` is probably not necessary anymore, as the problem was with version 3.2, and 3.2.1 was released in January.

Issue #120, opened 2016-07-19
requirements.txt misses some required packages


kzar:
I found I also needed to install `QScintilla` and `PyQt5` before mu would run. Perhaps those should be included in requirements.txt?

inglesp:
Hi @kzar -- are `PyQt5` and `QScintilla` pip-installable now?  Your issue suggests they might be.  They weren't when we first worked on packaging mu last year.

kzar:
Yep I think so, well I was able to install them with Pip anyway. (Pip for Python 3 as provided in Debian stretch.)

inglesp:
That's quite exciting, and it might make our whole packaging process a bit more streamlined.  I'm testing this out now.

willingc:
Interested in the results of #121. We use Qt in the Jupyter Qt Console (https://qtconsole.readthedocs.io/en/latest/installation.html). 

@takluyver, do you have additional insights on Qt and pip?

takluyver:
PyQt5 is now pip installable for Python 3.5, so we could express a dependency on it for the Qt console. I think the tricky point for us is that we depend on one of PyQt5 *or* PyQt4 *or* PySide, and there's no great mechanism to specify that, even if wheels were available in all cases.

carlosperate:
>  I think the tricky point for us is that we depend on one of PyQt5 or PyQt4 or PySide, and there's no great mechanism to specify that, even if wheels were available in all cases.

This is also a problem when trying to package using PyInstaller, as the mechanism used to detect available libraries uses `__import__` and that doesn't work with PyInstaller, so the workaround can be quite ugly: https://github.com/ntoll/mu/commit/c3bd2a8ec2a26798c2f1643d078b25ad0f1b59b8#diff-5ff4c2f00727c388514182dd8570e7d1R33

takluyver:
All the freeze tools' dependency detection machinery is terrible, in my experience. Glyph has spoken and written about this as well. Application packaging should use an explicit list of requirements, as I do in Pynsist.

carlosperate:
In this case is not about PyInstaller's ability to detect dependencies, but whatever mechanism they use to import their packaged modules does not seem to work when calling `__import__` directly. `qt_loaders.has_binding()` uses this method of importing to check if an specific module is available, so it fails during runtime even if the module has been packaged with the application.

Btw @takluyver, do you have any links about Glyph's articles about this? Should definitely make for an interesting read.

takluyver:
[...]

PR #189, opened 2016-11-18
Move the CI macOS build to Py3.5 PyQt5 from PyPi.

carlosperate:
Since PyQt has moved to PyPI that seems to be the way forward, so we need to start moving all the builds to the system. It's not a very straightforward move as there is pretty much issues in one way or another for all platforms. But since we can get this one out, we should, it'll be a step towards addressing https://github.com/mu-editor/mu/issues/120 and be able to have all dependencies in the requirements.txt file.

This first PR is just for the OS X builds, it will produce builds using Python 3.5 and PyQt5.7 from PyPI like https://s3-us-west-2.amazonaws.com/ardublockly-builds/microbit/tests/osx/mu-2016-11-18_16_11_12.zip which I have just tested and seems to be working perfectly fine.

It also compartmentalises a bit more the OS X and Linux travis environments, giving us slightly more control (for instance could move to the docker linux images, significantly reducing the time to get the tests feedback).

Not included here are updates to the Linux nor Windows builds. There is an issue with the Qscintilla wheel for Linux, so we'll have to wait. I am hopeful that once that is out it should work just fine, with the added benefit that moving to Python 3.5 seems to fix the inexplicable segmentation that is raised when the travis tests attempt to start.


ntoll:
So all the tests have passed. I'll merge. Thank you!

PR #190, opened 2016-11-20
Multi board for Adafruit

ntoll:


ZanderBrown:
We are going to need a system to translate 'Micro:Bit' to whatever is actually attached

ntoll:
Exactly... this is but a first step. I'm going to try to find the time this week to turn the roadmap and associated comments into issues people can discuss pick up so we can gauge levels of interest / importance for each of the things. Multi-board is one aspect of this exercise.

carlosperate:
Since the microfs module basically runs python code on the repl, I assume the "files" functionality also works in all the other micropython boards "out of the box". But just wondering, has anybody had the chance to test it just in case?

PR #191, opened 2016-11-20
Update Appveyor script to run tests in CI server.

carlosperate:
This PR updates the Appveyor script and Makefile to be able to run the tests on the Windows CI server. With that I had to do a minor update to a test that was failing on Windows.

I've also updated the PyQt version to the latest available for Python 3.4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,9 @@ clean:
 	rm -rf mu.egg-info
 	rm -rf .coverage
 	rm -rf docs/_build
-	find . \( -name '*.py[co]' -o -name dropin.cache \) -print0 | $(XARGS) rm
-	find . \( -name '*.bak' -o -name dropin.cache \) -print0 | $(XARGS) rm
-	find . \( -name '*.tgz' -o -name dropin.cache \) -print0 | $(XARGS) rm
+	find . \( -name '*.py[co]' -o -name dropin.cache \) -delete
+	find . \( -name '*.bak' -o -name dropin.cache \) -delete
+	find . \( -name '*.tgz' -o -name dropin.cache \) -delete
 
 pyflakes:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -17,9 +17,9 @@ clean:
 	rm -rf mu.egg-info
 	rm -rf .coverage
 	rm -rf docs/_build
-	find . \( -name '*.py[co]' -o -name dropin.cache \) -print0 | $(XARGS) rm
-	find . \( -name '*.bak' -o -name dropin.cache \) -print0 | $(XARGS) rm
-	find . \( -name '*.tgz' -o -name dropin.cache \) -print0 | $(XARGS) rm
+	find . \( -name '*.py[co]' -o -name dropin.cache \) -delete
+	find . \( -name '*.bak' -o -name dropin.cache \) -delete
+	find . \( -name '*.tgz' -o -name dropin.cache \) -delete
```
carlosperate: The reason I had to change this is because on windows the `rm` application from MYSYS complained about trying to execute it without arguments.
Anyway, even though this would be slightly slower (since there would be no parallelism at all), it's not really an issue for this projects.

carlosperate: Oh, and I must admit I have no idea what's going on with the `dropin.cache`, why was that added?

carlosperate:
Appveyor result: https://ci.appveyor.com/project/carlosperate/mu/build/1.0.357

Makefile
ntoll: That was added just over a year ago by @alistair-broomhead. IIRC they're used by Twisted (which we don't use). Al, why are they here? Am I correct in guessing this was perhaps an unneeded remnant from a cut and paste from another project?

PR #203, opened 2016-12-15
Run tests natively on AppVeyor (Windows with no MinGW)

carlosperate:
At the moment the tests are run on AppVeyor using the makefile, for which we need to run it under MinGW. While working on previous PRs I've noticed that the POSIX implementation does some translation (like path slashes) hiding some `py.test` failures due to platform differences.
This PR updates AppVeyor to run both `make test` and `py.test`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -62,7 +62,10 @@ install:
 # Not a project with an msbuild file, build done at install.
 build: None
 
+# Makefiles support provided by MinGW, which does POSIX translations, so run tests natively as well
 test_script:
+  - cmd: make clean
+  - cmd: py.test
   - cmd: make check
 
 # Push artifacts to s3 bucket and list all


ntoll:
Sorry... I completely missed this. :-(

.appveyor.yml
```diff
@@ -62,7 +62,10 @@ install:
 # Not a project with an msbuild file, build done at install.
 build: None
 
+# Makefiles support provided by MinGW, which does POSIX translations, so run tests natively as well
 test_script:
+  - cmd: make clean
+  - cmd: py.test
```
ntoll: Isn't the output of `py.test` also covered by `make check`..?

carlosperate: Yes, but because make is run through MinGW, there are some POSIX translations that are not representative of running the code directly on Windows. Say somebody submits a PR with hardcoded slashes instead of using `os.path.join()`, tests (or the application) might pass with `make clean` but fail when running py.test.

ntoll: Aha... of course. That makes sense.

PR #205, opened 2016-12-15
Select older OS X version for Travis build

### LINKED ISSUE #201: Unable to launch version 0.9.13 on Mac
I have downloaded the Mac package for version 0.9.13 from http://codewith.mu/ but it will not open. Running from the command line I get:

```bash
$ /Applications/mu\ 3.app/Contents/MacOS/mu 
Error loading Python lib '/var/folders/6q/v8c0hl0n1p1cy8jfzcwzqdt8tzd1h2/T/_MEIKwrhy1/Python': dlopen(/var/folders/6q/v8c0hl0n1p1cy8jfzcwzqdt8tzd1h2/T/_MEIKwrhy1/Python, 10): Symbol not found: _faccessat
  Referenced from: /var/folders/6q/v8c0hl0n1p1cy8jfzcwzqdt8tzd1h2/T/_MEIKwrhy1/Python
  Expected in: /usr/lib/libSystem.B.dylib
 in /var/folders/6q/v8c0hl0n1p1cy8jfzcwzqdt8tzd1h2/T/_MEIKwrhy1/Python
```

carlosperate:
With Pyinstaller on OS X we are only guaranteed compatibility on the OS version used for the build or newer. Travis was building on the default image 10.11 at the moment, so this PR reduces this to the lowest available, 10.9.5. Unfortunately this version is no longer supported and will be removed from Travis in January 20th, but we should use it for as long as we can and include this information in the release notes. 

Fixes #201

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,16 @@ matrix:
       sudo: required
       language: generic
 
+    # To maximise compatibility pick earliest image, 10.9 to be removed on 20/01/2017
     - os: osx
+      osx_image: beta-xcode6.1
       sudo: required
       language: generic
 
 before_install:
   # OS and Python info
   - uname -a
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then sw_vers; fi
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then python3 -c "import sys; print(sys.executable)"; fi
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo python3 -c "import sys; print(sys.executable)"; fi
 


ntoll:
To whom can we make representations about keeping this image (unsupported) for legacy reasons..? Just a thought...

carlosperate:
I think https://github.com/travis-ci/travis-ci/issues/6765 might be the current opened issue, no sure what the chances are for them to keep it.
Looks like brew is dropping support, don't really know for how long Apple will support it, some conversation online suggested that not for long, but my Google-foo failed me to find any definite information.

ntoll:
OK.... I've made our case and let's see what they say. Builds are taking a LONG time to finish. :-/

carlosperate:
@ntoll We have one week until the older OS X image is removed. Perhaps we should integrate this and try to make a minor release asap, so at the very least users with older versions of OS X can have Mu in its current state.

PR #208, opened 2016-12-19
Add safe file write

carlosperate:
Although not really fixing issue 200, this PR would introduce a safer way to write files that could would prevent data loss when issues arise during file writing.

Using the "w" mode on python erases the file before it starts writing, so if anything goes wrong during this operation the previous content of the file is lost. If on top of that the application crashes during this process then the user basically loses all their work. This is far from ideal from a text/code editor, so this PR introduces a the AtomicFile class to write to a temporary file and replace the original on file close.

I had a look at different python implementations of this type of functionality, and the most comprehensive modules/packages did very OS specific tricks in different platforms to ensure the operations are as atomic as possible using ctypes to access lower level OS functionality. I preferred to use a more "python only" approach, so forked https://github.com/sashka/atomicfile into https://github.com/carlosperate/atomicfile, enabled testing on more platforms with travis and appveyor and fixed some issues.

### DIFF
--- a/mu/contrib/atomicfile.py
+++ b/mu/contrib/atomicfile.py
@@ -1,0 +1,139 @@
+# -*- coding: utf-8 -*-
+import errno
+import os
+import sys
+import tempfile
+import codecs
+
+
+def _get_permissions(file_path):
+    """
+    Get the file permissions, if it can't access them returns system default.
+    """
+    try:
+        st_mode = os.lstat(file_path).st_mode & 0o777
+    except OSError as err:
+        if err.errno != errno.ENOENT:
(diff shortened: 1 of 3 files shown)


ZanderBrown:
Where are the temporary files saved?

carlosperate:
On the same location as the original source code file, with a prefix in the filename of ".<file_name>-", so "foo.py" would end up being something along the lines of ".foo.py-dsf564sdf6".

ZanderBrown:
Ok great.

ntoll:
Great stuff! Thanks.

PR #215, opened 2016-12-21
Implement make in python

tjguk:
Since I had to do something of the sort anyway, here's my working prototype of a make which implements[*] pycodestyle, pyflakes, clean, check, test

[*] Roughly; I haven't bothered to divine the exact meaning of some of the bash incantations. I don't think it's far off, though

### DIFF
--- a/make.py
+++ b/make.py
@@ -5,6 +5,10 @@ import shutil
 import subprocess
 import textwrap
 
+PYTEST = "pytest"
+PYFLAKES = "pyflakes"
+PYCODESTYLE = "pycodestyle"
+
 INCLUDE_PATTERNS = {
     "*.py"
 }
@@ -61,21 +65,21 @@ def test(*args):
     """Call py.test to run the test suite with additional args
     """
     print("\n\ntest")
-    return subprocess.call(["pytest"] + list(args))
+    return subprocess.call([PYTEST] + list(args))
 
 @export
 def coverage(*args):
     """Call py.test with coverage turned on
     """
     print("\n\ncoverage")
-    return subprocess.call(["py.test.exe", "--cov-config", ".coveragerc", "--cov-report", "term-missing", "--cov=mu", "tests/"])
+    return subprocess.call([PYTEST, "--cov-config", ".coveragerc", "--cov-report", "term-missing", "--cov=mu", "tests/"])
 
 @export
 def pyflakes(*args):
     """Call pyflakes on all .py files outside the docs and contrib directories
     """
     print("\n\npyflakes")
-    return _check_code("pyflakes.exe", *args)
+    return _check_code(PYFLAKES, *args)
 
 @export
 def pycodestyle(*args):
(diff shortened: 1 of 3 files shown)


make.py
```diff
@@ -0,0 +1,141 @@
...
+def check(*args):
+    """Run pyflakes + pycodestyle
+    """
+    print("\n\nCheck")
+    pyflakes(*args)
+    pycodestyle(*args)
+    coverage(*args)
+
+@export
+def clean(*args):
```
carlosperate: should we also clean any python bytecode files?

tjguk:
On 21/12/2016 17:48, Carlos wrote:
> *@carlosperate* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In make.py
> <https://github.com/mu-editor/mu/pull/215#pullrequestreview-14014879>:
>
>> +
> +@export
> +def pep8(*args):
> +    return pycodestyle(*args)
> +
> +@export
> +def check(*args):
> +    """Run pyflakes + pycodestyle
> +    """
> +    print("\n\nCheck")
> +    pyflakes(*args)
> +    pycodestyle(*args)
> +    coverage(*args)
> +
> +@export
> +def clean(*args):
>
> should we also clean any python bytecode files?

Yep: I just knocked this up really quickly!

TJG

carlosperate:
Now that https://github.com/mu-editor/mu/pull/218 is merged, might be a good idea to rebase this branch with master? Although now that I think about it, GitHub also offers a rebase option for PR merges, but I am not quite sure how much to trust it if some conflicts arise.

tjguk:
I'll rebase (if I can remember how) and push again...

ntoll:
`git rebase master` ;-)

tjguk:
I knew that!

tjguk:
rebased & pushed

carlosperate:
Looks like something didn't go that well, there is still 33 commits listed in the PR including master commits from your merge. Might be better to merge squash?

tjguk:
You can tell I don't use git much, can't you? I'm happy to do whatever's 
needed, but I'm not sure what's needed at this point.

Having already rebased, merged & pushed, what's the best way to produce 
the right result?

tjguk:
It's only one file, so I could just kick off a new branch directly from 
Master and issue a PR from there?

tjguk:
... and yes, I know: https://xkcd.com/1597/

carlosperate:
Haha, you'd be surprise how often I end up doing that as well... That or just making a copy of a git directory in case I botch everything up. Being a single file it would be quite easy to just add it to a new branch from master.

Btw, are we leaving the makefile in this repository or removing it?
If we now type in the terminal "make ..." most systems are configured to use the default Python interpreter, which is generally Python 2. Looks like at the moment this script is Python 3 only, so this is something to consider as well.

tjguk:
I'll create a new branch.

re the Makefile: at the risk of creating a perpetual interim, I'm 
inclined to leave it there until we're sure we've covered all the options.

I'm happy to make the script Py2/3 but I'd assumed that the base version 
for the project in general was Py3.

tjguk:
Superseded by #220

PR #220, opened 2016-12-22
Produce a functionally equivalent make.py 

tjguk:
This supersedes the branch for #215 and is branched directly from the newest master.

This make.py implements the same targets as the current Makefile but requires nothing beyond the stdlib


ntoll:
PEP8 errors breaking the build. :-(

tjguk:
I'm on it. Ironically my "make.py check" doesn't actually check make.py!

carlosperate:
From your previous comment:

>re the Makefile: at the risk of creating a perpetual interim, I'm inclined to leave it there until we're sure we've covered all the options.
>I'm happy to make the script Py2/3 but I'd assumed that the base version for the project in general was Py3.

This will depend on individual configurations, but for some people running something like `make clean` will default into `python2 make.py clean`, which will fail. Of course you can do `python3 make.py clean`, but by having the same name we might be blocking access to the "make" command.

tjguk:
If I were to put a shebang of "#!python3" or "#!/usr/bin/env python3" 
would that do the trick?

carlosperate:
It would definitely help, and it would probably be enough on linux and os x, but it would only work on Windows if they have the python launcher (py.exe, which is what allows you to do `py -2`, `py -3.5`, etc) installed and on the path environmental variable before they have any other python installation directories.

Unfortunately the Windows installer, interactions with older python installations, and out-of-the-box defaults have changed so much I would not be able to tell you how likely it is to be one way or another.

tjguk:
Ok, thanks. I'll add the shebang line for now and let's push it forward 
on that basis. We can address issues as they come up. In principle, 
make.py is a tool for developers -- however young -- who can be expected 
to understand some of the complexities of Python versions etc.

tjguk:
If no-one objects, I'll merge this PR -- at least it will get something in place for Windows devs (ie me...) to run the checks. We can then refine as we go

ntoll:
Apologies, I should have done this! 

It looks good to merge but for one thing: can we mention in the README how a Windows based developer can use the new `make.py` file?

Also (and I'm showing my Windows ignorance here) should the `make.py` file have some sort of executable flag? In any case, I'm asking myself "if I were an inexperienced Windows based developer do I have all the information I need to get going?"  This is quite an important question since I hope and imagine teachers and enthusiastic students (many of whom will be relatively inexperienced Windows based developers) will want to take a look at the code.

Basically, not only should the editor itself be beginner friendly, but so should the code base too (for educational reasons). :-)

Does this make sense?

tjguk:
> It looks good to merge but for one thing: can we mention in the README
> how a Windows based developer can use the new |make.py| file?

I'll do this

>
> Also (and I'm showing my Windows ignorance here) should the |make.py|
> file have some sort of executable flag?

No -- normally you'd just type "make.py test"; but in a venv (which is 
where I'm usually running this) you have to invoke Python explicitly. 
I'll try to put some words around it. I'm just firing up my Fedora box 
to test it on Linux, including the executability of it.

>
> Basically, not only should the editor itself be beginner friendly, but
> so should the code base too (for educational reasons). :-)
>
> Does this make sense?

Definitely

TJG

tjguk:
[On a train; this may come through twice]

Just making small mods for Linux. I think it would simplify things if we 
were to encourage and assume that we're running inside a virtualenv. The 
current README effectively does this, although it's not quite clear if 
this is aimed specifically at OS/X users or not.

The point is that, on Windows and on Linux at least, the "python" 
invocation differs according to whether you're in a venv or not. (It's 
more complicated on Windows because of the global py launcher which 
usually grabs the .py file association). I'd like to be able to say:

* Fire up a venv -- here's how you set one up (including how you get 
PyQt5 in there if the wheels aren't available) and here's how you 
activate it.

* For anything you do, use "python xxx" and it will use the venv setup. 
Job done.

As I say, the README already has most of this, only it's a kind of 
sidelight while I want to put it in the spotlight and say: all our 
instructions assume you're doing this.

Are we ok with this?

carlosperate:
If there are issues with the pyqt installation and virtual environments we could try using the --system-site-packages flag, and add that to the instructions. I would prefer that rather than adding the shim from the latest commit.

Ideally if users are using python 3.5 they shouldn't have that issue any more, but I would assume this would still be relevant if using the windows installer, apt-get or brew.

tjguk:
That's fine; I'll back out the shim and work up the instructions in the 
README. I can test on Windows & Linux 3.4/3.5 but not on Mac so I'll 
assume that the present (fairly comprehensive) instructions are correct.

I think it's good for us to state our baseline version -- which I'm 
assuming 3.4 (altho' I recognise your point the other day about dev 
tools and I'll test the make.py against 2.7 on each platform).

README.rst
```diff
@@ -163,28 +166,73 @@ For this to work you'll need to have Qt5 and Python 3 installed.
...
 
-     or::
+ or::
 
-        $ mkvirtualenv -p /usr/local/bin/python3 --system-site-packages py3
+    $ mkvirtualenv -p /usr/local/bin/python3 --system-site-packages py3
+
+or::
+
+    c:\dev\mu> py -3.5 -mvenv .venv35
```
carlosperate: Doesn't windows also need the `system-site-packages` flag?

carlosperate: Also, out of curiosity more than anything, is there a reason to invoke the python module rather than use a similar convention with `mkvirtualenv -p C:\Python35\python.exe`?

README.rst
```diff
@@ -163,28 +166,73 @@ For this to work you'll need to have Qt5 and Python 3 installed.
...
 
+If your setup hasn't yet included installing PyQt5 (eg because you're using
+Python 3.5+ on Windows for which wheels are available from PyPI)::
+
+    (.venv35) c:\dev\mu> pip install PyQt5 QScintilla
+
 To run the local development version of "mu", in the root of this repository
 type::
 
     $ python3 run.py
```
carlosperate: If we assume a virtual environment has been created, then should this be `python` instead of `python3`?

make.cmd
```diff
@@ -0,0 +1 @@
+python make.py %*
```
carlosperate: Is the make.py file python 2 compatible?

carlosperate: Should the same be done in the makefile?

tjguk:
Thanks for all the comments, @carlosperate. I forced myself to push a 
bunch of outstanding changes last night to get myself to move the whole 
thing forward.

I'll put together a "rationale" commentary on the PR and the revisit 
your comments to see if I have anything different to say :)

PR #229, opened 2017-01-18
Force Travis OS X to install Python 3.5

carlosperate:
Been testing this on my fork, so it should be working. However since travis seems to be taking almost 2h for each OS X build to commence (their worker backlog after 3pm goes from 0 to almost a 1000 jobs queued) I'm leaving this PR submitted tonight and hopefully should be all green in the morning.

We have until Friday before the current OS X 10.9 Mavericks image is retired and the oldest version we can be compatible becomes 10.10 Yosemite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo python3 -c "import sys; print(sys.executable)"; fi
 
 install:
-  # Python 3 and pip 3 installation are OS dependant
+  # Python 3 and pip 3 installation are OS dependent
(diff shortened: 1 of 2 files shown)


ntoll:
Looks like it worked!

carlosperate:
Thanks Nicholas, once the os x build from master is done (https://travis-ci.org/mu-editor/mu/jobs/192996515) we should do a final sanity check to make sure nothing has broken the build silently.

carlosperate:
Output: https://s3-us-west-2.amazonaws.com/ardublockly-builds/microbit/osx/mu-2017-01-18_11_39_32.zip

PR #230, opened 2017-01-19
Update CI servers to include branch and commit in filename

carlosperate:
Small change on the CI/build servers, changes the current name from:
 `mu_<date>_<time>`
to
`mu_<date>_<shorter time>_<branch name>_<7 digits commit hash>.<extension>`

So it is a bit longer, but it in essence it goes from something like:
 `mu-2017-01-17_18_16_36`
to
`mu_2017-01-17-18_16_ci_branch_hash_3c651fbf25.bin`
This way it maintains chronological order (up to a minute resolution) in http://ardublockly-builds.s3-website-us-west-2.amazonaws.com/index.html?prefix=microbit/ and it now also contains a lot more valuable information.

This is specially useful to be able to tell the difference between master builds, which should be more stable, and builds from WIP branches. It will also be really useful to identify specific test builds to keep around when cleaning up the old builds, as there is a few that we might want to keep around (for instance there is a windows executable with console output that will definitely keep coming in handy).

Apart from I have also updated AppVeyor to only deploy builds from master (this was the current behaviour for Travis already). So, when working in other branches we can easily and temporarily update the travis and appveyor scripts to include our WIP branch, this way we don't unnecessary pollute the AWS S3 bucket. Appveyor stores the artifacts on each build anyway, so that windows builds can always be downloaded without sending the .exe files to AWS.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -83,3 +83,5 @@ deploy:
   bucket: ardublockly-builds
   set_public: true
   folder: microbit\windows
+  on:
+    branch: [master]
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -57,7 +57,7 @@ install:
   - cmd: pyinstaller package\pyinstaller.spec
   - ps: Rename-Item -path .\dist upload
   - cmd: dir upload
-  - ps: Rename-Item .\upload\mu.exe mu-$(get-date -f yyyy-MM-dd_HH_mm_ss).exe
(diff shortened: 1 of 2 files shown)


carlosperate:
We are starting to see this error happen more often again:

```
py.test --cov-config .coveragerc --cov-report term-missing --cov=mu tests/

============================= test session starts ==============================

platform darwin -- Python 3.5.2, pytest-3.0.5, py-1.4.32, pluggy-0.4.0

rootdir: /Users/travis/build/mu-editor/mu, inifile: 

plugins: cov-2.4.0

collected 181 items 

tests/test_app.py ..

tests/test_interface.py ..................make: *** [coverage] Segmentation fault: 11
```

It's similar but not quite the same than the error we see when trying to run the test on the Travis CI Linux environment (this one is not intermittent, it always fails like this): https://travis-ci.org/mu-editor/mu/jobs/187949922#L380

I've restarted the travis job, hopefully it will pass this time.

Issue #231, opened 2017-01-19
mu crashes on OSX when launched via Finder when saving degree symbol


bobh46320:

ntoll:
Hi,

Thanks for reporting this.

Unfortunately, I cannot reproduce the problem. I assume you mean the "°" symbol..? I'm able to display the symbol in Mu, save and load files containing the symbol and flash code containing the symbol onto a micro:bit.

I'll need a bit more information to help me work out the problem:

* What operating system are you using?
* Can you attach the log file?
* Do you know which version of Mu you're using (are you using the latest version?)?
* What steps do I need to follow to reproduce the problem?

Thanks! :-)

bobh46320:
from microbit import *
display.scroll("°")

The above code will crash MicroPython for a MicroBit.  At the time of the crash, it will delete all source code from the source program file and then crash MU. I appreciate that the degree sign can't be written to the MicroBit LED array,  but I am outputting 4 bit strings to a 16 x 2 LCD screen where the symbol is legitimate.
I am running MU5 on a MAC.
Mac Stack Dump:-
Path:                  /Users/USER/Documents/*/mu-5.app/Contents/MacOS/mu
Identifier:            mu
Version:               0.0.0 (???)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           mu [25223]
User ID:               501

Date/Time:             2017-01-19 13:37:09.610 +0100
OS Version:            Mac OS X 10.12.2 (16C67)
Report Version:        12
Anonymous UUID:        D2E21157-AF7F-3AD2-B491-B1E9C6114EA2

Sleep/Wake UUID:       21634F2E-47A4-4978-B97E-AA62CC709E7A

Time Awake Since Boot: 150000 seconds
Time Since Wake:       210 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff95e5cdd6 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff95f48787 pthread_kill + 90
2   libsystem_c.dylib             	0x00007fff95d73497 raise + 26
[...]

PR #234, opened 2017-01-22
Add sys.excepthook to log errors and quit cleanly

inglesp:
This should make debugging things like #231 simpler.

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -39,7 +39,17 @@ def setup_logging():
     log_fmt = '%(asctime)s - %(name)s(%(funcName)s) %(levelname)s: %(message)s'
     logging.basicConfig(filename=LOG_FILE, filemode='w', format=log_fmt,
                         level=logging.DEBUG)
+    sys.excepthook = excepthook
     print('Logging to {}'.format(LOG_FILE))
+
+
+def excepthook(*exc_args):
+    """
+    Log exception and exit cleanly.
+    """
+    logging.error('Unrecoverable error', exc_info=(exc_args))
+    sys.__excepthook__(*exc_args)
+    sys.exit(1)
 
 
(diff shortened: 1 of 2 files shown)


mu/app.py
```diff
@@ -42,12 +42,21 @@ def setup_logging():
...
+    sys.exit(1)
+
+
 def run():
     """
     Creates all the top-level assets for the application, sets things up and
     then runs the application.
     """
     setup_logging()
+    sys.excepthook = excepthook
```
carlosperate: Just a suggestion, we could move this line into `setup_logging()`.

mu/app.py
```diff
@@ -42,12 +42,21 @@ def setup_logging():
     print('Logging to {}'.format(LOG_FILE))
 
 
+def excepthook(*exc_args):
+    """
+    Log exception and exit cleanly.
+    """
+    logging.error('Unrecoverable error', exc_info=(exc_args))
```
carlosperate: Might be useful to also propagate the exception through the normal route, that way we can still see the trace from running from the python source or executable command line:
```
sys.__excepthook__(exc_args)
```

carlosperate:
A bit unrelated to the PR itself (which LGTM), but for the sake of understanding #231 a bit better:

> This should make debugging things like #231 simpler.

I thought that issue caused a sudden crash and didn't get any Python exception raised, or it would have printed it to the console, no?

inglesp:
> I thought that issue caused a sudden crash and didn't get any Python exception raised, or it would have printed it to the console, no?

The exception does get written to stdout, but where does stdout go for an application launched from Finder?

carlosperate:
Right, I forgot it worked correctly when launched from the terminal.

PR #245, opened 2017-02-07
support for small screens

jg00dman:
button bar resizes according to window height. Both on load and after window resizes. Support for raspberry pi small screens.
My first ever contrib through git hub, first pyqt etc. Total newbie. Sorry if its bad.


ZanderBrown:
Replaced by #246

PR #250, opened 2017-02-21
Catch `OSError`s when saving

alexjh:
If an `OSError` occurs during file save, it should be caught and passed
along to the user.


ntoll:
Thank you for your contribution. Many apologies for not getting to this sooner.

alexjh:
No problem, thanks for merging it!

PR #251, opened 2017-02-21
Mock get_workspace_dir() to avoid file operations

alexjh:
`get_workspace_dir()` opens `settings.json`, which shouldn't be
happening during a unit test.

This also causes tests to fail on Python 3.4.3 (Ubuntu 14.04). The mocked `read_data` was being consumed when `json.load()` was being called, so when `f.read()` was being called, there wasn't any data left.

This was a really weird bug to track down, it's briefly alluded to here:
https://docs.python.org/3.4/library/unittest.mock.html#mock-open

> Changed in version 3.4: Added readline() and readlines() support. The mock of read() changed to consume read_data rather than returning it on each call.
> Changed in version 3.4.4: read_data is now reset on each call to the mock.


ntoll:
Thanks again. Same apology applies here too. ;-)

PR #254, opened 2017-02-28
Add python3-pyqt5.qtwebkit dependancy

gbaman:
Add python3-pyqt5.qtwebkit as a needed dependency for Debian/Ubuntu (tested on Raspberry Pi running Raspbian Jessie) for blocks support.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,7 +157,7 @@ You may need to grab the GIT submodules used for the blockly based UI::
 For this to work you'll need to have Qt5 and Python 3 installed.
 
 * On Debian based systems this is covered by installing: python3-pyqt5,
-  python3-pyqt5.qsci and python3-pyqt5.qtserialport.
+  python3-pyqt5.qsci, python3-pyqt5.qtserialport and python3-pyqt5.qtwebkit.
 
 * On Mac OS, first install PyQT5::
 
--- a/debian/control
+++ b/debian/control
@@ -8,6 +8,6 @@ Standards-Version: 3.9.1
 Package: mu
 Architecture: all
 Depends: ${misc:Depends}, ${python3:Depends}, python3-setuptools, python3-pyqt5,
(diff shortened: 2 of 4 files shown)


ZanderBrown:
@gbaman good catch!. It would be great if this could also be updated [here](https://github.com/mu-editor/mu/blob/blocks/debian/control), [here](https://github.com/mu-editor/mu/blob/blocks/package/README.rst) & [here](https://github.com/mu-editor/mu/blob/blocks/docs/development.rst)

gbaman:
@ZanderBrown fixed.

tjguk:
Just to make things interesting, it looks as though the QtWebKit stuff was deprecated earlier and removed as of 5.6: http://wiki.qt.io/New_Features_in_Qt_5.6#Removed_Modules

I was trying to find out what to pip-install on Windows and coming up blank. I'll dig a little further, but we may have to specify PyQt5 <= 5.5

carlosperate:
I think we had a bit of an issue with the available versions of PyQt for Python 3.4 and 3.5 and what they included in terms of the webkit or webengine widgets in the default Windows installation. Last time I looked into it was around Christmas in a conversation in the slack channel, so I don't really have the details at hand, but it is definitely something we need to take in consideration again once this branch is ready for merge.
On top of that we have a (hopefully temporary) compatibility issue with Python 3.5 and pyinstaller, so the CI servers are building Mu with Python3.4 and the PyQt installer (the PyQT wheels have been updated to also support 3.4, but last I checked some of its dependencies were still 3.5 only).

tjguk:
We seem to be at the unfortunate crux of changing PyQt support for: 
versions of Python; wheels; webkit/webengine (and, possibly, pyinstaller!)

ntoll:
Blocks isn't going to be in Mu any time soon (just use EduBlocks instead, then move over to Mu). I'll close this PR as a result. Thanks for the contribution @gbaman.

Issue #244, opened 2017-02-03
Investigate Travis segmentation fault during test


carlosperate:
When py.test is run under the Travis linux image it always fails during collection: https://travis-ci.org/mu-editor/mu/jobs/187949922#L380

```
$ py.test --cov-config .coveragerc --cov-report term-missing --cov=mu tests/

============================= test session starts ==============================

platform linux -- Python 3.5.2, pytest-3.0.5, py-1.4.32, pluggy-0.4.0
rootdir: /home/travis/build/mu-editor/mu, inifile: 
plugins: cov-2.4.0
collecting 2 items/home/travis/build.sh: line 57:  3327 Aborted                 (core dumped) py.test --cov-config .coveragerc --cov-report term-missing --cov=mu tests/

The command "py.test --cov-config .coveragerc --cov-report term-missing --cov=mu tests/" exited with 134.
```

Under OS X it seems to fail occasionally with the following issue, but restarting the build usually solves it:

```
py.test --cov-config .coveragerc --cov-report term-missing --cov=mu tests/

============================= test session starts ==============================
platform darwin -- Python 3.5.2, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /Users/travis/build/mu-editor/mu, inifile: 
plugins: cov-2.4.0
collected 188 items 

tests/test_app.py ...
tests/test_interface.py ..........make: *** [coverage] Segmentation fault: 11


The command "if [ "$TRAVIS_OS_NAME" = "osx" ]; then make check; fi" exited with 2.
```
This issue has been present for a long time and has been seen more often lately: https://github.com/mu-editor/mu/pull/243#issuecomment-277130113, https://github.com/mu-editor/mu/pull/156#issuecomment-248279367, https://github.com/mu-editor/mu/pull/103#issuecomment-211288636, https://github.com/mu-editor/mu/pull/230#issuecomment-273771037
The last time in https://github.com/mu-editor/mu/pull/243, it failed 4 times in a row, so this is something we should investigate.

carlosperate:
The Travis job from #243 merged into master is exhibiting the same issue: 
https://travis-ci.org/mu-editor/mu/jobs/197946892#L337

ntoll:
This completely not my area of expertise, but a quick Google for "Segmentation fault: 11" leads me to this: http://stackoverflow.com/questions/12762944/segmentation-fault-11 (the accepted response suggests the seg-fault is due to lack of memory). 

[...]

PR #264, opened 2017-04-22
Support for custom MicroPython runtimes

plummet555:
By default, Mu uses the MicroPython runtime that is bundled in uFlash.
This change enables custom runtimes to be used. The user can specify
the runtime in the settings.json file, e.g.:
"runtime_hex": "custom_runtime.hex"
The file can be a fully-qualified path and file name, or it can
just be a file name in which case it should be located in the
workspace directory.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -157,6 +157,37 @@ def get_workspace_dir():
     return workspace_dir
 
 
+def get_runtime_hex_path():
+    """
+    Returns the path to the hex runtime file - if this has been
+    specified under element 'microbit_runtime_hex' in settings.json.
+    This can be a fully-qualified file path, or just a file name
+    in which case the file should be located in the workspace directory.
+    Returns None if no path is specified or if the file is not present.
+    """
+    runtime_hex_path = None
+    settings_path = get_settings_path()
+    settings = {}
+    try:
+        with open(settings_path) as f:
+            settings = json.load(f)
+    except FileNotFoundError:
+        logger.error('Settings file {} does not exist.'.format(settings_path))
+    except ValueError:
+        logger.error('Settings file {} could not be parsed.'.format(
+                     settings_path))
+    else:
+        if 'microbit_runtime_hex' in settings and \
+                settings['microbit_runtime_hex'] is not None:
+            runtime_hex_path = os.path.join(
+                get_workspace_dir(),
+                settings['microbit_runtime_hex'])
+            if not os.path.exists(runtime_hex_path):
+                runtime_hex_path = None
+
+    return runtime_hex_path
+
+
 def check_flake(filename, code):
     """
     Given a filename and some code to be checked, uses the PyFlakesmodule to
@@ -387,11 +418,6 @@ class Editor:
             information = ("Your script is too long!")
             self._view.show_message(message, information, 'Warning')
             return
-        # Generate a hex file.
-        python_hex = uflash.hexlify(python_script)
-        logger.debug('Python hex:')
-        logger.debug(python_hex)
-        micropython_hex = uflash.embed_hex(uflash._RUNTIME, python_hex)
         # Determine the location of the BBC micro:bit. If it can't be found
         # fall back to asking the user to locate it.
         path_to_microbit = uflash.find_microbit()
@@ -411,9 +437,15 @@ class Editor:
         logger.debug('Path to micro:bit: {}'.format(path_to_microbit))
         if path_to_microbit and os.path.exists(path_to_microbit):
             logger.debug('Flashing to device.')
-            hex_file = os.path.join(path_to_microbit, 'micropython.hex')
-            uflash.save_hex(micropython_hex, hex_file)
+            # Flash the microbit
+            rt_hex_path = get_runtime_hex_path()
+            uflash.flash(paths_to_microbits=[path_to_microbit],
+                         python_script=python_script,
+                         path_to_runtime=rt_hex_path)
             message = 'Flashing "{}" onto the micro:bit.'.format(tab.label)
+            if (rt_hex_path is not None and os.path.exists(rt_hex_path)):
+                message = message + "\nRuntime: {}". \
+                    format(rt_hex_path)
             information = ("When the yellow LED stops flashing the device"
                            " will restart and your script will run. If there"
                            " is an error, you'll see a helpful message scroll"
@@ -671,7 +703,8 @@ class Editor:
         session = {
             'theme': self.theme,
             'paths': paths,
-            'workspace': get_workspace_dir()
+            'workspace': get_workspace_dir(),
+            'microbit_runtime_hex': get_runtime_hex_path()
         }
         logger.debug(session)
         settings_path = get_settings_path()
--- a/mu/contrib/uflash.py
+++ b/mu/contrib/uflash.py
@@ -12,6 +12,7 @@ import os
 import struct
 import sys
 from subprocess import check_output
+import time
 
 #: The magic start address in flash memory for a Python script.
 _SCRIPT_ADDR = 0x3e000
@@ -32,7 +33,7 @@ Documentation is here: https://uflash.readthedocs.io/en/latest/
 
 
 #: MAJOR, MINOR, RELEASE, STATUS [alpha, beta, final], VERSION
-_VERSION = (1, 0, 5, )
+_VERSION = (1, 0, 8, )
 
 
 def get_version():
@@ -89,7 +90,10 @@ def unhexlify(blob):
     for line in lines:
         # Discard the address, length etc. and reverse the hexlification
         output.append(binascii.unhexlify(line[9:-2]))
-    # Strip off "MP<size>" from the start
+    # Check the header is correct ("MP<size>")
+    if (output[0][0:2].decode('utf-8') != u'MP'):
+        return ''
+    # Strip off header
     output[0] = output[0][4:]
     # and strip any null bytes from the end
     output[-1] = output[-1].strip(b'\x00')
@@ -138,19 +142,30 @@ def extract_script(embedded_hex):
     Returns a string containing the original embedded script.
     """
     hex_lines = embedded_hex.split('\n')
-    # Blob will hold the extracted hex values representing the embedded script.
-    blob = ''
-    # Find the marker in the hex that comes just before the script
-    if ':020000040003F7' in hex_lines:
-        start_line = max(loc for loc, val in enumerate(hex_lines)
-                         if val == ':020000040003F7')
-        # Recombine the lines after that, but leave out the last 3 lines
-        blob = '\n'.join(hex_lines[start_line:-3])
-    if blob == '':
-        # If the result is the empty string, there was no embedded script
-        return ''
-    # Pass the extracted hex through unhexlify
-    return unhexlify(blob)
+    script_addr_high = hex((_SCRIPT_ADDR >> 16) & 0xffff)[2:].upper().zfill(4)
+    script_addr_low = hex(_SCRIPT_ADDR & 0xffff)[2:].upper().zfill(4)
+    start_script = None
+    within_range = False
+    # Look for the script start address
+    for loc, val in enumerate(hex_lines):
+        if val[0:9] == ':02000004':
+            # Reached an extended address record, check if within script range
+            within_range = val[9:13].upper() == script_addr_high
+        elif within_range and val[0:3] == ':10' and \
+                val[3:7].upper() == script_addr_low:
+            start_script = loc
+            break
+    if start_script:
+        # Find the end of the script
+        end_script = None
+        for loc, val in enumerate(hex_lines[start_script:]):
+            if val[9:41] == 'FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF':
+                end_script = loc + start_script
+                break
+        # Pass the extracted hex through unhexlify
+        return unhexlify('\n'.join(
+            hex_lines[start_script - 1:end_script if end_script else -3]))
+    return ''
 
 
 def find_microbit():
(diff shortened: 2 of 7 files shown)


README.rst
```diff
@@ -138,6 +138,18 @@ In terms of features - it's a case of less is more:
 
 That's it!
 
+MicroPython runtime developers
```
carlosperate: Maybe this is something we can add to https://github.com/mu-editor/help-pages ?

mu/logic.py
```diff
@@ -234,6 +234,55 @@ def check_pycodestyle(code):
...
+    else:
+        if 'runtime_hex' in settings and settings['runtime_hex'] != "":
+            runtime_hex_path = os.path.join(
+                get_workspace_dir(),
+                settings['runtime_hex'])
+
+    return runtime_hex_path
+
+
+def get_runtime_hex_str():
```
carlosperate: Is this this function really necessary? uFlash itself can take the path to a runtime hex file, so it'd be better not to duplicate efforts and pass that to uFlash directly

carlosperate:
Thanks @plummet555 , I've added a couple of comments, but I haven't had a chance to test it locally. 
One other thing I was thinking about is making the setting name a bit more explicit, as in the future Mu might be able to target more than just the micro:bit, maybe something along the lines of `microbit_runtime_hex`?

README.rst
plummet555: Sounds good, I'll push something for that

mu/logic.py
plummet555: I'd prefer that too, but I don't see an equivalent of the embed_hex function that takes a hex file path? I see that we could call flash() which will take a hex file path plus the python file, but logic.py already seems to duplicate some of that?

plummet555:
Hi - agree with microbit_runtime_hex, I'll do that once we've figured out the embed_hex vs. flash question. Thanks for the feedback

plummet555:
Hi - I've pushed a change for  https://github.com/mu-editor/help-pages and submitted a pull request

plummet555:
Pushed the changes except for the get_runtime_hex_str change for now

plummet555:
Hi @carlosperate 

I've just been looking more at whether we can avoid the need for get_runtime_hex_str.

I think the issue is that logic.py currently does this to flash the device:

- uflash.hexlify
- uflash.embed_hex
- uflash.save_hex

rather than simply calling uflash.flash which can take care of all of those steps.

I can't see a good reason why it is done that way? If you agree, I can rework logic.py a bit more to avoid those steps. That way we can optionally pass in the runtime file name and we would no longer need get_runtime_hex_str. 

I think we would keep the code in logic.py that determines the path to the microbit (rather than leaving that to uflash.flash) as that code supports prompting the user for the path and that is something uflash.flash doesn't do.

ZanderBrown:
I think the point is that `uflash.flash` takes a filepath whereas `uflash.hexify`, `uflash.embed_hex` & `uflash.save_hex` work with strings.

This means the current file doesn't need to be saved to be flashed.

Perhaps a `flash_string` method should be added to uflash?

plummet555:
>> This means the current file doesn't need to be saved to be flashed.

Good spot. That must be why it is done the current way.

Would we want to change uflash locally though?

ZanderBrown:
I was meaning open a issue / PR on [ntoll/uflash](https://github.com/ntoll/uflash)

carlosperate:
Rather than have the same type of operation duplicated in both Mu and uFlash I think it would be better to refactor uFlash a bit and expose a way to just send a string for the code and a path for the hex file.
@ntoll thoughts?

ZanderBrown:
Perhaps we could take
```python
def flash(path_to_python=None, paths_to_microbits=None, path_to_runtime=None):
```
and expand to 
```python
def flash(path_to_python=None, python_source=None, paths_to_microbits=None, path_to_runtime=None):
```
with `uflash.flash` then calling `uflash.hexify`, `uflash.embed_hex` & `uflash.save_hex` internally allowing us to call
```python
uflash.flash(python_source=python_script, path_to_runtime=get_runtime_hex_path())
```
with `get_runtime_hex_path()` adjusted to return `None` when the field is empty (instead of an empty string)

plummet555:
That sounds very good @ZanderBrown. I'd be happy to rework/rebase my change if/when that expanded uflash function is merged.

ZanderBrown:
I'm sure @ntoll will probably mention the fact I haven't added any test (sorry, I understand the importance, not so much the implementation) but one PR ready for comments/review/merging

mu/logic.py
```diff
@@ -234,6 +234,56 @@ def check_pycodestyle(code):
     return style_feedback
 
 
+def get_runtime_hex_path():
```
ZanderBrown: To keep similar things together would this not be better put after `get_workspace_dir` rather than `check_pycodestyle` (keep all the settings related functions together)

mu/logic.py
```diff
@@ -234,6 +234,56 @@ def check_pycodestyle(code):
...
+            settings = json.load(f)
+    except FileNotFoundError:
+        logger.error('Settings file {} does not exist.'.format(settings_path))
+    except ValueError:
+        logger.error('Settings file {} could not be parsed.'.format(
+                     settings_path))
+    else:
+        if 'microbit_runtime_hex' in settings and \
+                settings['microbit_runtime_hex'] != "":
+            runtime_hex_path = os.path.join(
```
ZanderBrown: Don't make the mistake i did implementing `get_workspace_dir`!

Check the path exists (and is a file) here rather than later then this function will always return a *safe* value. (i.e. a valid path or nothing)

Regardless of whether or not my PR gets merged in uflash it would probably be better to return `None` instead of an empty string & check that the setting wasn't set to "" (more pythonic and all that).

mu/logic.py
```diff
@@ -413,7 +463,12 @@ def flash(self):
             logger.debug('Flashing to device.')
             hex_file = os.path.join(path_to_microbit, 'micropython.hex')
             uflash.save_hex(micropython_hex, hex_file)
+            rt_hex_path = get_runtime_hex_path()
             message = 'Flashing "{}" onto the micro:bit.'.format(tab.label)
+            if (rt_hex_path != "" and os.path.exists(rt_hex_path)):
```
ZanderBrown: Rather pointless checking if the path exists now (things would have already broken)

ZanderBrown:
*Warning mile long comment incoming*

Should my PR be meregd in uflash this PR will need to be updated with that new uflash version, `get_runtime_hex_path` changed to return None when a custom hex isn't set, `get_runtime_hex_str` could be scrapped (removing duplication highlighted by @carlosperate) and

```python
        # Generate a hex file.
        python_hex = uflash.hexlify(python_script)
        logger.debug('Python hex:')
        logger.debug(python_hex)
        micropython_hex = uflash.embed_hex(get_runtime_hex_str(), python_hex)
        # Determine the location of the BBC micro:bit. If it can't be found
        # fall back to asking the user to locate it.
        path_to_microbit = uflash.find_microbit()
        if path_to_microbit is None:
            # Has the path to the device already been specified?
            if self.user_defined_microbit_path:
                path_to_microbit = self.user_defined_microbit_path
            else:
                # Ask the user to locate the device.
                path_to_microbit = self._view.get_microbit_path(HOME_DIRECTORY)
                # Store the user's specification of the path for future use.
                self.user_defined_microbit_path = path_to_microbit
                logger.debug('User defined path to micro:bit: {}'.format(
                             self.user_defined_microbit_path))
        # Check the path and that it exists simply because the path maybe based
        # on stale data.
        logger.debug('Path to micro:bit: {}'.format(path_to_microbit))
        if path_to_microbit and os.path.exists(path_to_microbit):
            logger.debug('Flashing to device.')
            hex_file = os.path.join(path_to_microbit, 'micropython.hex')
            uflash.save_hex(micropython_hex, hex_file)
            rt_hex_path = get_runtime_hex_path()
            message = 'Flashing "{}" onto the micro:bit.'.format(tab.label)
            if (rt_hex_path != "" and os.path.exists(rt_hex_path)):
                message = message + "\nRuntime: {}". \
                    format(rt_hex_path)

            information = ("When the yellow LED stops flashing the device"
                           " will restart and your script will run. If there"
                           " is an error, you'll see a helpful message scroll"
                           " across the device's display.")
            self._view.show_message(message, information, 'Information')
        else:
```

would become

```python
        # Determine the location of the BBC micro:bit. If it can't be found
        # fall back to asking the user to locate it.
        path_to_microbit = uflash.find_microbit()
        if path_to_microbit is None:
            # Has the path to the device already been specified?
            if self.user_defined_microbit_path:
                path_to_microbit = self.user_defined_microbit_path
            else:
                # Ask the user to locate the device.
                path_to_microbit = self._view.get_microbit_path(HOME_DIRECTORY)
                # Store the user's specification of the path for future use.
                self.user_defined_microbit_path = path_to_microbit
                logger.debug('User defined path to micro:bit: {}'.format(
                             self.user_defined_microbit_path))
        # Check the path and that it exists simply because the path maybe based
        # on stale data.
        logger.debug('Path to micro:bit: {}'.format(path_to_microbit))
        if path_to_microbit and os.path.exists(path_to_microbit):
            logger.debug('Flashing to device.')
            # Flash the microbit
            uflash.flash(paths_to_microbits=[path_to_microbit],
                         python_script=python_script,
                         path_to_runtime=get_runtime_hex_path())
            rt_hex_path = get_runtime_hex_path()
            message = 'Flashing "{}" onto the micro:bit.'.format(tab.label)
            if (rt_hex_path != "" and os.path.exists(rt_hex_path)):
                message = message + "\nRuntime: {}". \
                    format(rt_hex_path)

            information = ("When the yellow LED stops flashing the device"
                           " will restart and your script will run. If there"
                           " is an error, you'll see a helpful message scroll"
                           " across the device's display.")
            self._view.show_message(message, information, 'Information')
        else:
```

mu/logic.py
plummet555: Agree with None instead of "" (I'm still an old C guy...)

Regarding checking whether the path exists: the reason I didn't do that here is because of this flow:

1. User puts a path in the settings.json but hasn't actually put the hex file at that location yet
2. They run Mu and flash
3. This function would then return None as the file specified in settings.json doesn't exist
4. They exit Mu and the None gets persisted to settings.json
5. They get round to putting the hex file in the right place and run Mu again, only to find the path magically disappeared from settings.json

Which I think might be annoying?

Maybe there is a way around that where we just don't touch whatever path they put in the settings.json even if it is garbage - but then I think we'd need some other way of reading it so it can be echoed back when the settings.json is recreated?

mu/logic.py
plummet555: Agreed

mu/logic.py
plummet555: That's just to determine whether to display the path in the popup dialog when flashing. We only want to show the path if it was actually used (it might point to a path which doesn't exist, in which case the default runtime will have been used).

mu/logic.py
ZanderBrown: I see your point but i'd have thought that if a user has done the (very manual) editing of the settings file they have probably already acquired their hex and put it where they said they did

I must admit our current handling of settings is distinctly imperfect.

plummet555: Yep OK, I think you are right. I'll rework it that way

plummet555:
I'm ready to rework this once:

ZanderBrown:
Just waiting for @ntoll to start the chain reaction.

ntoll:
OK... the upstream PR in uflash is merged. Knock yourself out! :-)

plummet555:
PR updated. I did a force push rather than end up with a chain of commits..hope that is normal practice.

plummet555:
I don't know what that continuous build failure is about..

ZanderBrown:
That OSX travis error has me befuddled as well

Whilst not a problem it would probably have been better to leave as is for @ntoll to decide whether or not to merge & squash

mu/logic.py
```diff
@@ -411,9 +437,15 @@ def flash(self):
         logger.debug('Path to micro:bit: {}'.format(path_to_microbit))
         if path_to_microbit and os.path.exists(path_to_microbit):
             logger.debug('Flashing to device.')
-            hex_file = os.path.join(path_to_microbit, 'micropython.hex')
-            uflash.save_hex(micropython_hex, hex_file)
+            # Flash the microbit
+            uflash.flash(paths_to_microbits=[path_to_microbit],
+                         python_script=python_script,
+                         path_to_runtime=get_runtime_hex_path())
+            rt_hex_path = get_runtime_hex_path()
```
ZanderBrown:
```py
# Flash the microbit
uflash.flash(paths_to_microbits=[path_to_microbit],
                   python_script=python_script,
                   path_to_runtime=get_runtime_hex_path())
rt_hex_path = get_runtime_hex_path()
```
Do you see what i see? Two calls to `get_runtime_hex_path()`, 1 for the flashing, 2 for possible inclusion in the message.

What would probably make more sense is:
```py
# Flash the microbit
rt_hex_path = get_runtime_hex_path()
uflash.flash(paths_to_microbits=[path_to_microbit],
                   python_script=python_script,
                   path_to_runtime=rt_hex_path)
```

carlosperate:
I've restarted the travis build, we've encountered some intermittent issues with travis: https://github.com/mu-editor/mu/issues/244
Hopefully it will run correctly the second time.

mu/logic.py
plummet555: Agreed

plummet555:
New commit pushed

ZanderBrown:
LGTM

plummet555:
Pushed one more commit to get test coverage back to 100%. That should be it now...

ZanderBrown:
@carlosperate we covered everything ​now?

plummet555:
All gone quiet.. anything I need to do?

ZanderBrown:
Let's give @ntoll & @carlosperate a ping.

ntoll:
Hurrah! Merged and many thanks for all your efforts with this. Apologies for the tardy response time on my part.

plummet555:
Great - thanks!

carlosperate:
Yeah, thanks @plummet555 this will be quite a useful feature!
We should try to get something written down in the documentation as well: https://github.com/mu-editor/help-pages

carlosperate:
Ups, just noticed you've already done that as well!
https://github.com/mu-editor/help-pages/pull/1

plummet555:
Oh yup that needs to be reviewed/merged as well

PR #269, opened 2017-05-23
prevent opening the same file twice, but also help the user by focusi…

matt-land:
prevent opening the same file twice, but also help the user by focusing the tab they attempted to open

### DIFF
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -653,7 +653,7 @@ class Window(QStackedWidget):
     @property
     def current_tab(self):
         """
-        Returns the currently focused tab.
+        Returns the currently focussed tab.
         """
         return self.tabs.currentWidget()
 
--- a/mu/interface.py
+++ b/mu/interface.py
@@ -600,7 +600,7 @@ class FileTabs(QTabWidget):
         """
         window = self.nativeParentWidget()
         modified = window.current_tab.isModified()
-        if (modified):
+        if modified:
             msg = 'There is un-saved work, closing the tab will cause you ' \
                   'to lose it.'
             if window.show_confirmation(msg) == QMessageBox.Cancel:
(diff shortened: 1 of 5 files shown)


mu/logic.py
```diff
@@ -566,6 +566,15 @@ def load(self):
         """
         path = self._view.get_load_path(get_workspace_dir())
         logger.info('Loading script from: {}'.format(path))
+        # see if file is open first
+        for widget in self._view.widgets:
+            if widget.path is None:  # this widget is an unsaved buffer
+                continue
+            if path in widget.path:
+                self._view.show_message('Requested file is already open.')
```
ZanderBrown: How about 'that file is already open' or include the filename.
Remember that the message needs to be clear perferably using simple language but without appearing condescending

matt-land: I changed the wording to "That file is already open"
The other suggestion-I was not sure how to express the filename to the user while being simple.

ZanderBrown: Something along the lines of 'The file "myprog.py" is already open' (without file path)
But that's getting petty over a single sentence in a solution to a problem we probably should have anticipated

mu/interface.py
```diff
@@ -653,7 +653,7 @@ def connect_zoom(self, widget):
     @property
     def current_tab(self):
         """
-        Returns the currently focused tab.
+        Returns the currently focussed tab.
```
ZanderBrown: Is it supposd to have a double s?

mu/logic.py
matt-land: OK

mu/interface.py
matt-land: My IDE flagged it, so I fixed it. When I googled, it seemed valid.  Since I had doubts I pulled it.

ntoll:
Thanks for the contributions and many apologies for taking so long to getting around to reviewing/merging the PR. I had to make a few minor changes to fix PEP8 and test coverage problems.

Issue #268, opened 2017-05-14
Open file from command line


peter-jensen:
Unable to open a file from the command line. Double clicking on a python script does therefore not work.
I.e ./run.py test.py does not open file.
Version of "mu".  37f4c7e365797151bfdd32b143098d0a7cd0f033
Platform Ubuntu 16.04.

ZanderBrown:
Good one, this should probably be a priority with our change of focus

peter-jensen:
If files can be opened by clicking on them, then there is a high probability that several invocations of the editor may exist. This could affect common resources such as .local/share/mu/settings.json and .cache/mu/log/mu.log.

ZanderBrown:
Single instance is probably worth looking into

carlosperate:
Implemented in #270, thanks @matt-land!

PR #270, opened 2017-05-23
Feature: open from command line

matt-land:
This addresses some of "Open from the command line"
https://github.com/mu-editor/mu/issues/268

    This patch will allow mu to handle the following case:
    - mu is not running
    - mu is registered with the OS to handle the file
    - the user double clicks on the file
    
    Since the file being 'invoked' may or may not have been in the previous mu session,
    that is detected while doing session restore, and the 'invoked' file is opened and focused last.


ntoll:
Hey hey... thanks for this..! Before merging I simplified the tests and removed a duplication in the tests. Great work!

carlosperate:
Yes, thanks! we can finally "open with" the Mu executable, including drag-and-drop files onto it.

PR #288, opened 2017-09-26
Repackage

XECDesign:
Updated packaging files and added missing dependency (pyqt5.qtserialport).

I don't know how the setup.py change affects other systems, but this makes sure that the resulting deb includes the python3-pyqt.qtserialport dependency, which is otherwise missing. Let me know if that breaks anything.

I've attached the resulting .deb file for testing:
[mu_1.0.0~beta8_all.zip](https://github.com/mu-editor/mu/files/1334293/mu_1.0.0.beta8_all.zip)

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -3,3 +3,5 @@
 export PYBUILD_NAME=mu
 %:
 	dh $@ --with python3 --buildsystem=pybuild
+
+override_dh_auto_test:
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+mu (1.0.0~beta8) stretch; urgency=medium
+
+  * Repackage
+
+ -- Serge Schneider <serge@raspberrypi.org>  Tue, 26 Sep 2017 17:39:30 +0100
+
 mu (0.9.13-1) unstable; urgency=low
 
   * Add ability to change default Python directory in the settings file. Thanks to Zander Brown for the contribution. See #179.
--- a/debian/control
+++ b/debian/control
@@ -1,13 +1,15 @@
 Source: mu
-Maintainer: Nicholas Tollervey <ntoll@ntoll.org>
 Section: python
 Priority: optional
-Build-Depends: python3-setuptools, python3, debhelper (>= 9)
-Standards-Version: 3.9.1
+Maintainer: Serge Schneider <serge@raspberrypi.org>
+Build-Depends: debhelper (>= 9), dh-python, python3-all, python3-setuptools
+Standards-Version: 3.9.8
+Homepage: https://codewith.mu/
+X-Python3-Version: >= 3.2
 
 Package: mu
 Architecture: all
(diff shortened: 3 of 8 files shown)


ntoll:
Thank you!

debian/rules
```diff
@@ -3,3 +3,5 @@
 export PYBUILD_NAME=mu
 %:
 	dh $@ --with python3 --buildsystem=pybuild
+
+override_dh_auto_test:
```
carlosperate: This might be a stupid question, but woudn't the colon from this line indicate that perhaps there should be a command after it?

carlosperate:
Thanks @XECDesign!
I hope you don't mind me asking, as I'd like to learn a bit more about deb packages: how are the rest of the python dependencies installed? Is the setup.py file automatically used to install them via pip? If that's the case, how is that invoked in these files?
What does the `py3dist-overrides` file does? does it overwrite the named dependencies from the setup.py file with the listed apt packages?

debian/rules
XECDesign: It's not required. It's just a makefile and we're telling that target to do nothing. We could add a 'true' there to make it more readable, but it behaves the same way.
I didn't add anything because this is the way the example is given in the dh manpage.

XECDesign:
The python dependencies are determined automatically by reading setup.py. They're then added to the `${python3:Depends}` placeholder in the control file. In this case `py3dist-overrides` is helping dh_python determine the correct packages to add as dependencies because it couldn't determine a few of them.

Instead of using pip to install any dependencies, it's pulling in the existing apt packages instead.

ntoll:
One thing to note @XECDesign, there's no `pyqt5.qtserialport` package in PyPI (it's automatically included in the full PyQt related wheels). This is in contrast with the Debian packages that seem to have been split up rather than contained within the Qt5 deb. 

Ergo, I've had to remove `pyqt5.qtserialport` from `setup.py` because it was causing an error.

There's always an exception... :-/

XECDesign:
No worries, we can always do it another way:
https://github.com/mu-editor/mu/pull/289

ntoll:
So I notice... ;-)

carlosperate:
Thanks for the explanation @XECDesign!
I take it there is now a `python3-pycodestyle` package? This Python package changed named a while ago, but some distros were a bit behind with the packaging. 
Also, does dh_python find the correct pyflakes for Python 3? This one was also a slightly odd package in Jessie.

XECDesign:
I can't really vouch for all the dependencies being correct, but it certainly installs those two packages. I've only tested that it builds, runs and seems to work.

carlosperate:
If it opens then that's already a pretty good sign that all imports are, at least, satisfied. Maybe you could run the code checker (thumbs up icon) once as a sanity test, but I'd be surprised if there is a problem.

XECDesign:
Yup, that works.

carlosperate:
Cool, thanks @XECDesign!

PR #296, opened 2017-11-09
Explain special developer installation prerequisites for the Pi

romilly:
Clarifies the need to install Raspbian Packages and use a custom requirements file when building for the Raspberry Pi.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -67,10 +67,18 @@ repository with the following command::
 
     $ git clone https://github.com/mu-editor/mu.git
 
-Ensure you have the correct dependencies for development installed by creating
-a virtualenv and running::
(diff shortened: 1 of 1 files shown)


carlosperate:
Thanks @romilly!

It's also possible that some `pip` dependencies will try to compile against the python headers, so it's probably a good idea to add the `python3-dev` package as well. I assume that's the package name in raspbian but I cannot check at the moment, could you check that it installs correctly with `apt-get`?

ntoll:
@romilly thank you.

I made some minor edits as per suggestions from @carlosperate. I also took this opportunity to correct some Makefile related output inconsistencies.

PR #321, opened 2018-01-27
add more adafruit boards

ladyada:
ez pz lemon squeezy


ntoll:
@ladyada is this an update or can it wait (for three weeks or so when I've finished a bunch of other Mu related work so it's all wrapped up into a single release)..? Happy to do a quick cut/release if needed.

ladyada:
no rush on these, they dont 'exist' yet :) 3 weeks is fine!

PR #359, opened 2018-03-02
up pyzero dependency to the latest version 1.2

yonghuming:
up pyzero dependency to the latest version 1.2

the pygame zero have updated to version 1.2


ntoll:
Absolutely we should do this! :-) Thank you!

Issue #361, opened 2018-03-02
Mu always adds .py to a file, even if the filename as a file extenstion.


martinohanlon:
When saving a file Mu will always add a .py extension to the end of the file. 

Super, unless of course you want to create a file which has a different file extension, e.g. a .txt file which your program loads in at runtime.

Using this model is really common when getting learners to work with files.

.e.g 
First create a file which has a list of names in it and called it names.txt. 
Open names.txt in your Python program.
Read in the names.
etc, etc.

martinohanlon:
I would recommend that if a filename is presented which doesnt have a file extension it adds .py, if a file extension is included it doesn't.

ntoll:
+1 on that suggested behaviour.

martinohanlon:
Submitted PR https://github.com/mu-editor/mu/pull/362

PR #362, opened 2018-03-02
Resolve issue 361

martinohanlon:
1 line code change to resolve https://github.com/mu-editor/mu/issues/361


ntoll:
Oooh... thank you!

Umm... are there tests..? Don't worry if not, I'll add some when I've finished squashing the bug I'm looking into right now.

martinohanlon:
I was going to make a note re tests. I couldn't find an obvious save test I could amended.

If you make a test could you make a reference to it, so I can see how you implemented it.

ntoll:
All code branches have tests, so the original will be tested somewhere. You'll see the test when I merge your code. :-)

martinohanlon:
I have found it. I will try and write the test.

ntoll:
Perfect! Thank you @martinohanlon! :-)

mu/logic.py
```diff
@@ -525,8 +525,8 @@ def save(self):
             tab.path = self._view.get_save_path(workspace)
         if tab.path:
             # The user specified a path to a file.
-            if not os.path.basename(tab.path).endswith('.py'):
-                # No extension given, default to .py
+            if os.path.splitext(os.path.basename(tab.path))[1] == '':
```
ZanderBrown: Call to `os.path.basename()` is unnecessary as `os.path.splitext` handles it internally

```py3
>>> import os
>>> os.path.splitext('/non.sense/to.test/this/thing.py.in')
('/non.sense/to.test/this/thing.py', '.in')
>>> os.path.splitext('/non.sense/to.test/this/thing')
('/non.sense/to.test/this/thing', '')
>>> 
```

tjguk:
@ZanderBrown given that the PR is already merged, I think we could leave this under the heading of "redundant but harmless"?

PR #364, opened 2018-03-02
No file encoding specified

tjguk:
This PR addresses the isssue of saving / autosaving when the editor text contains a codepoint which can't be encoded via the Windows default cp1252 (a variation on iso-8859-1).

It also refactors very slightly and tidies up the file-writing mechanism

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -529,7 +529,7 @@ class Editor:
                 # No extension given, default to .py
                 tab.path += '.py'
             try:
-                with open(tab.path, 'w', newline='') as f:
+                with open(tab.path, 'w', newline='', encoding="utf-8") as f:
                     logger.info('Saving script to: {}'.format(tab.path))
                     logger.debug(tab.text())
                     write_and_flush(f, tab.text())
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -138,8 +138,8 @@ logger = logging.getLogger(__name__)
 
 def write_and_flush(fileobj, content):
     """
-    Writes content to the fd, then flushes and fsyncs to ensure the data is, in
-    fact, written.
+    Writes content to the fileobj, then flushes and fsyncs to ensure the data is, 
+    in fact, written.
     """
     fileobj.write(content)
     fileobj.flush()
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -136,14 +136,19 @@ MOTD = [  # Candidate phrases for the message of the day (MOTD).
 logger = logging.getLogger(__name__)
 
 
-def write_and_flush(fd, content):
+def write_and_flush(fileobj, content):
     """
     Writes content to the fd, then flushes and fsyncs to ensure the data is, in
     fact, written.
     """
-    fd.write(content)
-    fd.flush()
-    os.fsync(fd)
+    fileobj.write(content)
+    fileobj.flush()
+    #
+    # Theoretically this shouldn't work; fsync takes a file descriptor,
+    # not a file object. However, there's obviously some under-the-cover
+    # mechanism which converts one to the other (at least on Windows)
+    #
+    os.fsync(fileobj)
 
 
 def get_settings_path():
@@ -511,6 +516,22 @@ class Editor:
         """ for loading files passed from command line or the OS launch"""
         self._load(path)
 
+    def save_tab_to_file(self, tab):
+        try:
+            with open(tab.path, 'w', newline='', encoding="utf-8") as f:
+                logger.info('Saving script to: {}'.format(tab.path))
+                logger.debug(tab.text())
+                write_and_flush(f, tab.text())
+            tab.setModified(False)
+            self.show_status_message(_("Saved file: {}").format(tab.path))
+        except OSError as e:
+            logger.error(e)
+            message = _('Could not save file.')
+            information = _("Error saving file to disk. Ensure you have "
+                            "permission to write the file and "
+                            "sufficient disk space.")
+            self._view.show_message(message, information)
+    
     def save(self):
         """
         Save the content of the currently active editor tab.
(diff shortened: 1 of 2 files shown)


mu/logic.py
```diff
@@ -136,14 +136,19 @@
 logger = logging.getLogger(__name__)
 
 
-def write_and_flush(fd, content):
+def write_and_flush(fileobj, content):
     """
     Writes content to the fd, then flushes and fsyncs to ensure the data is, in
```
carlosperate: `Writes content to the fd` -> `Writes content to the fileobj`

carlosperate:
So we are writing the files as UTF-8, but what happens when a file is opened in a different format? Did that work before (open and save in the local encoding)?

Just wondering if we are going to break somebody's script by forcing the encoding in mu, specially with the autosave feature. It's even worse if Mu crashes during save and the script is lost altogether.

tjguk:
"what happens when a file is open in a different format?" I was considering this a different issue, but ISTM that two things can happen: a UnicodeDecodeError or mojibake. If the former, we can at least spot it happening and raise some kind of warning/error -- even offering to auto-detect (via chardet etc.). If the latter, we won't necessarily know, although we could pre-emptively chardet and warn if it looks like something else, although UTF-8 will decode it.

mu/logic.py
tjguk: fixed; thanks

ntoll:
Hi folks,

@tjguk, thanks for fixing this up.

I'm going to play devils' advocate here: Mu is for beginner programmers whose code is likely to only ever be opened with Mu. If you're using Mu and you're NOT a beginner programmer (I'm assuming you may perhaps be using it for development on one of Adafruit's boards), then you *probably* have enough experience to understand the impact of different encodings. 

Assuming the above, perhaps the best solution is related to documentation: we explain we enforce UTF-8 when saving a file so experienced developers can see what we're up to (and adjust accordingly) and it means beginner programmers never have to encounter any mixed-encoding problems themselves since they're only using Mu.

Does this make sense?

tjguk:
The edge case is someone who's previously saved under Mu using cp1252 in such a way etc. etc. I think we can just decode with a charref replace handler and then at least something appears in the editor which they can then fix up.

ntoll:
Is that something we can include in this PR, so the fix and guard against faulty prior behaviour all lands at the same time..?

carlosperate:
I'm still concerns about autosave, @ntoll is that a feature that is always on?

Say I'm a user that is used to opening a non-utf-8 files with Mu, now after an update I open the same file, the unrecognised characters get all mangled up, and before I even notice the autosave has rewritten my file to this "corrupted-looking" version.
Is that a possibility?

ntoll:
@carlosperate good points... Here's the context (I guess we have a classic case of requirements pulling against each other, in which case the solution will be a compromise. Then the trick becomes finding the simplest/best compromise so people understand *why* we have such a compromise.)

Autosave was a "must have" from Carrie Anne. The number of classroom hours that are wasted because kids forget to save their work is, apparently, a source of great frustration to all involved. When I demo this to teachers, it's like when I show them the zoom features for when they're using a projector: I get an "at last, someone has listened to our concerns" type reaction.

However, for the Adafruit mode this feature is disabled (https://github.com/mu-editor/mu/blob/master/mu/modes/adafruit.py#L36) because of the auto-restart feature built into CircuitPython. I believe this is an understandable (see above) and simple compromise.

Turning to your question, if you're a user opening non-UTF-8 files with Mu, then the possibilities you outline exist. However, I would argue that you *shouldn't* be opening non-UTF-8 files that could be corrupted in this way. That you're doing this in the first place suggests you're advanced enough a user to be using more than one development tool, and secondly, that you know enough about the different encodings to be using different ones in different contexts.

I believe my point about beginner programmers using only Mu for their Mu related projects still stands. However, I'm also painfully aware that I'm likely blind to obvious edge cases against this line of thought. So please, go full on Socrates and knock holes in it. ;-)

ἔοικα γοῦν τούτου γε σμικρῷ τινι αὐτῷ τούτῳ σοφώτερος εἶναι, ὅτι ἃ μὴ οἶδα οὐδὲ οἴομαι εἰδέναι

;-)

tjguk:
[Annoyingly github doesn't seem to ingest email replies to PR emails. This was my response yesterday]

According to my proposal above: yes, it is a possibility [that a file can be opened in UTF8 and be garbled]. Of course, the autosave could do what these things sometimes do elsewhere: keep a shadow copy until you exit and then copy back. (Maybe that's what the atomic save did? Haven't checked, sorry). But that still leaves the user with no way of backing out in the situation you describe.

One suggestion: if, henceforth, we put an encoding cookie at the top of the Python file (not a bad practice anyway) and use UTF-8, but any file we open without a cookie, we do as present and use the Python default for that platform as we'd have done in the past. When we save, we go UTf-8 with the cookie. Does that work?

carlosperate:
@ntoll Thanks for the autosave explanation, given that I completely agree with the current implementation. However in this case:

> Turning to your question, if you're a user opening non-UTF-8 files with Mu, then the possibilities you outline exist. However, I would argue that you shouldn't be opening non-UTF-8 files that could be corrupted in this way. That you're doing this in the first place suggests you're advanced enough a user to be using more than one development tool, and secondly, that you know enough about the different encodings to be using different ones in different contexts.

I was more concern about the default encodings on default applications. For instance in a foreign language version of Windows ([win code pages](https://en.wikipedia.org/wiki/Code_page#Microsoft_code_pages)) a user could have written a python script with something like notepad in a completely unexpected encoding and the user would be completely unaware of what that is.

Maybe if there is a way to detect if the opened script is UTF-8 (or compatible) the user could be warned that the data could be garbled and at that point do something to mitigate the risk? Could be just a confirmation if they want to open it, create a file backup, or something else.

> One suggestion: if, henceforth, we put an encoding cookie at the top of the Python file (not a bad practice anyway) and use UTF-8, but any file we open without a cookie, we do as present and use the Python default for that platform as we'd have done in the past. When we save, we go UTf-8 with the cookie. Does that work?

Is a cookie the "shebang" encoding at the top of the file (`# -*- encoding: <encoding name> -*-`)?
That would be a good solution too, hopefully there is an internal python module capable to decode that.

tjguk:
I'll raise a separate PR to offer a solution with the encoding cookie (that thing at the top of the of file). I'll outline the logic over there.

I'm reluctant to move into probabilistic assessment of encodings. I think a warning at this level is an attractive nuisance. It's rather like "Invalid Certificate" warnings on websites: they're meaningful, but most people don't know what to do with them. I would be happy with detecting UTF8/16 BOM, as Python does, and I propose to include that logic in my PR.

open() without an encoding specified uses locale.getpreferredencoding() [*] which ultimately calls GetLocalInfo which on my Windows gives cp1252. A quick experiment suggests that notepad effectively does the same. This is basically saying that, up to now, we've allowed Windows to pick the encoding as it does for other programs.

[*] Thanks, Victor Stinner -- I spent ages trying to track this down in the CPYthon implementation -- http://vstinner.readthedocs.io/unicode.html#python-3

tjguk:
Superseded by #371

PR #366, opened 2018-03-03
Fix macOS Travis builds

carlosperate:
The travis builds were failing, this was mostly because pygame is only available as a mac wheel for Python 3.6, and the macOS installation script was still using Python 3.5.

This PR updates that, and a few other small things around that needed updating to keep up with Mu dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ matrix:
       sudo: required
       language: generic
 
-    # To maximise compatibility pick earliest image, 10.9 to be removed on 20/01/2017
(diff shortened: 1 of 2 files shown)


ntoll:
Stonking stuff! Thank you! :-)

See #367 for OSX installation target tracking (PyInstaller appears to be a no-go).

carlosperate:
Thanks @ntoll!
Yes, I just left the PyInstaller there for the time being until a new packaging method is implemented :)

PR #368, opened 2018-03-04
Replace Pygame Zero icon with current logo

lordmauve:
The icon used for Pygame Zero in the project type switcher is a sprite from Kenney.nl.

This replaces it with the Pygame Zero logo, now it has one.


ntoll:
Thank you! I did wonder... I like how the new logo matches the icon I've used for the "play" button. ;-)

lordmauve:
BTW, instant feedback - I spent ages looking for how to get into the project type chooser dialog.

ntoll:
Yeah... that's something that needs explaining up front (and will be in the docs on the website).

lordmauve:
I was thinking it should perhaps be a big round icon.

ntoll:
Sure, but the thing is, everything in the gutter in Mu is about how Mu works. Everything that's a big button is about what you want to do (with Mu), IYSWIM.

Issue #370, opened 2018-03-05
Check for UTF-8 encoding cookie


tjguk:
This issue is a spin-off from issue #360 / PR #364 where the question arose of what to do if Mu tried to read a file in a different encoding from expected. Currently Mu uses open() without a specific encoding when reading and writing files. Python will call locale.getpreferredencoding() to determining the encoding for the file contents. On Windows this calls the GetLocaleInfo API; presumably it does something equivalent on other platforms.

This will raise a UnicodeEncodeError and lose the text if a file is saved or autosaved which contains a codepoint not valid with that encoding. For example, on my (UK-based) Windows, the default encoding is cp1252 which is essentially a variant of iso-8859-1 / latin-1. Putting μ ("Greek Small Letter Mu" U+03BF) into a document will cause it to crash and burn as soon as the autosave kicks in since that codepoint won't encode in cp1252. [And, yes, the μ character was chosen for maximum εἰρωνεία]

The proposal in #364 is to load and save everything as UTF-8 regardless, but this raises the question of what to do when text is encountered which either does not decode via UTF-8 or which produces mojibake.

I propose here that Mu do the following, similar to Python's own file-opening semantics.

When reading a file:

* When reading a file, check for a valid UTF8/UTF16 BOM. If it is found, decode using the appropriate codec
* If no BOM, see if the first line has a PEP 263 encoding cookie. If it does, decode using the appropriate codec
* If there's no BOM and no encoding cookie, use an unadorned open() which use the underlying locale encoding as it always has done

When saving/autosaving a file:

* Add a UTF-8 encoding cookie if it's not already present
* Encode the file contents as UTF-8

This should cover us for both uses: where a new-style Mu file is found (UTF-8 encoded); and where an old-style Mu file is found (locale-encoded). If it's neither of those, Mu will behave no worse than it did previously: attempting to read with the locale encoding and either failing or producing mojibake if the encodings don't match.

If we go with this I would leave PEP 263 cookie present in the top line. If other proposals go ahead to remove comment lines, eg, for micro:bit, these lines will also be removed.

tjguk:
[...]

PR #371, opened 2018-03-06
Issue370 encoding cookie

tjguk:
DO NOT MERGE

This is a PoC PR to test the approach advocated in #370 

There are a number of unresolved issues, including:

(My answers in square brackets)

* Do we inject a missing encoding cookie when we load? [Yes]
* Do we honour/retain an existing encoding after loading? [No]
* Does our [New] file become the single-line encoding cookie? [Not sure]
* Do we add an explanatory comment to the end of the cookie line? [Yes]
* What do we put on the error dialog?
* What happens when we extract a hex file? [I've left it untouched for now; not sure]
* What's the internal newline policy when manipulating the text [\n -- think it's worth specifying that somewhere; I note the newline='' from #133]

I've also added a "design note" under the new design folder in docs/. It's currently orphaned, but it's something I mooted a while ago and never got around to pursuing. If people like the idea, I'll add a design.rst to collect them together.


ntoll:
This pull request **introduces 1 alert** when merging cd2b1cace8712df272c8cf273f5f902ed8d489a7 into 8236b4b9e9f4c13db98e166fcb3b4ebfd1d2567b - [view on lgtm.com](https://lgtm.com/projects/g/mu-editor/mu/rev/pr-679f9db8dd669937b2bd81aa8e19277f4975d0c9)

**new alerts:**

* 1 for Encoding error

---

*Comment posted by [lgtm.com](https://lgtm.com)*

carlosperate:
Haven't got a chance yet to look into the code or the CI failures, but if you are printing or displaying any type of unicode on AppVeyor is likely you'll have to add before running the tests:
```
- cmd: chcp 65001
```

tjguk:
Thanks, @carlosperate . I'd leave it for now; I'm still working through issues. I pushed it just so others could see the approach / progress.

mu/__init__.py
```diff
@@ -1 +1,17 @@
+import gettext
+import locale
+import os
+
+# Configure locale and language
+# Define where the translation assets are to be found.
+localedir = os.path.join('mu', 'locale')
```
ntoll: This needs to be the same as what is currently (in master) in app.py:

```localedir = os.path.abspath(os.path.join(os.path.dirname(__file__), 'locale'))```

ntoll:
Wow... this is great work. Please do say when it's ready for review etc... :-)

tjguk:
So obviously this PR has sprawled somewhat. I've yet to devise tests (or to fix those which now fail because of the injection of the encoding cookie). I've raised a couple separate issues to highlight issues which I've attempted to address in the course of the PR

Issue #380 -- standardising on \n for newlines internally to Mu while honouring the original data
Issue #379 -- refactoring gettext support so that logic.py can be imported on its own 

I still have unresolved questions from above, and I'm conscious that this has turned into a set of changes which are likely to have widespread impact, as they address loading  & saving of files. I haven't, for example, considered the effect on embedded boards where we're packaging and sending code down the wire.

My next step will be to add and fixup testing so we have a clean starting point for discussion, but please feel free to comment and/or challenge the approaches I've taken here. @carlosperate I'd like to hear from you especially on the line-ending change as I wasn't 100% sure of of the motivation behind your previous PR which used newline=""

mu/__init__.py
tjguk: Thanks; fixed

tjguk:
So testing just got a bit more complicated. Because I'm opening the file twice, once in binary, once in text mode, the current mock_open framework can't cope. (Took me about 20 mins to work out why my open(..., "rb") was not returning bytes!

I remember discussing before the question of mocking, tho' I can't find the conversation anywhere. For now, I'm going to try for a filesystem-based test for a couple of tests, leaving the mocks in place for the Qt stuff. If that is gives a cleaner result, I'll move to leave it in. Let's see..

tjguk:
This got mired in push races and merge conflicts, so I'm replacing it by several, smaller PRs including #389 and #390 with more to come.

PR #374, opened 2018-03-07
add chinse translation to appveyor branch

yonghuming:
hi, @ntoll 
i am a teacher teaching kids to code in china and i want to teach them pygame zero as begainning.
but the latest build for windows does not contain chinese so i want help.

hope the newest build contin Chinese.

also i think if all the mode could be install and uninstall like vscode plugin will be ok?

also i find installing the latest version of mu on windows will cause qt could not find error
image


ntoll:
Hmmm... I'll check this. I must have made a mistake or misconfigured something. I thought your amazing work for Chinese support in Mu was already in the installer. In any case, it is important it works and I'll check things and try to fix it! :-)

carlosperate:
@ntoll I've just noticed this has been merged into the `appveyor` branch. Is this already in master? Or do we need to merge it there as well?

ntoll:
@carlosperate this will need to be revisited since the text for the UI will have changed in minor ways.

Issue #356, opened 2018-02-28
Fix OSX themes


ntoll:
The themes don't update the background colour of the button bar when in OSX (works fine in Windows, Linux and RPi).

carlosperate:
Based on the comment from https://github.com/mu-editor/mu/pull/382#issuecomment-373044893 looks like this has been fixed.

PR #382, opened 2018-03-11
Constancy between the themes

ZanderBrown:
Currently the UI jumps around and behaves differently in different themes, this brings consistency between the themes

Unfortunately there doesn't seem to be a way to style the margin without overly complex code (isn't visible to CSS :disappointed: )
peek 2018-03-11 11-47

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ __pycache__
 # Distribution / packaging
 .Python
 env/
+.env/
 build/
 develop-eggs/
 dist/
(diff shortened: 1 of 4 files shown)


ZanderBrown:
WIth the "stuff mu does on top how mu does it below" philosophy we seem to be working with perhaps theme should move to the statusbar?

ZanderBrown:
What do people thing?

ntoll:
Looks good to me... I also tested it on OSX and it fixes #356

Issue #258, opened 2017-03-25
Remove comments before flashing program


webmaster442:
Comments should be removed before flashing program to the microbit, to save space.

ZanderBrown:
This has been brought up before.

Is it something you could add in a PR?

carlosperate:
I'd be against removing comments unless is absolutely necessary, and even when doing so there should be a warning to the user of what's about to happen. This is because the script can be recovered from the .hex file output.
This is also discussed in https://github.com/mu-editor/mu/issues/192.

kamar535:
Just got hit by this when developing an assignment for my students. In the source I've have plenty of comments for documenting the API's of functions etc.

rhubarbdog:
write a sed script to rip comments out `sed 's/\( *#.*\)//' my-file.py > comment-free.py`

kamar535:
@rhubarbdog The comments serves a purpose during reading/editing the source and I would prefer if Mu had an option to remove the comments prior to flashing the program to the Microbit.  Using an external tool, for example sed, to remove the comments before loading the file in Mu is not an ideal workflow for (novice) students.

carlosperate:
The master branch now offers the option in the settings to minify the micro:bit Python code (added in https://github.com/mu-editor/mu/commit/bf8bb7586e37f51e3f7434f17bb56e996eaf1f28), and will be included in the next release of Mu:
<img width="727" alt="image" src="https://user-images.githubusercontent.com/4189262/38954330-3d8f481a-4349-11e8-8f64-9f1298e503c0.png">
I'll close this issue, but please feel free to reopen if you have any questions or suggestions.

kamar535:
This is great!

PR #385, opened 2018-03-11
Microbit/minifyscript

### LINKED ISSUE #192: Idea for saving some script space
Nicholas and I met 'in a pub' for 'a pythonic meet up' last night. I forget who had the idea (we might have had it simultaneously, there was definitely a synchronised 'light bulb moment'!!

Regarding squeezing more python script source code into the limited 8K space on the micro:bit - the idea was that if you look at a typical script, where does a sizeable portion of the flash go to, and is there any low-hanging-fruit to save a bit of space by simple script compression?

Then we realised that there are a lot of spaces in Python source - each indent will probably be 4 spaces, and more than half of a typical large script will have at least one indent, possibly two or three, at the start of most lines. Each indent takes 4 spaces (so 4 bytes of flash).

e.g. in this 217 line script: https://github.com/whaleygeek/microbit_python/blob/master/connect4x4/connect4x4.py

There are easily 143 lines with an indent, and many lines with 2 or 3 levels of indent.

So, the idea would be for the editor (e.g. Mu) to compress all indent spaces at the left of any line to 1 space per indent - it would still be syntactically correct and semantically identical to the Python interpreter running on the micro:bit.

Even at 4 chars per indented line, that would be a saving of 429 bytes. But it is better than that, as many lines have multiple indents (and each additional compressed indent level will save another 3 bytes of flash).

If a .hex is opened by Mu or other editor, it could turn indent spaces at the start of the line back to multiple spaces when re-loading into the editor.

As a nice touch, there could be an option to put a magic comment at the top of the file to specify the preferred indent size, so that when the editor re-loads any script, it would know the user preference for spaces per indent.

Clearly it would still have to work with loading a .hex file that has a Python script with multiple spaces per indent.

But in the common case, quite a few bytes of flash could be stored in the persisted version of the script, for what looks like a relatively simple algorithmic compression process that would easily be possible inside the editor (and also inside a web editor) when serialising the script into a .hex. Before hexlifying the script, it could be put through an indent compression step. The compression algorithm is probably streamable, meaning you could probably compress and hexlify 'on the fly'.

ZanderBrown:
On the whole this should resolve #192 (and #258), I've decided to leave whitespace mangling to someone else

Hopefully it's clear what is going on here and this should be far more bulletproof than a regex-based approach


ntoll:
@ZanderBrown this is great stuff! However, I think it could be in the wrong place. Why not add this to uflash (so all apps that re-use this module get to benefit from your wonderful work)? AFAICT this should be easy to copy over and means the changes to Mu's codebase may only be a change to the way it calls into `uflash`.

See https://github.com/ntoll/uflash/issues/44 for the ticket in uflash to track this work. :-)

Thank you!

carlosperate:
Great addition @ZanderBrown, I'm sure a lot of people will find this very useful! I also agree with @ntoll in that this would be able to reach more people/projects in uFlash as well.

And since we are refactoring that portion of the code in Mu, the script length should probably come from uFlash as well rather than being a magic number in Mu.

carlosperate:
Also, an extra step can be done to first convert all `\r\n`end lines to just `\n`. That should save a few extra characters.

ntoll:
@carlosperate good catch on the hard-coded number in Mu. This should derived from whatever is in uflash. ;-)

ZanderBrown:
Sure will take a look at moving this to uFlash (may take me a few days)

Definitely makes sense to move all magic numbers too uFlash, as for `\r\n` i think tokenize handles that internally but I'll investigate

ntoll:
Good stuff! :-)

ZanderBrown:
So I've decided to split the logic out into its own module [nudatus](https://pypi.org/project/nudatus/) so other micropython related projects can decide to probably not use it

Will be making a PR to uflash soon but this will need to thoroughly tested to make sure I'm not breaking peoples scripts

ZanderBrown:
As seems inevitable a bug around the handling of dicts appears just after I publish, does anyone have/know of a script that demonstrates all python syntax?

ZanderBrown:
uFlash PR: https://github.com/ntoll/uflash/pull/45

ZanderBrown:
ntoll/uflash#45 has been merged, once a new uFlash makes its way into Mu I'll workup some nudatus integration for Mu unless @ntoll wants to do that

ntoll:
I'll pick this up, since I have a good idea where this could be configured (based on some work I'm doing right now).

ZanderBrown:
Don't want to be a nuisance but hows this going?

ntoll:
On my to-do list. But the initial work for the "admin" tabs (replacing the logs) is in master. Just need to add a micro:bit tab for minification setting (among other things).

carlosperate:
Looks like this feature is fully implemented now in https://github.com/mu-editor/mu/commit/bf8bb7586e37f51e3f7434f17bb56e996eaf1f28 :)

ZanderBrown:
Yes & No?

It seems @ntoll 's logic is only considering minification after long scripts have already been rejected, meaning scripts that where already small enough will be minified unnecessarily and those that could benefit are rejected already

https://github.com/mu-editor/mu/blob/bf8bb7586e37f51e3f7434f17bb56e996eaf1f28/mu/modes/microbit.py#L238

vs 

https://github.com/mu-editor/mu/blob/bf8bb7586e37f51e3f7434f17bb56e996eaf1f28/mu/modes/microbit.py#L265

I think roughly what we want is

```
if script < max or (minify and minified < max)
    flash
else if not minify
    alert "To  big try minify"
else
    alert "Too big"
```

ntoll:
Doh... :man_facepalming: 

All PR's gratefully received and reviewed on Monday! :-)

ntoll:
Closing this PR since it's been superseded by @ZanderBrown's awesome nudatus work.

PR #386, opened 2018-03-12
CI Linux: Enable tests on the Linux Travis builds

carlosperate:
This has been an itch I've been dying to scratch for quite a long time. Fixes https://github.com/mu-editor/mu/issues/244 and the reason why we coudn't run the tests on Linux Travis.

Turns out Qt was failing during launch/object initialisation in odd ways when running headless in Travis (in retrospect that's one of the first things I should have checked, I've tripped over the same stone before...) and pytest swallowing the stack trace didn't help either. But anyways, using xvfb on Linux is all that's needed.

There is a bit of a clean up and I've also moved from setting up our own Python environment to use the ones provided by Travis itself, so with Linux we are testing on Python 3.5 and 3.6.


ntoll:
Thank you!

PR #389, opened 2018-03-16
Refactor i18n

tjguk:
This is the small refactor of the i18n / gettext support so that it's part of the package import. Apart from a bit of tidiness, this means:

* it no longer needs to be imported and initalised in app.py
* logic.py can be imported independently (useful for testing); previously it implicitly relied on *something* importing gettext to start with and therefore...
* conftest.py no longer needs to set up gettext for the tests to work


tests/test_logic.py
```diff
@@ -1483,3 +1484,19 @@ def test_rename_tab_avoid_duplicating_other_tab_name():
                                               'A file of that name is already '
                                               'open in Mu.')
     assert mock_tab.path == 'old.py'
+
+def test_logic_independent_import_logic():
```
carlosperate: Looks like pystyle is complaining about not having 2 blank lines between functions: `E302 expected 2 blank lines, found 1`.
Same for the one below.

tjguk: (Sigh; once again a reply hasn't made it through to GH)

Thanks; good catch. In short: I didn't run make check. I'll go back and rejig.

BTW I've never encountered a codebase which actually puts two blank lines between functions (except possibly to mollify nagging linters...)

ntoll:
Thank you. I fixed the style issue before merging. ;-)

You mention not seeing the two line convention... These past few years, I've seen nothing *but* this convention. C'est la vie.

tjguk:
Thanks

PR #390, opened 2018-03-16
Support for encodings

tjguk:
This support adds testing & functional support for reading & writing Mu files with encodings and line-endings. But it does not actually integrate that functionality in the current Mu codebase. With the pace of change in the codebase at the moment I found myself mired in push races and merge conflicts. So this does the minimum possible to get the support in now, ready for the more intrusive changes afterwards.

This will supersede #370


tjguk:
Hold until I run and sort out make check

ntoll:
@tjguk this is great..! Please do let me know when I can check/merge.

You mention more intrusive upcoming changes. What might they be?

ntoll:
Hi @tjguk, could the tests in `test_logic_read_write` move to `test_logic`..? The convention used so far is to put tests pertaining to a certain file `foo.py` into `test_foo.py`. I realise this means `test_foo.py` may become large, but at least I know where to look for tests since there's a convention that's always followed (and I can use my editor's find functionality). Sorry about that... I should have made that clear. Once fixed I'm happy to merge this.

tjguk:
Certainly. The reasons for the move were twofold: one was, as you suggest, to prevent test_logic from becoming even bigger. But the principal one was because of the conflicts I'd experienced the previous time round. I'll bring them back in now.

Meanwhile: can you see what Appveyor is complaining about? I can't.

ntoll:
Thank you! :-)

Appveyor shouldn't complain now... I fixed the problem in master.

PR #394, opened 2018-03-17
CI Windows: Clean up AppVeyor file, reduce make checks 

carlosperate:
_**Edit:** The text below from the original PR does no longer apply to the final squashed commit._

***
MinGW, along many other things, adds a POSIX layer on top of Windows, so we need to run the tests more "natively" as well without it.

This PR adds a pytest call after the `make check` and cleans up very minimally the AppVeyor script.

Apart from that I've removed the `make check` that runs as part of every `make win32` and `make win64`, as otherwise we have the same checks running 4 times on every CI build. If you think that should still be there anyway let me know and I'll revert it.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,9 +13,6 @@ init:
   - cmd: ECHO Processor architecture - %PROCESSOR_ARCHITECTURE%
   - cmd: wmic OS get OSArchitecture
 
-  # Utilise pre-installed MinGw for make
-  - ps: $env:PATH = 'C:\MinGW\msys\1.0\bin;C:\MinGW\mingw32\bin;C:\MinGW\bin;' + $env:PATH
-
   # As AppVeyor has multiple python install, check which one uses by default
   - cmd: ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH%
   - cmd: python --version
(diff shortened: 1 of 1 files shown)


carlosperate:
Actually, this might not be needed at all, the tests on AppVeyor are running on the make.cmd file. Not quite sure why it's not the same on my dev environment :/

carlosperate:
Okay, so taking in consideration AppVeyor is definitely running the make.cmd instead of make on MinGW I have reverted the unnecesary addition of this line in the appveyor file:
```
- cmd: pytest --cov-config .coveragerc --cov-report term-missing --cov=mu tests/
```
I've also removed the unnecesary MinGW configuration, and kept the other AppVeyor script clean up and removal of the 'make check' that runs on every 'make win32' and 'make win64' (as before, if you prefer to have them let me know and I'll revert that as well).

Since I've rebased, squashed and ammended, the history shown in the conversation view of the PR will be a bit wonky, but the PR itself is just a single commit with what I've descripbed in this message.

ntoll:
@carlosperate this looks good except for one thing: you're right to suspect I'd prefer to keep `make check` whenever `make win32` and `make win64` are run. This ensures it's impossible to use `make` to create installable assets that fail the test suite. I realise that it means the builds take slightly longer, but since they're automated and happen "in the cloud" I'm not too fussed if they take a little longer than they might do. I just don't want us to automatically build broken assets. ;-)

If you could fix these changes up then I'll merge this PR.

Thank you! :-)

carlosperate:
Thanks @ntoll, I've updated the commit to only include the small clean up in the appveyor file.

ntoll:
Thanks @carlosperate!

PR #406, opened 2018-03-22
Drop files

### LINKED ISSUE #74: Drop file on support
It would be great to be able to simply drop a .py file onto the application in the menu bar. Although not a majorly important feature, it would be great to see it there. Have had a few times so far where I have gone to drop a script on it and then remembered I can't just yet.

ZanderBrown:
Fix #74 (only 2 years late...)

If a file is dropped on the editor (doesn't listen on panels) a signal is bubbled up and then handled by opening the file

However the signal is only fired if the dropped file is `*.py or *.hex` (case insensitive) for everything else the current behaviour is preserved

Also included is a styling change following on comments on 51847232b37f7266e2fbb4108da23d9ebf64c4b0
(Yes i know it shouldn't really be in this PR)


ZanderBrown:
Not sure what Travis doesn't like here

mu/interface/editor.py
```diff
@@ -86,6 +90,44 @@ def __init__(self, path, text, newline=NEWLINE):
...
+            links = []
+            # Iterate over each of the urls attached to the event
+            for url in event.mimeData().urls():
+                # Check the url is to a local file
+                # (not a webpage for example)
+                if url.isLocalFile():
+                    # Grab a 'real' path from the url
+                    path = url.toLocalFile()
+                    # If the dropped file is a script or hex
+                    if os.path.splitext(path)[1].lower() in ['.py', '.hex']:
```
carlosperate: Is there any disadvantage in letting all files pass through, and then have the `Editor._load` decide which files to process? This way there is also an informative message box saying if the file is not compatible.

ntoll: +1 to re-using existing functionality for this.

ntoll: OK... I'll fix this to use existing functionality.

ntoll:
Thank you for this! :-)

Issue #409, opened 2018-03-26
Specifying a file on the command-line can duplicate if it's also in the restored session


tjguk:
If you specify a file on the command line (eg mu myfile.py) and that file is also stored in the session settings, it will be loaded twice.

ntoll:
good catch

ntoll:
Hmm... I've just tested this and everything works as expected on Linux (I even get told that `my_file.py` is already open). Could this be Windows specific..?

ZanderBrown:
Perhaps it's a problem of relative vs absolute paths?
https://github.com/mu-editor/mu/blob/40fbd75c46f5ae10c458fb478bf5b0af98d1416c/mu/logic.py#L647
As far as I can tell we never 'normalise' path to be absolute, if that's the case it would explain oddities like this

ntoll:
I thought of that... the `session.json` file stores the paths as absolute. I was starting Mu with a relative path. Having said that, I think you're right and I've just not found the right edge case. Given that logic.py's `restore_session` is dealing with absolute paths from the session.json file then if we normalise the `passed_filename` this problem shouldn't happen. This should be a quick fix in app.py.

ntoll:
@tjguk doe 297669a9cbebdeb8991232995d524e6b80cf47e2 fix your problem?

tjguk:
Haven't tried yet, but since I was able to reproduce without even having a file on the command line, here's the contents of my session file:

```
{
  "theme": "day",
  "mode": "python",
  "paths": [
    "C:/Users/goldent/mu_code/t.py",
    "C:\\Users\\goldent\\mu_code\\t.py"
  ]
}
```

ntoll:
hah... it looks like Window paths strike again. What would be really useful is to understand how each of the paths was created. I'm assuming the second path is the absolute path of the file you specified on the command line? In which case, something (QWidget) is storing paths in a way that's more like a URL rather than an OS specific filesystem path. We should choose one (and my vote would be for the URL like convention -- in which case, how do I replace abspath with the right thing?).

tjguk:
os.abspath (which calls os.normpath) on Windows will produce a path of the form c:\users\goldent\... from either of the two forms above.

tjguk:
With your patch applied, I see the following behaviour:

* settings.json contains "C:/Users/goldent/mu_code/t.py"
* python -mmu t.py
* t.py is opened successfully in the editor pane, but an alert appears "The file t.py is already open"

[...]

PR #411, opened 2018-03-27
Impove handling of cli files

ZanderBrown:
Further work for #409 

Whilst this probably does resolve that issue it does resolve an issue that issues caused so-to-speak

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -104,9 +104,16 @@ def run():
     # Setup the window.
     editor_window.closeEvent = editor.quit
     editor_window.setup(editor.debug_toggle_breakpoint, editor.theme)
-    # capture the filename passed by the os, if there was one
-    passed_filename = sys.argv[1] if len(sys.argv) > 1 else None
-    editor.restore_session(os.path.abspath(passed_filename))
+    # Restore the previous session
+    editor.restore_session()
+    # capture the filename passed by the os, if there were any
+    for passed_filename in sys.argv[1:]:
+        try:
+            # abspath will fail for non-paths
+            editor.direct_load(os.path.abspath(passed_filename))
+        except:
+            logging.warning('Can\'t open file from command line {}'.
+                            format(passed_filename))
     # Connect the various UI elements in the window to the editor.
     editor_window.connect_tab_rename(editor.rename_tab, 'Ctrl+Shift+S')
     status_bar = editor_window.status_bar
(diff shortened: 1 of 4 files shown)


mu/app.py
```diff
@@ -104,9 +104,18 @@ def run():
...
-    # capture the filename passed by the os, if there was one
-    passed_filename = sys.argv[1] if len(sys.argv) > 1 else None
-    editor.restore_session(os.path.abspath(passed_filename))
+    # Restore the previous session
+    editor.restore_session(without_empty=len(sys.argv[1:]) > 0)
+    # capture the filename passed by the os, if there were any
+    for passed_filename in sys.argv[1:]:
+        try:
+            # abspath will fail for non-paths
+            editor.direct_load(os.path.abspath(passed_filename))
```
ntoll: hmmm.... I'm not sure about this approach. The code in `app.py` is essentially there to instantiate things and connect them together. This feels like logic that shouldn't be here (i.e. it should be within the editor, and all the editor needs are a bunch of paths for the passed in files). Does this make sense?

ZanderBrown: Yep, I'll have a look at rearranging that

ZanderBrown:
Sorry that took a while

mu/app.py
```diff
@@ -104,9 +104,10 @@ def run():
     # Setup the window.
     editor_window.closeEvent = editor.quit
     editor_window.setup(editor.debug_toggle_breakpoint, editor.theme)
-    # capture the filename passed by the os, if there was one
-    passed_filename = sys.argv[1] if len(sys.argv) > 1 else None
-    editor.restore_session(os.path.abspath(passed_filename))
+    # Restore the previous session
+    editor.restore_session(without_empty=len(sys.argv[1:]) > 0)
```
ntoll: Hmmm... so why pass this flag at all. Why not pass in `sys.argv[1:]` and let `restore_session` handle it..? Happy to be talked round, but it feels like there should be one call to `restore_session` with the passed in files and then it'll work out what to do with it given the previous state of the session and the new files. Does that make sense?

ntoll:
No need to apologies for anything @ZanderBrown... I'm just incredibly grateful that people are taking the time to improve Mu. :-)

mu/app.py
ZanderBrown: I had them separate in case we wanted to do single instance things but I'll move the call inside `restore_session`

ntoll:
Thanks! :-)

PR #412, opened 2018-03-27
Open logging for output as utf-8

tjguk:
The entire contents of the editor is written out via logging.debug() at various points. Previously, no encoding was specified when opening the logger; this will use the locale default which (on my Western Windows machine) won't encode, eg, Chinese characters.

This PR specifies the Mu default encoding (currently utf-8) when logging


ZanderBrown:
utf-8 all the things!

ntoll:
Thanks for this @tjguk -- on Linux this wasn't a problem, perhaps due to different encoding setting between OS's, so great catch there for Windows users. ;-)

PR #413, opened 2018-03-28
Style context menus

ZanderBrown:
This is particularly important for the high contrast theme

At the moment there is no distinction between items and those that are disabled and the user has no indication of the hovered item (i.e. the clickyness)

Current master:
peek 2018-03-28 11-12

Proposal:
peek 2018-03-28 11-13

I seem to be generating a lot of PRs lately, sorry if I'm filling your inbox :smile:

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -25,7 +25,7 @@ import platform
 import pkgutil
 import logging
 from logging.handlers import TimedRotatingFileHandler
-from PyQt5.QtCore import QTimer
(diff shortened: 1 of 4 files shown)


ntoll:
Thanks... this is great... I made two adjustments...

* Fix the tests
* Add a higher contrast to the menu border in "dark" mode.

:-)

ZanderBrown:
Did I break a test?

Didn't think to check as it's majority CSS

PR #418, opened 2018-03-31
Rethink cookies

tjguk:
In lieu of always writing an encoding cookie, Mu will now preserve and honour and existing cookies (ie the encoding specified in the cookie will be used when writing the data out). But no new cookie will be added.

For reading, Mu will continue to pick up obvious clues (BOM, encoding cookies) and decode appropriately. But in their absence it will try first UTF-8 and then the locale default. If that fails an error popup will be displayed as before indicating which codecs were tried.

Absent a specific encoding cookie (which can be added mid-session) the file will be saved encoded as UTF-8


ntoll:
@tjguk I've just had a look through the code and this looks good. I notice there's a PEP8 error when you `make check` on Travis so please appease the PEP8 god with an appropriate sacrifice. ;-)

I'll be back on Wednesday, so if marked as "merge-able" will do an in-depth code review across available platforms  upon my return.

As always, thank you for your invaluable work! :-)

tjguk:
So there's a corner-case which I don't quite know how to handle. If a file in the editor has an encoding cookie (from the initial load or added later; doesn't matter) and now contains characters which can't be encoded with that codec, we get a crash.

ie:
1) Create an empty file
2) Add an iso-8859-1 cookie; # -*- coding: iso-8859-1 -*-
3) Add a line of Chinese characters: 恭喜發財
4) Save

... and crash.

The reason for the crash is clear (that codec can't encode those characters). The problem is that I'm not sure how best to handle it. Options include:

a) Predict or trap the exception and raise a message box suggesting that the encoding cookie be removed
b) Override the cookie and save as utf-8 (possibly manipulating the cookie as we did before or just remove it)

There's even a variant of this which I hit while testing:

1) Create an empty file
2) Start to type an encoding cookie (but don't complete it)
3) When the autosave kicks in, it will crash because it attempts to save with an invalid encoding

At this stage, you're all thinking: this encoding cookie thing is more trouble than it's worth. And you might be right. Especially since the default source encoding for Python 3.x does seem to be utf-8. Really the only thing stopping me from ditching the support for it is that I don't like the idea of loading a file which has such a cookie and then removing it.

carlosperate:
> a) Predict or trap the exception and raise a message box suggesting that the encoding cookie be removed

That sounds like the most sensible option to me.

ZanderBrown:
:+1: for **a**, We are trying to avoid 'magic' behaviour (hence this issue) and **b** seems a little *magic* to me

ntoll:
Oh boy... this is turning into an EPIC. Thank you for your work on this @tjguk.

Given the "explicit is better than implicit" Zen of Python, I lean towards a) as the simplest, most helpful, least surprising solution.

As for the incomplete encoding cookie problem... can we just save as UTF-8 until a valid cookie is detected..? In any case, as you point out, UTF-8 is now the default source encoding in Python 3.x. If a Mu user is typing in their own encoding cookies by hand, they deserve all the associated pain they bring on themselves... :-) Mu is, after all, a beginner's editor. :-P

Do let me know when you think this ready for review / merging.

tjguk:
Looks like (a) is the People's Choice, so I'll put something in place for that. Meanwhile I agree with @ntoll suggestion of just saving utf-8 if the codec is unrecognised (for any reason). Should we flag a warning anywhere? At least in the log, I imagine.

ntoll:
+1 to flagging corrupt cookies in the log.

(I can't believe I just typed that sentence...)

tjguk:
Ok I think we're ready for review and merging now. All the tests / checks pass and I've run it backwards and forwards a few times through the front-end. It's not always pretty: if you get caught with a half-baked cookie when autosave kicks in, you get an awkward popup which -- if you're slow enough -- will pop up again seconds later as the *next* autosave kicks in. But this situation is already fairly niche, and this just makes it nichier (?).

Where I was changing anyway, I've very slightly reworked some of the exception handling to make it tighter around the affected areas. I've also used logger.exception inside an exception handler where the existing code had logger.error. If the house style is to avoid logger.exception then let me know.

ZanderBrown:
> Sod's law is a name for the axiom that "if something can go wrong, it will", with the further addendum, in British culture, borrowed from Finagle's law, that it will happen at "the worst possible time". This may simply be construed, again in British culture, as "hope for the best, expect the worst".

tjguk:
Odd... Travis is throwing pep8 errors but I ran a full make check before pushing. Are the settings different?

ZanderBrown:
Travis is just running `make check` so they shouldn't be any different (especially when the complaint is about line-length)

tjguk:
Yeah; looks like I've messed up the Windows make.py -- I think it's only running pycodestyle on make.py itself! Investigating now...

tjguk:
(Made the mistake of switching between two different computers without syncing git properly; hopefully sorted now so the Travis builds might succeed)

ntoll:
OK... this LGTM :-)

I'd feel more comfortable if the `sniff_*` functions had specific unit tests. I'll add them and then merge.

ntoll:
Hurrah.

PR #419, opened 2018-04-01
set a default help url for prereleases

willingc:
This PR fixes the case where Help opens a not found page when the mu version is a beta or prerelease.

Checks if there is a suffix beyond the major and minor version, if so open the default latest help page. Otherwise, open the language and release specific help.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -872,7 +872,7 @@ class Editor:
         major_version = '.'.join(__version__.split('.')[:2])
 
         logger.info('Open help URL for {} version {}'.format(language_code,
-            __version__))
+                                                             __version__))
         if (len(__version__.split('.')) > 3):
             # pre-release so show latest
             url = 'https://codewith.mu/help/latest'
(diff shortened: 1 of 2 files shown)


ZanderBrown:
Perhaps there should be a separate help page for prereleases that links to the latest help as well as the GitHub issues page

ntoll:
Hi folks,

@willingc thank you for the contribution. I think, once the new website lands (very soon, see: https://github.com/mu-editor/mu-editor.github.io/tree/v1), the help button will automatically work in the correct way. 

The "latest" path also adds a completely new subdirectory to the website and I'm anxious to put as much of the content under correct language codes (e.g. `/en/tutorials` etc...). To this end, the only page that's not categorized in this way is the root "landing" page (which will also point people at the various translations available).

Does this make sense..?

@ZanderBrown I like your point about adding extra pointers to help because it's a pre-release.

Both, given the closeness of the new website, do you see any advantage to having a "latest"..? If not, perhaps we can mark this as already resolved (pending the website)..?

Happy to be persuaded otherwise. :-)

N.

ZanderBrown:
I don't see the lack of a 'latest' being a problem

willingc:
Cool. Feel free to close. We can always resurrect if it becomes necessary down the road. 👍

PR #422, opened 2018-04-03
searching /sbin/mount in adafruit mode

fmorton:
By default, non-root users do not have "mount" in their path, but can access it using "/sbin/mount". This pull requests first uses "mount", and then if not found, attempts to use "/sbin/mount". Other locations may be easily added to the array of possible locationsl

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -17,6 +17,7 @@ These include (but are not limited to):
 * Tibs / Tony Ibbs (tibs@tonyibbs.co.uk)
 * Zander Brown
 * Alistair Broomhead (alistair.broomhead@gmail.com)
+* Frank Morton (fmorton@mac.com)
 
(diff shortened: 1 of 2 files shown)


ZanderBrown:
What distro is that on? (aka I don't seem to have it)

fmorton:
Make a style change.

ZanderBrown:
I'm confused why was this closed and #423 opened? They look identical.

fmorton:
Sorry for the confusion. I did not do a "make check" on the first pull request and had a styling issue that kept it from passing the tests. So, I closed this one, fixed the styling change (too many spaces in an array definition) and submitted the new pull request.

ZanderBrown:
Just for future reference:
You can push further commits after opening a PR. In fact, you successfully did that here! You may not have noticed but your 3rd commit is listed on this PR after my initial comment as well as being included in your new PR.

PR #423, opened 2018-04-04
searching /sbin/mount in adafruit mode

fmorton:
By default, non-root users on macintosh do not have "mount" in their path, but can access it using "/sbin/mount". This pull requests first uses "mount", and then if not found, attempts to use "/sbin/mount". Other locations may be easily added to the array of possible locations.

I still get one error with "make check", but get the same thing without my changes.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -17,6 +17,7 @@ These include (but are not limited to):
 * Tibs / Tony Ibbs (tibs@tonyibbs.co.uk)
 * Zander Brown
 * Alistair Broomhead (alistair.broomhead@gmail.com)
+* Frank Morton (fmorton@mac.com)
 
 We welcome contributions from anyone! Please see :doc:`contributing` for more
 information.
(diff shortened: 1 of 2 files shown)


ZanderBrown:
LGTM but returning to my initial comment

> What distro is that on? (aka I don't seem to have it)

fmorton:
If I am understanding your question, this is from the current master branch. I forked it yesterday.

ZanderBrown:
Sorry I meant the OS you are using where `mount` can be found at `/sbin/mount`

fmorton:
MacOS High Sierra 10.13.2

ntoll:
Hi @fmorton, thank you for your contribution! I'm not near my Mac at the moment, but I will be tomorrow! I'll review and hopefully merge then.

N.

ntoll:
@fmorton I added a test to bring coverage back up to 100%. Thanks for your contribution. :-)

Issue #397, opened 2018-03-17
macOS tab close button has dissapeared?


carlosperate:
Not sure how this has happened, or if maybe this is only on my current environment. So hoping other macOS users can replicate.

Basically this is the way Mu currently looks for me. The tab doesn't show the close button any more:

<img width="1232" alt="image" src="https://user-images.githubusercontent.com/4189262/37560479-c5c17370-2a30-11e8-8b79-e528340677ff.png">

There is a clear space on the left side of the tab name, and if you click there it will close the tab, but that's it.

Tested on the latest master at the time of writting: 6121350e5261df15cd7542fa938c2e0aab6b4fc2

carlosperate:
Looking at older screenshots from https://github.com/mu-editor/mu/issues/387 this was already an issue for a while:

<img width="998" alt="image" src="https://user-images.githubusercontent.com/4189262/37348992-5193d90e-26cd-11e8-9054-be2f40c82dc4.png">

ZanderBrown:
So clearly I didn't break it :smile:!

Looking at the screenshot it's trying to put the close button on the left, which would make some sense, is there any hover / click sensitivity there at all?

ZanderBrown:
[An upstream bug](https://bugreports.qt.io/browse/QTBUG-61092) Are you running against Qt 5.9.0 perhaps?

carlosperate:
Does that mean that changing the tab mode to "documet mode" would be a work around?
To be honest, after reading the Qt documentation I don't quite understand what that does.

ZanderBrown:
I'm not sure that it'll help with the close button (probably need newer/older Qt5 for that) but it should improve the behaviour of tabs, currently, they are working more like pages in a dialogue that documents on macos

From my non-mac owning or using perspective I believe that property will change the tabs alignment and populate the Window menu

ntoll:
Hmmm... so the problem persists in Qt5.10 on OSX (I just checked). I'll report upstream since they believe it fixed.

ntoll:
FYI: https://bugreports.qt.io/browse/QTBUG-61092?focusedCommentId=396123&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-396123

ntoll:
@ZanderBrown WRT `setDocumentMode(True)`, this does, indeed, make the tabs behave more like on Windows and Linux.

[...]

PR #425, opened 2018-04-05
Specify our own close tab icon

ZanderBrown:
Partly for cross-platform consistency but also an attempt at a workaround for #397 

No idea if it'll work but if anyone has a mac it's worth a shot

Half the Qt documentation later I finally worked out how to reference images form resources, nothing like helpful docs :smile:


ZanderBrown:
I can provide my source SVGs if that'll help but I'm no icon designer

[close.zip](https://github.com/mu-editor/mu/files/1884013/close.zip)

They are drawn at 24x24px but were exported at 16x16px

ntoll:
In the end, after quite a bit of experimentation I've used the "cross" from font awesome (the same source as the icons for the buttons).

ZanderBrown:
Seems a good choice then

willingc:
Nice! Confirming that this works now :-) Thanks @ZanderBrown and @ntoll.

PR #427, opened 2018-04-06
Alter the close icon to be more central

ZanderBrown:
So the close icon was selected from Font Awesome and that's ~awesome~ fine but then I noticed something seemed a bit odd with the alignment.
Eventually I ended up in GIMP at 2,000x and it seems the FA icon is slightly off centre in its box

Unfortunately because I'd noticed this once I continued noticing it so I've done a little tweak so the icon is evenly distributed

This has to be one of the most ridiculous/petty PRs of all time but here it is


ntoll:
Epic. :-)

PR #421, opened 2018-04-03
Edit requirements and app for readability and maintainability

willingc:
@ntoll and folks,

Here's a maintenance PR that I generated while learning the code base. Feel free to close if you don't want style changes.

PR #438, opened 2018-04-12
Update setup.py to be able to install from source on Raspberry Pi

carlosperate:
Also updates the docs as well with the simplified instructions.
No more messing around editing the setup.py, just:
```
sudo apt-get install ...
pip install -r requirements_pi.txt
pip install .
```

The requirements_pi.txt has also been updated to follow the changes in requirements.txt from https://github.com/mu-editor/mu/pull/421.


setup.py
```diff
@@ -10,6 +11,25 @@
...
+try:
+    machine = platform.machine()
+    if machine.lower().startswith('arm'):
+        exclude = ('pyqt5', 'qscintilla', 'qtconsole', 'PyQtChart')
+        install_requires = [requirement for requirement in install_requires
+                            if not requirement.startswith(exclude)]
+except Exception as e:
+    # Something unexpected happened, so simply keep all requires
+    pass
+
```
carlosperate: This is a bit ugly to have in the setup.py, but I found it to be the best solution to dynamically chose what to include in the `install_requires`. Otherwise mu-debug doesn't work.
Using something like extra_requires is a lot messier during installation.

docs/setup.rst
```diff
@@ -46,7 +46,7 @@ used to launch user's Python script with the debugging scaffolding in place to
 communicate with Mu, acting as the debug client). As a result, it's essential
 to run the following to ensure this command is available in your virtualenv::
 
-  python setup.py develop
+  pip install --editable .
```
carlosperate: This is the pip equivalent of `python setup.py develop`.
There might be some subtle differences, but since we are pretty much using pip for every step here I though it might be better to use pip here as well.

carlosperate:
Thanks!

Issue #414, opened 2018-03-28
Application id?


ZanderBrown:
I discovered the other day we've been [packaged for Fedora](https://apps.fedoraproject.org/packages/mu) and of course, we have Debian packaging stuff in the repo already

I know we don't currently need one as such and we decided against putting any effort into #318 but it could be good if we define an application id and name our .desktop file accordingly (and macOS equivalents)

Personally I think `mu.codewith.editor` seems an obvious choice but what do people think

(@ntoll , @carlosperate , @bennuttall  raw thoughts?)

ntoll:
Just about to step onto a train, but is the naming convention is the reverse of a URL, then that looks fine. ;-)

ZanderBrown:
Yes reverse DNS is the normal system

bennuttall:
SGTM :+1:

ZanderBrown:
I have a branch `base/desktop-integration` where I'd started to work with this

The changes being:
* Rename `.desktop`
* Rename icon installed for `.desktop`
* Use setDesktopFileName on the QApplication object

My branch also started some appdata niceties for Linux packaging

ntoll:
Fixed by #440. Thanks @ZanderBrown :-)

PR #440, opened 2018-04-13
Base/desktop integration

ZanderBrown:
Follow up on #414 

`appdata.xml` is going to need some work but is useful for packing on Linux

Sets the WMClass & associates the desktop file with it for some GNOME niceties

Also associates with python3 mimetype so we show up in 'Open With'

All these things are primarily for Linux & tests are broken due to extra QApplication calls


ZanderBrown:
Hopefully that's a little less rough around the edges now

conf/mu.appdata.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+	<id>mu.codewith.editor.desktop</id>
+	<metadata_license>CC0-1.0</metadata_license>
+	<project_license>GPL-3.0</project_license>
+	<name>Mu</name>
+	<summary>A Python editor for beginner programmers</summary>
+	<description>
```
ntoll: I'll update the description once this is landed in master... Thanks! :-)

mu/app.py
```diff
@@ -114,10 +114,17 @@ def run():
 
     # The app object is the application running on your computer.
     app = QApplication(sys.argv)
+    # By default PyQt uses the script name (run.py)
+    app.setApplicationName('mu')
```
ntoll: Capital "M"... I'll fix this once it's in master.

ZanderBrown: No! It's process name/wmclass, it would be capital M for setDisplayName but we aren't using that because it would append '- Mu' to the window title and "Mu - file.py - Mu" would look a little odd :smile: 

Changes to master mean changing this to `mu-editor` would make sense as long as the .desktop WMClass entry is kept in sync

debian/mu.install
```diff
@@ -1,2 +1,2 @@
-conf/mu.desktop usr/share/applications/
-conf/mu.png usr/share/icons/hicolor/256x256/apps/
+conf/mu.codewith.editor usr/share/applications/
```
ZanderBrown: Well that's embarrassing, missed the .desktop here

Oops

mu/app.py
ntoll: Aha... understood. ;-)

debian/mu.install
ntoll: What should it be and I'll update now on master.

ZanderBrown: `conf/mu.codewith.editor usr/share/applications/` -> `conf/mu.codewith.editor.desktop usr/share/applications/`

ntoll: +1

Issue #445, opened 2018-04-18
Feature suggestion: Offer user to change mode when device is connected


microbit-carlos:
When the beta 15 hit the Raspbian repos we had a few reports of users struggling to flash the micro:bit because they were in a different mode without realising it (e.g., they clicked "run" and got a `ModuleNotFoundError` errors).

Beta 15 was released before the "modes" button was added, so having this button in the next version should help users realise they have different modes for different purposes.

However because some users might upgrade Mu without knowing that it has evolved from a "micro:bit editor" to a multipurpose Python editor, we were wondering if there was anything we could do to help with the transition.

Mu currently scans USB devices for anything new connected and prints that info in the log and status bar. What are your thoughts about updating it so that if a micro:bit or CircuitPython device is detected it asks the user if they want to change modes? 

![plugged_ask](https://user-images.githubusercontent.com/29712657/38948394-6337fd08-4337-11e8-9253-b95d349db375.gif)

Something similar could be done as well when loading hex files:

![open_ask_mode](https://user-images.githubusercontent.com/29712657/38948488-b0eb4f78-4337-11e8-82ad-abcbf7320f3b.gif)

If a user plugs a new device, or opens a micro:bit hex file, is very likely that they will want to use Mu in that mode, so even though the pop-up message is a bit more intrusive I think it offers useful information and action that the user likely want to take. What do you think?

Of course, we'd be happy to provide a PR for this or whatever that might come out of this discussion.

ZanderBrown:
I quite like both these proposals but the "we noticed a device" should be fairly easy as we are already polling for devices

microbit-carlos:
Thanks @ZanderBrown, and yes the changes should be relatively simple (proof-of-concept code in https://github.com/microbit-carlos/mu/commits/detect_device fork branch).

ntoll:
+1 LGTM... care to rustle up a PR..? ;-)

To be clear:

* If a new micro:bit is detected and Mu's not in micro:bit mode, it should ask.
* If a new Adafruit board is detected and Mu's not in Adafruit mode, it should ask.
* If both an Adafruit and micro:bit are plugged in, Mu does nothing and assumes the user knows what they're doing.
* Tests!

ntoll:
Fixed by #452

PR #452, opened 2018-04-26
If a new device is connected ask the user if they want to change mode

microbit-carlos:
Based from the discussion on https://github.com/mu-editor/mu/issues/445:

- [x] If a new micro:bit is detected and Mu's not in micro:bit mode, it should ask.
- [x] If a new Adafruit board is detected and Mu's not in Adafruit mode, it should ask.
- [x] If both an Adafruit and micro:bit are plugged in, Mu does nothing and assumes the user knows what they're doing.
- [x] Tests!

@ntoll I don't have a CP Express device to test, so if you could do that it would be great.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1010,17 +1010,16 @@ class Editor:
         logger.info('Showing available modes: {}'.format(
             list(self.modes.keys())))
         new_mode = self._view.select_mode(self.modes, self.mode, self.theme)
-        if new_mode and new_mode is not self.mode:
-            self.mode = new_mode
-            self.change_mode(self.mode)
-            self.show_status_message(_('Changed to {} mode.').format(
-                                     self.mode.capitalize()))
+        if new_mode and new_mode != self.mode:
+            logger.info('New mode selected: {}'.format(new_mode))
+            self.change_mode(new_mode)
 
     def change_mode(self, mode):
         """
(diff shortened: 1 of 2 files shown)


microbit-carlos:
Looks like assert_called is not available in Py3.5? I'll look into the tests errors later today.

ntoll:
This looks good, but before merging I hope you don't mind me making sure that if you unplug a device it gets removed from `connected_devices`. I'll update the tests in light of my modification and then merge.

microbit-carlos:
At the moment it doesn't remove devices if they are no longer connected. In the previous version a new `set()` was assigned to `connected_devices` every time a new device was found. So the set was kept up-to-date by always containing a fresh list, but I think in reality it might have only really contained the last found device, as any addition to the list removed any previously saved entries.

The reason I have kept the found devices permanently is because the `find_device` method returns a port for a found device, so if we were to remove devices from the `connected_devices` set then we would end up removing them every time Mu failed to find the port for the device, which might happen more often than we anticipate.

It's not uncommon for the REPL to fail to connect when the micro:bit has been recently reset or flashed, but now that I think about it maybe that's an issue with connection instead of discovery. I was also unsure how often the CP devices auto-reset, is it possible to lose the serial port at any point with those devices?

PR #455, opened 2018-04-28
WIP - Scan code for imports and function names to detect code

carlosperate:
Been thinking about this one for a while, and the comment @ntoll made yesterday about having things like a "bottle" mode got me interested to visit this one today.

The PR is incomplete, the code present here only offers the core functionality to parse a given python code AST and look for imports (like `from microbit import *`) and function names (like pygame zero `def draw():`), but this will allow Mu modes to indicate if a Python script is likely to be for their device/framework.

As a quick example I haven't yet commited (code area is refactored in PR #453) I basically added a list of "unique modules" and "unique functions" to each mode that helps identify a compatible mode from the  Python code:

ezgif-4-609b00d3aa

### DIFF
--- a/mu/modes/base.py
+++ b/mu/modes/base.py
@@ -18,10 +18,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import json
 import os
-import os.path
 import csv
 import time
 import logging
+import ast
 from PyQt5.QtSerialPort import QSerialPortInfo
 from PyQt5.QtCore import QObject
 from mu.logic import HOME_DIRECTORY, WORKSPACE_NAME, get_settings_path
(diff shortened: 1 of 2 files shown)


carlosperate:
@ntoll any feedback?
One thing I would like to be able to do in the future with this as well, is to be able to auto depect imports and scan the mu_code directory for files. So if my main.py imports `my_module` and there is a `my_module.py`, it could detect it it and automatically (or by asking) it could copy it over to the micro:bit as well.

ZanderBrown:
Nice use of the ast module!

I can't help feeling that as we add more things like this we get closer to needing mode to be set at the file not application level?

Slightly concerned with all these "nice things" we are in danger of some feature creep and never actually releasing v1

carlosperate:
I'm happy to postpone this to after a v1, just wondering if it's something I should invest the time continue developing or if it falls outside of the scope for Mu.

ntoll:
FWIW I think this is a post 1.0 feature. I'm anxious to get Mu 1.0 out the door asap.

PR #463, opened 2018-05-03
Let microfs manage its serial connection and other serial updates

carlosperate:
Since microfs is a lot more aware of timings required for the serial port opening and closing, let that module manage its own serial instances.

Replace the test for microbit presence by trying to open a serial connection, by just scanning available usb ports instead.

Set the Window class to close Qserial connections when panes are removed instead of relying  on the destructor to do it.

### DIFF
--- a/mu/modes/microbit.py
+++ b/mu/modes/microbit.py
@@ -114,8 +114,7 @@ class FileManager(QObject):
         or emit a failure signal.
         """
         try:
-            with microfs.get_serial() as serial:
-                result = tuple(microfs.ls(serial))
+            result = tuple(microfs.ls())
             self.on_list_files.emit(result)
         except Exception as ex:
             logger.exception(ex)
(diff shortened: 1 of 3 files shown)


mu/modes/microbit.py
```diff
@@ -429,10 +425,7 @@ def add_fs(self):
         Add the file system navigator to the UI.
         """
         # Check for micro:bit
-        try:
-            microfs.get_serial()
-        except IOError:
-            # abort
+        if not microfs.find_microbit():
```
carlosperate: For our purpose `find_microbit` and `get_serial` would both offer the same functionality.

While get_serial does open a serial port, it doesn't mean MicroPython is running in the nRF51, as the USB serial functionality is done by DAPLink. Maybe we'll have to change the error message below as well?
We could then move the `The device must have MicroPython flashed onto it beforethe file system will work.` to whichever signal recipient deals with filesystem errors?

carlosperate:
Ups, shoudn't forget to add the tests. CI should be passing with the latest amend.

ntoll:
LGTM +1 thanks @carlosperate

Issue #446, opened 2018-04-19
settings.json on osx?


brjdk:
hi mu developers

new here, new to osx, new to mu and new to micro:bit - so quite new :-)

i have some questions, but i'm not sure where to ask them - please feel free to move them or guide me.
- where do i find the settings.json on osx? i have searched but not found.
- is it possible to disable the warning popup that pops up every time i press flash?
- is it there any keyboard shortcuts for eg. flash?

thanks
Jorgensen

ZanderBrown:
Hello, new people is what we are all about!

Do you know what version of Mu you are using? We've been working on a new version for a couple years now so some things have been implemented just aren't in the codewith.mu download yet..

The folder mu uses for settings.json is inside a hidden directory so your unlikely to find it  by searching

If your version of Mu has the cog icon in the bottom right you can click it to see the logs, near the top you should see the settings path listed

brjdk:
Hi Zander

I downloaded my version (can't figure out how to check the version) from codewith.mu, so i guess its an old version?

but no cog icon.

where do i find the lastest version for osx? 
i can't quite figure this page out 
[https://github.com/mu-editor/mu]

jorgensen

brjdk:
I've just tried download the latest version (v0.9.13) from here [https://github.com/mu-editor/mu/releases](https://github.com/mu-editor/mu/releases), but i don't see a cog in the buttom right corner.

Jorgensen

brjdk:
Thanks Zander - do you know how i download the latest version for osx?

Jorgensen

ZanderBrown:
macOS test packages are available here: http://mu-builds.s3-website.eu-west-2.amazonaws.com/?prefix=osx/ but you may have to try a few to find one that works

Alternatively if you have pip installed you could use `pip3 install mu-editor` in a terminal, there are [tutorials online](http://itsevans.com/install-pip-osx/) for how to install pip

brjdk:
hi zander

i went ahead and tried what you suggested - installed python3 and pip (not really knowing what i was doing). 

then i added install mu-editor, and a lot of packages got downloaded and installed, and then it ended like this 

Successfully built pgzero tornado simplegeneric backcall
[...]

PR #468, opened 2018-05-14
Check for exceptions when trying to check the locale

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/446
Also needed to launch Mu from finder when packaged in an app bundle.


ntoll:
Thanks @carlosperate.

Issue #401, opened 2018-03-22
print Chinese char in jupyter repl work well and could work when run a py with run button


yonghuming:
![image](https://user-images.githubusercontent.com/435079/37758673-fe0b2424-2deb-11e8-87a4-c21bfae38689.png)

ntoll:
Are you using the latest version? I fixed this yesterday! :-)

ntoll:
See latest builds here: http://mu-builds.s3-website.eu-west-2.amazonaws.com/?prefix=windows/

tjguk:
Is it possible that, because we don't write the encoding cookie until we save (and then reload) that Jupyter's complaining? If so, we could (a) add to the new file; or (b) reload when saving so we get it in place

yonghuming:
@ntoll i use windows 10 64bit and when i print chinese it will exit immediately

ntoll:
Please confirm that you are using the *latest* version of Mu: https://s3-eu-west-2.amazonaws.com/mu-builds/windows/mu_2018-03-21_18_47_master_5ef9518_64bit.exe

yonghuming:
yes i am sure,

yonghuming:
# windows 10

ntoll:
OK... so I've managed to re-create a problem. The latest build of Mu fixes your original problem @yonghuming, however, I now see a new problem when running this on Windows (NOTE: it works fine on Linux and OSX -- I've just checked). See the attached screenie... @tjguk it looks like a Windows specific problem relating to `charmap`.  Any ideas..?

![screenshot 2](https://user-images.githubusercontent.com/37602/37763598-7655266c-2db7-11e8-8493-8fb7aecd6d13.png)

NB: I'm using the latest build (referenced in previous comment) on Windows 10 64bit, just like @yonghuming. The code I was running:

```
print("你好")
```

ntoll:
To recreate, copy the `print("你好")` code into a file and run it. Works fine on Linux/OSX.

yonghuming:
bot newest 32 and 64bit mu could not run

yonghuming:
font{
        line-height: 1.6;
    }
    ul,ol{
        padding-left: 20px;
        list-style-position: inside;
    }


    
i have tried but notwork



    
        a#ntes-pcmail-signature-default:hover {
            text-decoration: underline;
            color: #199cff;
            cursor: pointer;
        }
        a#ntes-pcmail-signature-default:active {
            text-decoration: underline;
[...]

PR #471, opened 2018-05-20
Ensure embedded process is run with UTF-8 encoding on stdout/err

tjguk:
This PR -- currently without tests -- addresses issue #401. The stdout encoding defaults to something which couldn't handle the full range of codepoints. Attempting to print, eg, Chinese characters failed noisily with a UnicodeEncodeError.

By setting the PYTHONIOENCODING env var for the process to UTF-8 the bytes are able to be encoded to be picked up by the output handler.

Submitting the PR without tests for now so that it's out there and not going mouldy on my hard disk. Hopefully I'll come back later to test.


ntoll:
Hi @tjguk I fixed/amended the tests. Thanks for the fix!

PR #481, opened 2018-06-09
Fix issue where the margin background doesn't update when theme changes

stestagg:
As per jacobslusser/ScintillaNET#279, scintilla is quite sensitive to when the margin colors are set. 

By moving the call to setMarginsBackgroundColor() after
setLexer(), changing the theme now updates the margin background and foreground colours

An example of current behaviour on mu latest:
screenshot 2018-06-09 at 16 51 09

After this change (on master):
screenshot 2018-06-09 at 16 52 18

(Note, the main editor text color issue is separate)


ntoll:
Great spot @stestagg and thank you for the contribution. :-)

ZanderBrown:
Finally! Been trying to work this out for months

PR #479, opened 2018-06-05
Add french translation

gquintana:

PR #484, opened 2018-06-14
Translateable untitled

ZanderBrown:
Based on a comment by @gquintana in #479 

Okay/Cancel labels are coming from Qt itself, not sure what to do with them


ntoll:
@ZanderBrown I believe Qt's own translation for "Ok/Cancel" should kick in. Thanks for the spot and fix.

Issue #464, opened 2018-05-03
Failing tests still pass AppVeyor builds


carlosperate:
Forgot to commit the tests for my last PR, and the AppVeyor build still passed, so it looks like failures in the tests are not being detected: https://ci.appveyor.com/project/carlosperate/mu/build/1.0.947

I assume AppVeyor expect a non-zero exit code on failure, looking at make.py the pytest return codes are meant to be passed through. I've run a failing test on mac and the terminal does indeed get the right `1` exit code.

@tjguk, if you are not too busy, could you run a failing test on Windows and see what make.py returns? If you have any other thoughts about what might be happening that'd be great as well.

PR #495, opened 2018-06-19
makep.py: Check for return codes

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/464.

check() in make.py wasn't stopping on errors, causing failing builds in AppVeyor to be set as "passed all tests". This PR updates the script to check for the return value of each step in check() and exits early with the correct error code if there is a problem.

### DIFF
--- a/make.py
+++ b/make.py
@@ -166,10 +166,17 @@ def check():
     """Run all the checkers and tests
     """
     print("\nCheck")
-    clean()
-    pyflakes()
-    pycodestyle()
-    coverage()
+    funcs = [
+        clean,
+        pyflakes,
+        pycodestyle,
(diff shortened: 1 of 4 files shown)


tjguk:
LGTM. Thanks for picking this up

ntoll:
@carlosperate the tests fail on appveyor... Care to check what's going on..?

carlosperate:
The tests have been failing for a while, but AppVeyor didn't indicate so because of the issue resolved in the PR.

ntoll:
@carlosperate OK... I'll fix the tests in the morning.

ntoll:
Fixed. Thanks..!

PR #499, opened 2018-06-21
Context menu for local files in micro:bit mode

### LINKED ISSUE #174: Add 'open file in editor' option to file on host drive FILES view
It would be helpful to be able to click on the file in the right hand pane in the FILES view and open it to edit the host version of that file in the default editor for that file. 

e.g. if you have a text file with some data in it, instead of having to find out where Mu stores those files (userdir/python) and navigating there to edit it, you could just double click on it and open it in a text editor. I'd be happy if this action was limited to just .txt and .py files for example, and if it allowed that file to be opened in a new tabbed pane in the Mu editor.

ZanderBrown:
Finally resolve #174, when I said I'd look into it later I had intended a few days, not 16 months! (sorry @whaleygeek)

All local files gain an "Open" command which basically gives the file to Qt to *do the right thing™*
screenshot from 2018-06-21 23-16-31

Files with a `.py` or `.hex` extension have an additional "Open in Mu" which *does what it says on the tin™*
screenshot from 2018-06-21 23-16-38

And I was right, it wasn't too difficult! (apart from a moment of silliness with the tests)

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -457,6 +457,7 @@ class FileSystemPane(QFrame):
         self.font = Font().load()
         microbit_fs = MicrobitFileList(home)
         local_fs = LocalFileList(home)
+
         @local_fs.open_file.connect
         def on_open_file(file):
             # Bubble the signal up
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -25,12 +25,14 @@ import signal
 import string
 import bisect
 import os.path
-from PyQt5.QtCore import Qt, QProcess, QProcessEnvironment, pyqtSignal, QTimer
+from PyQt5.QtCore import (Qt, QProcess, QProcessEnvironment, pyqtSignal,
+                          QTimer, QUrl)
 from collections import deque
 from PyQt5.QtWidgets import (QMessageBox, QTextEdit, QFrame, QListWidget,
                              QGridLayout, QLabel, QMenu, QApplication,
                              QTreeView)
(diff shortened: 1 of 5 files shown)


mu/interface/panes.py
```diff
@@ -407,6 +410,34 @@ def on_get(self, microbit_file):
         self.set_message.emit(msg)
         self.list_files.emit()
 
+    def contextMenuEvent(self, event):
+        menu = QMenu(self)
+        local_filename = self.currentItem().text()
+        # Get the file extension
+        ext = os.path.splitext(local_filename)[1].lower()
```
ntoll: Will this work with files that don't have an extension (a distinct possibility when there are kids involved)..?

ntoll:
@ZanderBrown this looks great. Two things:

* The build fails because `make check` results in an error (PyCodeStyle related).
* Can you add a guard around getting the extension of the file..? Given the current code Mu would crash with an index error if the file didn't have an extension.

Thanks! :-)

mu/interface/panes.py
ZanderBrown: Yes, splitext is actually returning a turple not list and [1] is guaranteed to be atleast an empty string

ZanderBrown:
I'll fix the styling shortly

ZanderBrown:
Green marks all around

ZanderBrown:
Ping @ntoll

ntoll:
Ack... busy day tomorrow but it'll be in beta.17 for certain. ;-) Thanks for the kick.

ntoll:
Great stuff @ZanderBrown..! Thank you (once more) for your contribution.

Issue #498, opened 2018-06-21
Initial run mode oddness


ZanderBrown:
We now detect connected micro:bits and offer to switch to the relevant mode which is great!

However when we are on the initial run this leads to an interesting scenario where the mode selector pops up followed by the "would you like to switch". Interacting with the switch message goes roughly as expected but the mode switcher remains open.

All things considered this seems to be handled quite well but is a bit odd for users

ZanderBrown:
Running against master:
![screenshot from 2018-07-06 09-32-28](https://user-images.githubusercontent.com/17074021/42369366-a2047fb8-8101-11e8-8279-2cf64ca116a9.png)

Running against #525:
![screenshot from 2018-07-06 09-31-01](https://user-images.githubusercontent.com/17074021/42369365-a1beef8e-8101-11e8-8e24-750ca929df64.png)

Only change introduced is that the mode selector can't disappear behind the main window, in both branches only the mode switched accepts input. (The darkened area is a result of the WM knowing the modal relationship)

Also observed that the switch message can appear on top of the splash screen

carlosperate:
> the switch message can appear on top of the splash screen

I was just wondering if I opened a ticket for this one already, I don't think I did.
In my case, with the current Mu, the switch message appears behind the splash screen, but doesn't allow me to click on the spalsh screen to dismiss it, so I have to wait until the time-out removes it.

ntoll:
Heh... I keep seeing this too and it's on my "to fix in beta.17" list. I'll be all over this tomorrow.

tmontes:
I wanted to test this to verify what the current state of affairs would be, but I just can't seem to find my micro:bit! (why is it so micro? if it was bigger, I'm certain it would be easier to find) :)

Given how old the issue is and the note by @ntoll regarding an upcoming fix, can we say this issue is indeed fixed?

Context: I'm trying to contribute to some "issue triaging and organisation" help understand which of the bazillion open issues actually need attention. :)

carlosperate:
> I just can't seem to find my micro:bit!

If you still can't find it you can send me your details and I'll post you one :) (my twitter DMs are open, and/or my email is my github username at gmail.com).

tmontes:
Thanks a lot for your generous offer @carlosperate. :) :) :)

I'm positive that if I put a bit of effort into the task, I'll find it. It's just that when I went for it, I couldn't find it in one of the (not so few) obvious places.

dybber:
[...]

PR #525, opened 2018-07-05
Modal dialogs

ZanderBrown:
Mode & Admin were showing some odd behaviours (at least on mutter/GNOME but I suspect others are affected) largely because they were being modal to an unset parent

This is slightly WIP as FileDialogs still aren't working properly but I'm 99% sure that's a problem in `QGtk3FileDialogHelper` (aka upstream)

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -48,6 +48,9 @@ class ModeSelector(QDialog):
     Defines a UI for selecting the mode for Mu.
     """
 
+    def __init__(self, parent):
+        super().__init__(parent)
+
     def setup(self, modes, current_mode, theme):
         if theme == 'day':
             self.setStyleSheet(DAY_STYLE)
@@ -168,6 +171,9 @@ class AdminDialog(QDialog):
     Displays administrative related information and settings (logs, environment
     variables etc...).
     """
+
+    def __init__(self, parent):
+        super().__init__(parent)
 
     def setup(self, log, settings, theme):
         if theme == 'day':
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -48,7 +48,7 @@ class ModeSelector(QDialog):
     Defines a UI for selecting the mode for Mu.
     """
 
-    def __init__(self, parent):
+    def __init__(self, parent=None):
         super().__init__(parent)
 
     def setup(self, modes, current_mode, theme):
@@ -172,7 +172,7 @@ class AdminDialog(QDialog):
     variables etc...).
     """
 
-    def __init__(self, parent):
+    def __init__(self, parent=None):
         super().__init__(parent)
 
     def setup(self, log, settings, theme):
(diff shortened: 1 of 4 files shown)


carlosperate:
Would this change affect the behaviour described in https://github.com/mu-editor/mu/issues/498?
I'm wondering if having the two windows (mode selection and swicth to microbit/adafruit mode) could cause some weird scenario where one blocks the other.

mu/interface/dialogs.py
```diff
@@ -48,6 +48,9 @@ class ModeSelector(QDialog):
     Defines a UI for selecting the mode for Mu.
     """
 
+    def __init__(self, parent):
+        super().__init__(parent)
+
```
ntoll: Surely these redefinitions of `__init__` are surplus to requirements..? If they didn't exist the `__init__` on the base class would be called anyway (and that's simply what you're manually doing here). Right..? Or am I missing something here?

tests/interface/test_dialogs.py
```diff
@@ -57,7 +57,7 @@ def test_ModeSelector_setup():
     current_mode = 'python'
     mock_item = mock.MagicMock()
     with mock.patch('mu.interface.dialogs.ModeItem', mock_item):
-        ms = mu.interface.dialogs.ModeSelector()
+        ms = mu.interface.dialogs.ModeSelector(None)
```
ntoll: Why is `None` needed here..?

Also, CI complains because the tests don't pass. :-(

ntoll:
Hmm... if we can improve the behaviour of modal dialogs then this is a good thing (tm). But I'm confused (see comments) about some of the changes (I'm probably missing something here). Also, the tests aren't passing.

When the above is addressed, happy to merge this. :-)

mu/interface/dialogs.py
ZanderBrown: Yes the base `__init__` would be called but with `parent=None`. Qt tries to be clever and makes the dialog 'application modal' meaning it puts the dialog to the top whenever another window is raised above it. End result is magic disappearing then reappearing dialogs.

By setting the parent here Qt can set the correct Window Manager hints so the dialog is modal to the parent properly

I'm not sure I am explaining this well

tests/interface/test_dialogs.py
ZanderBrown: With hindsight `__init__` should have `parent=None`

ntoll:
@ZanderBrown the tests still don't pass... see the output from Travis, specifically:

```
>           mock_admin_display.assert_called_once_with()
tests/interface/test_main.py:1132: 
```

ZanderBrown:
:man_facepalming: 

Sorry haven't been running tests locally because they require QtChart which i don't currently have after debugging the RPM...

ntoll:
Great stuff..! Tests passing (but OSX build is knackered, I'm fixing it tomorrow).

PR #556, opened 2018-07-15
Portugal's Portuguese UI translation.

tmontes:
Nicholas (and contributors),

Here's a huge "thank you very much" and "hats off" and "fireworks" and "whatever-you-like" to you for the amazing Mu. I became aware of it about two years ago when you presented it at EuroPython in Bilbao, along with the BBC micro:bit with which I play a bit, every now and then.

You might remember me -- don't worry if you don't! -- from back then: I was your family's Python coach at the Beginner's Day, that year... :) I also promised I would contribute to the project someday. Well, today is the day.

This PR is exactly what it says on the tin. I faced a few challenges like, I assume, any translator does. When in doubt, I tried to keep a beginner audience in mind as much as I could, avoiding "technical jargon" as much as possible. I hopefully made good choices: time will tell. (and, by the way, I never faced the need to translate, for example, "Step Over/Into/Out", "Plotter", or "REPL"... Having worked with computers since forever, to me it's more than natural that computers speak English!).

Technical PS: Not being familiar with gettext, I'm including both the .po and .mo files. If any of those should be excluded, please advise.

Non-technical PS: While I appreciate the Zen of Python a lot, I would say that it is not very beginner friendly... Picture someone who's just started playing with their first "print" statements and simple number calculations. Now tell them that "Flat is better than nested"... What's their face?!... Hopefully, in time, it will make sense for them, of course.


ntoll:
Hi @tmontes Of course I remember you... and Mary and William both say "hi". Having you as a tutor at EuroPython was one of the highlights of the conference for them... they got to get a flavour of what I do.

Thank you for your kind words too... it means a lot. Here's hoping young/beginner Portuguese speakers find it helpful.

By the way, how different is Brazilian Portuguese to Portuguese Portuguese..? Could this translation work for both?

Many many many thanks for your contribution. :-)

tmontes:
Nicholas, thanks for merging and please say hi back to William and Mary, hoping you are all fine. :)

Regarding Brazilian Portuguese: I would say that it tends to be quite different, especially in technical contexts, as they tend to have a more creative approach to the language, whereas we tend to be more conservative, if I may call it that. A Brazilian speaking person will be much more at home with a `pt_BR` translation, even though `pt_PT` would probably be usable for a non-English speaking person. As an example, they tend to use "Salvar Arquivo" (rescue archive) for "Save File", while we often go with "Guardar Ficheiro" (keep/store file).

PS: By the way, from my a bit more limited experience, the same could probably be said about Spanish and its European and Central/South American variations.

More PS: Your very pertinent question raises a related one: clicking help is currently directing the web browser to a "language" only URL of the form `https://codewith.mu/<LL>/help/1.0`, where `<LL>` are the first two letters in the locale. Maybe at some point in the future, it may make sense to use the full string there, having the server side try its best to accommodate the request (trying, say, `pt_BR`, then going for `pt` only, falling back to the current `en` only, which I guess is actually `en_GB`).

ntoll:
The language specific help is on my to-do list. ;-)

Issue #552, opened 2018-07-12
Dark Themes no more readable on macOS Mojave beta


sstaub:
b17, MacBook Pro15'' 10.14 Beta (18A326h) Mojave
The dark themes are no more readable on macOS Mojave Beta.
This happens on the both modes of Mojave.
<img width="288" alt="screenshot" src="https://user-images.githubusercontent.com/6450325/42654717-e0fb70d8-8619-11e8-86ef-f538617c6fb7.png">
<img width="302" alt="screenshot-1" src="https://user-images.githubusercontent.com/6450325/42654720-e37e7620-8619-11e8-812f-bca887c7a01c.png">

ZanderBrown:
Ha! Saw this earlier when someone tried it out, didn't really get a chance to investigate.

Especially when Mojave is in beta it's hard to say if this is a bug with macOS, Qt or ourselves.

@carlosperate how do they look on your mac?

sstaub:
It seems it is an macOS Mojave problem, the new dark mode is more alpha than beta.

ZanderBrown:
> My macbook crashes enough as it is (probably a hardware issue though), so I wasn't too eager to install a beta OS. HS has been fine.

(Sorry I did mean non-Beta)

sstaub:
The mu normal mode is ok, only the dark mu themes are the problem, independently of the mode of mojave.

ZanderBrown:
So just to be clear: This happens with the Mojave light theme?

sstaub:
Yes, with light and dark theme.

stestagg:
I'm seeing this too, in Mojave with both OSX light and dark theme.

For some reason, it's caused by loading/using the font from a QFontDatabase.

I've got Source Code pro installed in my system, and running mu 1.0rc unmodified shows:
<img width="932" alt="screenshot 2018-07-17 at 23 57 54" src="https://user-images.githubusercontent.com/704526/42850116-27737e90-8a1e-11e8-968a-b66047b173df.png">
<img width="932" alt="screenshot 2018-07-17 at 23 57 56" src="https://user-images.githubusercontent.com/704526/42850119-293491a6-8a1e-11e8-8f28-959db7183a26.png">

Note, all text is black, just the background color changes.

But commenting out line 67 of themes.py:
```cls._DATABASE.addApplicationFontFromData(font_data)```
causes mu to use the system installed (identical) version of Source code pro, and the text colours immediately return:

[...]

PR #567, opened 2018-07-18
Temporary workaround for #552

stestagg:
Not my best work, but puts in a temporary patch to change the editor font from Source code pro to Monaco when run on OSX Mojave

I'm matching on any "Darwin-18.*" version here (Mojave), which may be over-generous, but it's better (in my opinion) to use a slightly different font for too long on Mojave than have people randomly getting black on black text after an upgrade.

Before:
screenshot 2018-07-18 at 22 42 22

After:
screenshot 2018-07-18 at 22 39 33

Tested on OSX Sierra (note - uses Source code pro still):
screenshot 2018-07-18 at 22 32 02

I haven't been able to test on Windows or Linux yet, but I'm reasonably confident this hasn't broken anything there

### DIFF
--- a/mu/interface/themes.py
+++ b/mu/interface/themes.py
@@ -37,7 +37,7 @@ def should_patch_osx_mojave_font():
     update FONT_NAME to use the default terminal font in OSX on mojave.
 
     This patch should be removed once the underlying issue has been resolved
-    
+
     github issue #552
     """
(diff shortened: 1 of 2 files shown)


stestagg:
@ntoll ^^

ntoll:
Thanks @stestagg ..! Tested on Linux and the tests all pass on Windows too.

PR #577, opened 2018-07-25
Add German translation for Mu

reneraab:
I translated all of the strings (from the current master branch) to German.


tjguk:
Thanks very much. We've just been told of an article in the German Make: magazine -- https://www.heise.de/make/meldung/Mu-Neuer-Python-Editor-fuer-Einsteiger-4119566.html so this is especially timely. I'm not sure when our next release is going to be, but hopefully the German translation will be in it whenever it is.

ntoll:
Thank you for your contribution! :-)

It'll be in the next release of Mu.

PR #583, opened 2018-07-29
Add new Adafruit board

ladyada:
spooky board with spooky PID :)


ntoll:
Thanks! :-)

ladyada:
@ntoll heya - would it be possible to get a v1.0.1 release bump so the new board is included :)

ntoll:
`<onholiday>`Hey hey... yes... I intend a point release to fix a few glitches in 1.0 before the start of September (i.e. new academic year). :-) `</onholiday>`

ladyada:
@ntoll hiya can we get a minor bump for this board? we just shipped 4000 of these board to adabox people and i could build my own release but would prefer not to!

PR #586, opened 2018-07-29
Add to the log the detected language code

carlosperate:
Tiny addition to add to the log the detected language code. This should be useful in cases where translation don't seem to be automatically loaded in operating systems configured in a supported language.


ntoll:
Thanks @carlosperate :-)

PR #588, opened 2018-07-29
File dialog boxes remember previous folder used

templed:
Hello,

I recently started using Mu for little projects and am enjoying it immensely. However, a tiny annoyance I have with it is the file dialog boxes always open on `~/mu_code`. This commit implements a shared memory for the save and open file dialog boxes across the editor whilst defaulting the `~/mu_code` if nothing has been opened/saved before during the lifetime of the editor.

Happy to hear any feedback you might have

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -190,6 +190,7 @@ class Window(QMainWindow):
     data_received = pyqtSignal(bytes)
     open_file = pyqtSignal(str)
     load_theme = pyqtSignal(str)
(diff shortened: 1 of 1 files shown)


ntoll:
Hi @templed, I've taken the liberty of adding a check that the `previous_folder` attribute is correctly updated. Otherwise, good call and thank you for your contribution! :-) (Sorry it's taken a while to get round to triaging these outstanding PRs).

Issue #487, opened 2018-06-18
80 character line


bennuttall:
I wondered if it was worth adding (either for everyone, or as an option), a visible 80 character line length ruler? No wrapping, just an indicator so you know where the 80 character limit is.

Personally I think I'd find it really useful, and it would help guide beginners.

The Check button will tell you if you go over 79 characters, but not where the 80th character is in your line, so it's hard to know where exactly you need to break the line.

martinohanlon:
Playing devils advocate on this one...

Is this going to be confusing for new learners? Will they think there is a hard limit? Does it add complexity where its not required? 

If you are writing programs where you need to worry about lines over 80 chars (I only tend to think about this in the later stages of a project before release), you have probably moved past Mu.

tjguk:
FWIW I'm with @martinohanlon on this. I think an advantage of Mu is its relative clean-ness: just a blank space and a row of buttons.

I don't remember whether one of the check tools picks this up, but I suspect that at least one does.

ZanderBrown:
PEP8 would / does

carlosperate:
We have a list of PEP8 rules to ignore: https://github.com/mu-editor/mu/blob/master/mu/logic.py#L413-L414
Perhaps it would be worth adding this one as another exception.

Although a bigger question could be if these exceptions are worth adding to the settings window as a new tab? Not all the pycodestyle rules, but popular/controversial ones as tick boxes with some same defaults for Python learners.
My first instinct would be "not necessary right now until more people request it", but I wonder what everybody else thinks about this one.

NDevox:
I was going to suggest some possible configuration around line length but in the end convinced myself that I'd rather not have to configure all my students machines.

This extends to tick boxing on rules.

In our dojo we do do some initial setup with editors etc but I'm 90% sure that in the end for things like this we'd just use whatever the default setting would be for this kind of thing.

I'd rather not keep this rule in by default. It's just off putting when a child is writing sentences of text in the editor only to be told they have to break it up.

But then I use 120 as my limit. So I'm not really one to argue on standards.

dybber:
I agree with what has previously has been said, that this will add more complexity for new learners, who are working with something already cognitive demanding. I think it's a good idea to make it possible for users to customize which pycodestyle rules are checked, e.g. by right-clicking on an error with a menu item to ignore the warning in the future.

[...]

PR #600, opened 2018-07-31
Adding a vertical line in the editor to make the column limit visible

osaatcioglu:
I agree with @bennuttall (#487). Visible vertical line will make it easier to understand the general guidelines for the beginners. Luckily, QsciScintilla provides this feature out of the box. I just added the configuration.


osaatcioglu:
No problem at all. I like the idea to configure it. Is it something I can help? I could change the PR to add the configuration as well. Thank you for making this software.

Issue #603, opened 2018-07-31
Handling ANSI color escape sequences for MicroPython REPL


dybber:
The REPL currently only handles a limited set ANSI escape sequences, and errs if it hits an unknown sequence. In particular color sequences as described here, are not handled:
https://stackoverflow.com/questions/4842424/list-of-ansi-color-escape-sequences?answertab=votes#tab-top

The error occurs in `process_bytes` at https://github.com/mu-editor/mu/blob/master/mu/interface/panes.py#L237
where the regex doesn't match every possible action and doesn't allow actions taking multiple arguments.

A work around is given here, where the regex is extended to match multiple arguments (semicolon seperated list), and added the "m" action to the list of recognized actions. In addition we should guard the use of `m.end()` and `m.group()`, in case some other action is received that is not on our list.
```
                m = re.search(r'(?P<count>[\d]*)(;?[\d]*)*(?P<action>[ABCDKm])',
                              data[i:].decode('utf-8'))
                if m:
                    # move to (almost) after control seq (will ++ at end of loop)
                    i += m.end() - 1

                    if m.group("count") == '':
                    ....
```

PR #608, opened 2018-08-04
Make MicroPython REPL process_bytes disregard color escape sequences

dybber:
Currently Mu crashes when receiving ANSI color escape sequences. With this patch, such sequences are ignored.

See issue  #603


ZanderBrown:
Might be nice to support them properly but this is at least a short-term fix

dybber:
Yes, the intent was to make a short-term fix to avoid crashes.

PR #618, opened 2018-08-18
Move the cursor to the end of the initial content

### LINKED ISSUE #616: # Write your code here :-) - put the cursor on line 2
I have observed a number of new users of Mu, and there is a minor change I think would be really helpful.

The first time Mu starts your are presented with:

```
# Write your code here :-)
```

Lovely...

What I have observed is that the cursor position is at the start of line 1 before the `#`.

This created a little bit of friction as when new users started to type their first program the text appeared before the comment.

```
print("hello") # Write your code here :-)
```

While it would work, many people were confused as the code didnt look like what they were expecting.

Could the cursor be placed at the start of on Line 2?

ZanderBrown:
Should resolve #616

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -282,6 +282,7 @@ class Window(QMainWindow):
         new_tab.setFocus()
         if self.read_only_tabs:
             new_tab.setReadOnly(self.read_only_tabs)
+        return new_tab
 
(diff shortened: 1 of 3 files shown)


ntoll:
@ZanderBrown stonking stuff. Thank you..! :-)

Issue #594, opened 2018-07-30
Cannot close all files in mu-editor on iMac


marcovanurk:
If I open mu-editor on my iMac, it always opens with the last script I edited. And I can't close that script. It sticks there 'forever'. If I open more scripts using 'New'. I can't even close that other script.

If I close mu-editor with more than one script. And open mu-editor again, all the scripts are there. I can close them all, but one.

The only workaround is, after closing mu-editor, to delete the file 'session.json' in 'Library/Application support/mu'.

marcovanurk:
No it is not the far left tab. If I open a new one. And make that the most left one, I can delete this one.

Rearranging does not effect this. If I rearrange, I could not delete the one who came during opening mu-editor. It is since the last version; I use mu 1.0.0. The one before this one did not had this problem; as far as I can remember.

marcovanurk:
By the way, to be complete: it is when using mu-editor in 'BBC micro:bit' mode.

ZanderBrown:
So it's the tab that started on the left? Very weird.

This defo feels like a Qt bug, @carlosperate can you replicate this?

jamesabela:
Yes, I have the same problem on my Mac.  The script to the far left refuses to go away even when I have other tabs.
![image](https://user-images.githubusercontent.com/16495873/43675746-8c9f80aa-9817-11e8-8eec-7b6db34beeed.png)

carlosperate:
I can confirm the issue on macOS. Even if the tab is moved around the other tabs the left close button doesn't seem to work.

ZanderBrown:
@carlosperate  it shouldn't be on the left anyway

https://github.com/mu-editor/mu/blob/bd722c72f63bd0d6f51847e82852bf0638c22002/mu/resources/css/day.css#L62-L66

Quick fix would be remove L65 and have macOS show close button on the left, not necessarily ideal but should work. Can't find a QTBUG for this.

carlosperate:
@ZanderBrown is it safe assumption that this is a QT bug, instead of a PyQt5 one? The assumption being that the CSS is process directly by QT, bypassing anything that PyQt5 wrapper might do?

If so, perhaps the most time-consuming task to open an issue would be to create a minimum example. Do you know if the QT tracker takes bugs reported with the Python bindings?
If we have to write a few lines of Python, maybe it would be more productive to do that in PySide2 since it is now the official Python bindings from QT?

[...]

PR #619, opened 2018-08-18
Quick workaround

ZanderBrown:
Avoids #594 until whilst we await any upstream patches


carlosperate:
Travis was being silly and giving out an odd test failure in one of the platforms. Nothing to do with the PR, so I'll restart it and I will likely fix it.

I've tested this PR in macOS, and this temporary workaround does fix the issue with the "stuck" first tab. The only downside is the obvious change in the close button position.

ntoll:
@ZanderBrown thank you for the fix. @carlosperate thank you for the test! :-)

Issue #633, opened 2018-08-30
Make mu (even more) friendly for dyslexic users


tim-mccurrach:
Mu actually does a really great job of this already, there are loads of aspects of the design that make it good for people with dyslexia. I had a few thoughts about what else could be done though.

- Lots of people who suffer ‘visual stress’ will use coloured overlays to help with reading performance, and the ability to customise the background/font colour can be helpful when reading off a screen. However the specific colour of the overlay can be important, and different people will need different colours. This will in turn mean you would also need to be able to customise the colours used for syntax highlighting, since there would probably be inevitable clashes. There are obviously, quite a few considerations and questions to ask in terms of how this could be best done, whilst maintaining the clean simplicity of mu’s design, but think it is definitely worth thinking about.

- The main principle above is that a high contrast is difficult can be difficult to read. In particular pure white, with pure black (as is used in the Day theme) can cause visual stress, and a slightly off white, with a very dark grey text is much easier to read. Such a change would make the default position much easier for dyslexic users, and make either no (or a slightly positive) difference to non-dyslexic users. Is it okay for me to just make a pull request here?

- I was also wondering about providing the option of a dyslexic-friendly font. There are quite a few out there. The current font is good as it is sans-serif which is generally helpful, but there are some fonts designed specifically for people with dyslexia that might help a lot. Perhaps this could be an option in the mu-admin dialogue?

- One interesting thing that I have found is (and this might just be me) I find the toolbar noticeably more difficult to navigate in the night and contrast modes, because the separators are virtually invisible. I’ve had some difficulty adjusting these settings, the `QToolbar::separator` selector allows some customisation, but I can’t find a property that adjusts the colour of the thin line in the middle. Is this possible?

ZanderBrown:
1) Like the idea but I'm not sure how doable it is, can see the benefits but adds a lot more settings/moving parts and the CSS is getting silly complex already. In an ideal world I'd rewrite it in less but I don't want to introduce extra (& non-python) build deps. Personally I'm red-green colour deficient which might explain some of the current colour choices :-)

2) Sure! Seems reasonable, like you say it won't be noticeable for most but may help some

3) If you can find a suitably licensed monospace font that sounds good, might cause more problems for Debian though...

4) That's semi-deliberate but QSS (Qt-CSS) does give us quite a bit of flexibility on this, what sort of thing where you looking for? I'm reluctant to change this too much now we have released 1.0

tim-mccurrach:
[...]

PR #644, opened 2018-09-10
Use off-white and dark-grey in day theme

tim-mccurrach:
I don’t think this fully addresses difficulties many users who find reading difficult due to ‘visual stress’ (pure white with pure black background can be very difficult for some to read) but it should make it slightly easier for most users. See issue #633

### DIFF
--- a/mu/interface/themes.py
+++ b/mu/interface/themes.py
@@ -73,7 +73,8 @@ class Font:
     """
     _DATABASE = None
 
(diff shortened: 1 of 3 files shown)


ntoll:
@tim-mccurrach thank you... also given your extensive research contributed to #633 (which I'll keep open since we still need to address many things related to that issue).

PR #646, opened 2018-09-10
Temporary fix for issue 622

tim-mccurrach:
To give full multi-line statement support as is in the Jupyter REPL, would need a significant re-write as currently, once we move on to the next line, the previous line becomes un-editable. This however is a temporary fix (which definitely improves the current behaviour), by moving the cursor to the end and then writing the whole line to std_out.

I felt a test should be added for this. I'm not sure if it has been done the best way, but have tried to follow approaches used in the surrounding tests.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -860,9 +860,15 @@ class PythonProcessPane(QTextEdit):
             self.backspace()
         if key == Qt.Key_Delete:
             self.delete()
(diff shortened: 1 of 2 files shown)


ntoll:
@tim-mccurrach this is a great sticking plaster for the problem. Thank you! Sorry it's taken so long to respond to this.

PR #667, opened 2018-10-09
Updating OSX to macOS in README

Craig88:
Updating the readme to refer to macOS (the new name for OS X)
(There are still references to OS X through the code, but that can be solved another time...)


ntoll:
Thank you! :-)

Issue #572, opened 2018-07-21
Locale settings are not reflected in macOS


MinoruInachi:
I checked Japanese translation at mu 1.0.0, but labels and messages remained in English.

The cause is that locale setting does not pass to python 3 when launching `mu-editor.app` from macOS 's Finder.  The values of the locale related environment variables at the time of starting python 3 were as follows:

```
LANG=
LC_COLLATE="C"
LC_CTYPE="C"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```
I made the following modifications to the startup script `/Applications/mu-editor.app/Contents/MacOS/mu_editor` and confirmed that the display will be in Japanese. I have confirmed this fix only in Japanese, and I have not confirmed whether it is okay with other languages.

```
--- mu-editor.org	2018-07-20 21:18:29.000000000 +0900
+++ mu-editor	2018-07-21 09:32:55.000000000 +0900
@@ -1,4 +1,5 @@
 #!/bin/sh
+"eval" "export LC_ALL=`defaults read -g AppleLocale | sed 's/@.*$//g'`.UTF-8"
 "exec" "`dirname $0`/../Resources/python/bin/python3" "$0" "$@"
 # EASY-INSTALL-ENTRY-SCRIPT: 'mu-editor==1.0.0','console_scripts','mu-editor'
 __requires__ = 'mu-editor==1.0.0'
```

carlosperate:
Yes, I can confirm that when launching the installed app from Finder the locale data is `(None, None)`:
<img width="932" alt="image" src="https://user-images.githubusercontent.com/4189262/43366913-a5a5bfe0-933c-11e8-8da3-120dcd2ae1bd.png">

tmontes:
Quick feedback:
* I confirm the behaviour noted by @MinoruInachi.
* I confirm the patch to the `mu-editor` shell script fixes the issue on my system.

A few extra observations, running macOS Sierra 10.12.6, where the "Language & Region" settings are:
* Preferred language: English
* Region: Portugal

With that, the `defaults read -g AppleLocale` prints out a non-obvious `en_PT`. This is a language string I had never seen before but that, given some thought, ends up making sense, I guess.

AFAICT, mu is currently using only the first two characters in the language code to determine its UI and help URL languages so this fact does not raise any limitation I can think of if a macOS fix for this based on `defaults read -g AppleLocale` is put in place.

[...]

PR #672, opened 2018-10-12
Improved language detection code with custom fallback for macOS.

tmontes:
Addresses #572 but it's not ready for merging yet. I'm mostly looking for feedback before moving foward. Thanks in advance for that!

Summary of changes:
* Created a new `localedetect.py` module exposing a single `language_code` function that uses `locale.getdefaultlocale()`, as before, but that now tries to run a platform specific language detection function when that fails; if that also fails, it falls back to a predetermined value which, for maximum *explicitness*, I also changed from `en` to `en_GB` (but happy to change back).
* Implemented a macOS specific language detection function:
  * It uses native APIs to get the preferred language from the "Language & Region" pane in "System Preferences".
  * It follows the sample code in #572 but uses `rubicon-objc` instead, which should simplify development, testing and packaging.
* Existing module design should make it trivial to add other platform specific detection functions if they're ever needed: it's a matter of adding platform specific functions named `_language_code_{platform}`.
* Changed existing `locale.getdefafultlocale()` calling code to use the now more clever function.
* Changed the logging message in "help" to include the URL being triggered.

Pending:
* Language detection / fallback tests are obviously failing and will need rethinking.

Food for thought:
* The language detection code is logging its behaviour but since the logging setup is completed later in the process, we end up not having those messages in the log file. I didn't look at the implications (if any!), but for maximum usefulness we might try to setup the logging earlier in the startup process.


ntoll:
This pull request **introduces 1 alert** when merging 8a7283d2280e1386e80c1cdccdb94c4adf070d45 into f89fc697ca67df9f71447b7f75aae521e4882fff - [view on LGTM.com](https://lgtm.com/projects/g/mu-editor/mu/rev/pr-932fd87e9aa9397b06f3b6afd779a366d724781c)

**new alerts:**

* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

tmontes:
Progress:
* Tests are in place.
* Have two unexpected results from CI:
  * No failures on Windows/Linux even though coverage on those platforms is <100% (I would have guessed that <100% coverage would lead to CI failures which, in turn, could lead us to bring in other tools into the CI to ensure 100% combined cross-platform code coverage).
  * Failure on macOS due to its platform dependency, `rubicon-objc`, not being installed.

Considerations:
* There seems to be somewhat of an overlap between the `install_requires` packages in `setup.py` and the package list in `requirements.txt`; the distinction is not 100% clear to me and maybe this could be improved.
* The [developer setup docs](https://mu.readthedocs.io/en/latest/setup.html) first suggest `pip install -r requirements.txt` and later `pip install -e .`.
* On my dev systems (macOS and Linux), `pip install -e .` is enough to run `mu`; the packages in the `requirements.txt` seem to be needed to run tests, code checking etc., but I didn't analyse that in detail.
* Maybe we could simply throw `requirements.txt` away and include the development requirements as an `extras_require` `dev` entry in `setup.py`: that way, `pip install -e .` would install the packages needed to run, and `pip install -e .[dev]` would install the packages needed to run *and* develop.

More:
* The tests failed on macOS because the Travis configuration is not running `pip install -e .` which, per this PR, installs the platform dependent packages: `rubicon-objc`.

What I'll do:
* I'll update the Travis config to `pip install -e .` before `pip install -r requirements.txt` which should make tests pass on all platforms. But doing both is an indication of possible improvements. I'll leave this up for discussion and/or addressing in the context of another issue and PR.

tmontes:
Status:
* PR is currently merge-able, AFAICT.
* Weather it should or not be merged, in its current form, depends on a few topics I'd love someone else to chime in on:
  * Names and structure: is `localedetect` a good name? should it be a module? is it in the right place?
  * Packaging: does this PR still produce running shippable binaries, given that a few changes were made to `setup.py`?
  * Is it a good/bad idea to have the "if anything fails default language code" be `en_GB` instead of `en`? (explicit is better than implicit, but...)

Related topics (for which we may want to create additional issues):
* Like I commented in the first message: maybe we could setup the logging process earlier such that locale detection messages are present in the log file.
* The `requirements.txt` vs `setup.py`'s `install_requires` + `extras_require` topic I commented about earlier; I feel there is room for improvement and simplification here, but I may not be aware of some details, in particular related to per-platform binary freezing/packaging.
* Lastly, I was under the assumption that PRs with <100% (absolute/incremental) test coverage would fail the CI: apparently they didn't. Is this something we may want to consider improving, going forward?

What I'll do next:
* Standby on some else's review and feedback.

Thanks.

tjguk:
Only commenting for long enough to say: thanks for taking this forward.

I'll need a little bit more clear water than I have today to look at what's proposed. I know that in the past @ntoll has been reluctant to add more modules. He can obviously speak for himself on this issue. (My own take is that breaking things out into more modules can make the code clearer although obviously one can go too far...).

tmontes:
@tjguk, thanks for your notice.

I'm in no hurry, so I'll be glad to get back to this once you, @ntoll, others take their time to review and comment.

ntoll:
@tmontes hey hey... sorry for the delay. I'll review this tomorrow when I have a day set aside for Mu related tasks. I want to get this work into the upcoming 1.0.2 bug-fix release. :-)

As always, many many thanks for your work (and patience). :+1:

tmontes:
@ntoll, glad to contribute. I’ll be here to address any improvement you may suggest. :)
Thank you, too.

ntoll:
@tmontes OK... I've had a chance to look through the code and have a think about things.

I'd like to start by saying thank you for all the effort you've put in and also Mary and William say "hello". :-) We all have very fond memories of meeting you in Bilbao.

Onto the feedback...

I'm playing devil's advocate here and I expect to be shot down. However, at the back of my mind is a desire for simple and easy to understand code which stands on its own as much as possible (i.e. try to avoid external dependencies / modules if possible since this makes packaging really complicated -- especially where Debian is concerned). I also have another motive: I'd like the code for Mu to be simple and obvious enough so a semi-competent teen programmer could take a look and contribute.

With this in mind...

Why not just grab the locale information in `darwin` platforms via the `defaults read -g AppleLocale` command? In Python it should be as simple as something like this:

```
import subprocess
result = subprocess.run(['defaults', 'read', '-g', 'AppleLocale'], stdout=subprocess.PIPE)
locale = result.stdout.decode('utf-8')  # locale will be something like en_GB
```

This mitigates the need for Beeware's "rubicon" (another requirement) and could simplify the reading of the locale to a conditional `if sys.platform == 'darwin':` containing the block of code above otherwise just use `locale.getdefaultlocale()` as we currently do.

It's only a small change and is relatively easy to test too (mock away `subprocess.run`).

Thoughts..?

Nota Bene: It is *highly likely* that I've missed something here. I am painfully aware that as a Linux using English speaker I'm in that sweet spot of "I don't have to worry about this sort of thing" so I'm approaching this with very naive eyes. I.e. I completely expect to be told that I've missed a really obvious edge case or missed something which means the code I've used above doesn't work.

tmontes:
First and foremost, thanks a lot for reviewing this.

> I'm playing devil's advocate here and I expect to be shot down. However, at the back of my mind is a desire for simple and easy to understand code which stands on its own as much as possible (i.e. try to avoid external dependencies / modules if possible since this makes packaging really complicated -- especially where Debian is concerned). I also have another motive: I'd like the code for Mu to be simple and obvious enough so a semi-competent teen programmer could take a look and contribute.

I hadn't thought about that before... I does make a lot sense, indeed!


> With this in mind...
> 
> Why not just grab the locale information in `darwin` platforms via the `defaults read -g AppleLocale` command? In Python it should be as simple as something like this:
> 
> ```
> import subprocess
> result = subprocess.run(['defaults', 'read', '-g', 'AppleLocale'], stdout=subprocess.PIPE)
> locale = result.stdout.decode('utf-8')  # locale will be something like en_GB
> ```
> This mitigates the need for Beeware's "rubicon" (another requirement) and could simplify the reading of the locale to a conditional `if sys.platform == 'darwin':` containing the block of code above otherwise just use `locale.getdefaultlocale()` as we currently do.
>

...and I'll raise you a `subprocess.getoutput('defaults read -g AppleLocale')`, which is even simpler for a beginner: more readable and handling all the interprocess pipe-based I/O for you. :)


(rubicon side note: very nice and lightweight library, facilitating access to many/most/all (?) of macOS native APIs without the need of binary wheels or compilers: really clever design and easy to package.)

> It's only a small change and is relatively easy to test too (mock away `subprocess.run`).
> 
> Thoughts..?

It makes a lot of sense. My approach stems from a "professional defect", striving for "by-the-book approaches", whatever that means, and clearly over-engineered, in this context.


> Nota Bene: It is _highly likely_ that I've missed something here. I am painfully aware that as a Linux using English speaker I'm in that sweet spot of "I don't have to worry about this sort of thing" so I'm approaching this with very naive eyes. I.e. I completely expect to be told that I've missed a really obvious edge case or missed something which means the code I've used above doesn't work.

It sounds as if you'd be punishing yourself for being an English speaker and not needing to "worry about this sort of thing"... Well don't. :) You are actually worrying, in dedicating time to this PR. :)
(and people should be grateful for your awareness and determination, in pushing mu forward; I am, for that and for your guidance in this PR review)

You did not miss "a really obvious edge case" as far as I can tell. You are just proposing a much simpler solution which, while being more fragile [1] and less efficient/elegant [2], is most probably adequate enough for current and future mu users on macOS.

[1] It assumes that the `defaults` command:
* Actually exists.
* Accepts those command line arguments.
* When given those command line arguments prints out a usable language code.
* Is in the user's shell PATH.

AFAICT, these have been true for as long as I have used macOS. But will they change the future?

[2] Motives:
* Spawning a process instead of using a documented API.

PS: Even though the API might also change in the future, I fathom that's much less likely as it would lead to breakage of Apple's own macOS software products, other than the OS.

**NEXT STEPS**

I will update this PR with a much simplified approach, in the line of your suggestion. Question: what is your position in having a new module to hold this code, like the `localedetect.py` I created? Keep it, rename it, belongs somewhere else, code should be included in existing module X? Hints?

PS: Merry Christmas. :)

tmontes:
Progress note:
* Updated language detection code following the simpler approach @ntoll suggested.
* Cleaned up rubicon dependency.
* Simplified tests language detection code.

I expect one test to fail on the macOS platform: `test_gettext_translation` which tests that the correct translation is set based on the `LC_ALL` env variable on any platform other than Windows. The previous code could cope with that (it either went with env variables or fell back to macOS native APIs). Current code is a little more naive.

Question:
* Improve language detection code or the failing test? (does that test even make sense?)

tmontes:
The `test_gettext_translation` test failed on macOS, as expected.

Two outstanding (groups of) questions:
* About the `localedetect.py` module I created: keep it as is, rename it, belongs somewhere else, code should be included in existing module X?
* About the test failure on macOS: previous, more complex code, did not fail here because it tried the environment var approach first via `locale.getdefaultlocale` calls, falling back to native macOS APIs. Current code takes a simpler approach and does not try that with `sys.platform == 'darwin'`. Question: improve language detection code, making it less obvious? Change the failing test? How? Skip it on macOS? Does the test even make sense at all?

Feedback welcome. Standing by.

tmontes:
I didn't standby and gave the `test_gettext_translation` test some thought. :)

From its docstring, `Test the right translation is set based on the LC_ALL environmental variable.`, it no longer makes sense: language detection details, including what might depend on the `LC_ALL` env variable, are now the responsibility of the `localedetect.py` module, which is tested elsewhere.

The test made sense before this PR, where the language detection code/tests were less isolated.

Thus, I removed it.

PS: Within the realms of testing `__init__` we may want to bring back `test_init.py` and, for example, confirm things such as a) the `localedir` is correctly determined, and/or b) that mocking `localedetect.language_code()` to a known-existing translation leads to a properly installed and working `_()` function. But I'd say that's clearly outside the scope of this PR (even though we may want to create an issue for that).

Thoughts?

ntoll:
@tmontes thank you for the refactoring in the PR and resolving the issue with the test (I agree with your approach). To answer your remaining question (what to do with the localedetect module), I think it's fine where it is for the time being. If need be, we can refactor its location as it becomes clear how/where the code is best suited. But for now, I'll merge the PR and it just remains for me to say "thank you" for your work. :-)

tmontes:
...thanks @ntoll for your persistence and guidance. :)

Issue #575, opened 2018-07-23
PyQt5/QtCore.so: undefined symbol: PySlice_AdjustIndices


slippers:
### starting mu-editor results in error
```
>mu-editor

Traceback (most recent call last):
  File "/home/projects/mu/.env/bin/mu-editor", line 7, in <module>
    from mu.app import run
  File "/home/projects/mu/.env/lib/python3.5/site-packages/mu/app.py", line 29, in <module>
    from PyQt5.QtCore import QTimer, Qt
ImportError: /home/projects/mu/.env/lib/python3.5/site-packages/PyQt5/QtCore.so: undefined symbol: PySlice_AdjustIndices
```

###  setup mu by
mkdir mu
cd mu
virtualenv -p python3 .env
activate
pip install mu-editor
mu-editor

### machine
using linux ubuntu and mint with same results

>uname -a

4.4.0-130-generic #156-Ubuntu SMP Thu Jun 14 08:53:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

>python --version

Python 3.5.2

>pip list

Package          Version   
---------------- ----------
appdirs          1.4.3     
backcall         0.1.0     
certifi          2018.4.16 
chardet          3.0.4     
cycler           0.10.0    
decorator        4.3.0     
gpiozero         1.4.1     
guizero          0.5.3     
idna             2.7       
ipykernel        4.8.2     
ipython          6.4.0     
ipython-genutils 0.2.0     
jedi             0.12.1    
jupyter-client   5.2.3     
jupyter-core     4.4.0     
kiwisolver       1.0.1     
matplotlib       2.2.2     
mu-editor        1.0.0     
numpy            1.15.0    
parso            0.3.1     
pexpect          4.6.0     
pgzero           1.2       
pickleshare      0.7.4     
pigpio           1.40.post1
Pillow           5.2.0     
pip              18.0      
pkg-resources    0.0.0     
prompt-toolkit   1.0.15    
ptyprocess       0.6.0     
pycodestyle      2.4.0     
pyflakes         2.0.0     
pygame           1.9.4     
Pygments         2.2.0     
pyparsing        2.2.0     
PyQt5            5.11.2    
PyQt5-sip        4.19.12   
PyQtChart        5.11.2    
pyserial         3.4       
python-dateutil  2.7.3     
pytz             2018.5    
pyzmq            17.1.0    
QScintilla       2.10.7    
qtconsole        4.3.1     
requests         2.19.1    
[...]

Issue #538, opened 2018-07-10
Betat 17 - QT error on Linix


jerryneedell:
I am having the same problem as in a closed issue  with Beta17
It installs fine via pip3 but gives the error below
same result using the virtual env.

``` 
Traceback (most recent call last):
  File "mu_venv/bin/mu-editor", line 7, in <module>
    from mu.app import run
  File "/home/jerryneedell/mu_venv/lib/python3.5/site-packages/mu/app.py", line 29, in <module>
    from PyQt5.QtCore import QTimer, Qt
ImportError: /home/jerryneedell/mu_venv/lib/python3.5/site-packages/PyQt5/QtCore.so: undefined symbol: PySlice_AdjustIndices
```

I see the reference to a new QT release, but it is not clear id there is recommended workaround.
I am on an Ubuntu 16.04 system
Thanks

ntoll:
OK... when on Linux, you should follow the advice in the instructions (https://codewith.mu/en/howto/install_with_python) and use a virtualenv (until such time as we manage to get Mu accepted as a Debian package -- which is work that's ongoing).

This sort of error happens when different versions of Qt are found on the same computer. A virtualenv isolates your Python apps.

The simple instructions are:

* Ensure virtualenv and Python 3 are installed.
* In a new folder somewhere safe (e.g. in `~/virtualenvs`) type: `virtualenv --python=/usr/bin/python3 muvenv` (this creates a new virtualenv called `muenv` and uses Python3.
* Activate the new virtualenv like this: `source muvenv/bin/activate` (you'll see the name of the virtualenv prepended to your shell prompt, indicating the virtualenv is active)
* With the virtualenv active, type: `pip install mu-editor`.
* Once installed, and still with the virtualenv active, start mu: `mu-editor`

It should just work :tm:. ;-)

Please do let me know how you get on.

ntoll:
OK... I'm on Ubuntu 17.10 and, following my instructions above, Mu installed fine. I'm not sure what more to suggest. :-/ Please do keep me posted.

jerryneedell:
tried on Ubuntu 16.04
```
jerryneedell@Ubuntu-Macmini:~/mu_venv$ virtualenv --python=/usr/bin/python3 muvenv
Already using interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/jerryneedell/mu_venv/muvenv/bin/python3
Also creating executable in /home/jerryneedell/mu_venv/muvenv/bin/python
[...]

PR #678, opened 2018-10-17
Update PyQt5 versions to restore Py3.5 compatibility

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/575.

New PyQt5 wheel are available and fix the current issue with missing symbols.

Replicated the issue in Python 3.5.2 on macOS and Windows and then tested this PR to ensure the new versions fix it.

I've also added Python 3.5 and 3.7 to the travis tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,18 @@ matrix:
       dist: trusty
       sudo: required
       language: python
+      python: 3.5
+    - os: linux
+      dist: trusty
+      sudo: required
+      language: python
       python: 3.6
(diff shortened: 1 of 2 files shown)


ntoll:
You :star:

darrell24015:
I seem to be having the same issues as posted in #686 #575 and #538 - will try to either update PyQt5 or uninstall/reinstall PyQt5 with 
<pip3 install --upgrade pyqt5>
or
<pip3 uninstall pyqt5>
<pip install pyqt5>

Is this what is intended to fix the issues?  I'm running Ubuntu 16.04.3, fully patched and has Python 3.5.2 installed. Thanks!

Issue #671, opened 2018-10-10
The Mu wheel ignores the Raspberry Pi check


bennuttall:
The wheel uploaded to PyPI was built on an x86 machine so the platform check obviously did not exclude packages not available for Raspberry Pi: https://github.com/mu-editor/mu/blob/master/setup.py#L25:L29

Wheels do not bundle `setup.py` - this is done at the point of building the wheel, so someone pip installing Mu on a Pi would get the wheel which didn't exclude those packages, and would fail (even if #669 didn't apply).

There's no good solution to this, as making the wheels platform-specific would mean providing one for every non-Pi platform. I have a couple of suggestions but really whether anything changes or not, the process will need documenting so people don't get tripped up by it.

One solution could be that I build a Pi platform wheel and host it on piwheels. Mu developer docs would have to suggest "To pip install on a Pi, run `sudo pip3 install mu-editor -i https://www.piwheels.org/simple`".

Another is you tell people to build from source when on a Pi, rather than using the wheel. That would be `sudo pip3 install mu-editor --no-binary mu-editor`.

Anyway, I just wanted to highlight this as an issue for anyone trying to `pip install` Mu on a Pi. Most users will be using apt, but if anyone wants an unreleased version or wants to develop for it, it will be handy to know what's going on.

ntoll:
@bennuttall thanks for this... great stuff!

How does this relate to the development install instructions in the dev docs here:

https://mu.readthedocs.io/en/latest/setup.html#raspberry-pi

???

illume:
I think environment markers are used for this these days.

An example from PEP 426: https://www.python.org/dev/peps/pep-0426/#dependencies
```
"dependencies":
  {
    "requires": ["SciPy", "PasteDeploy", "zope.interface > 3.5.0"]
  },
  {
    "requires": ["pywin32 > 1.0"],
    "environment": "sys_platform == 'win32'"
  },
  {
    "requires": ["SoftCushions"],
    "extra": "warmup"
  }
]
```

carlosperate:
While using the environment markers sounds like a good change, these are still included in the setup.py file, isn't?

If that's the case, based on @bennuttall post it wouldn't affect the built wheels unless we build for/from different platforms, so it sounds like that would still cause the same issue with the uploaded wheels.

[...]

Issue #669, opened 2018-10-10
Mu only supports Python 3.6 now?


bennuttall:
How come Mu now supports only Python 3.6 and above? Is there something in Mu which relies on 3.6? Raspbian Stretch has Python 3.5. If it's not a hard requirement we can build it anyway (and the version check should be removed), but if it really does require Python 3.6 we will be stuck on v1.0.0.

ntoll:
Eyup @bennuttall,

It's to do with latest Qt (5.11) wheels only working against 3.6+ :-/ Riverbank are aware, and please reach out to the to fix the problem. (The error we see is that when using Python 3.5 Qt complains about an unknown symbol that was introduced in Python 3.6... i.e. they've broken their complication/build against Python 3.5).

Does this help?

AFAICT, you could remove the requirement and build things anyway since you're relying on Debian packaging of Qt here, right?

bennuttall:
Ah ok, that makes sense. Should be fine for Pi. :+1:

LuRsT:
Can this Issue be closed?

carlosperate:
Does that mean the current Pi deb build process will include an additional step to apply a patch to remove those lines?

carlosperate:
Related issues: https://github.com/mu-editor/mu/issues/575 and https://github.com/mu-editor/mu/issues/538

I've been looking into this and I think the issues is actually with versions of Python below 3.5.4. I couldn't replicate with Python 3.5.6 and 3.5.4, and then I noticed most of the report we had were on Python 3.5.2, and after testing that version I was finally able to replicate in macOS and Windows.

After a google search I've found this, so it looks the the symbol was introduced in 3.5.4:
https://riverbankcomputing.com/pipermail/pyqt/2018-July/040514.html

Thankfully Riverbank has released new wheels a couple of weeks ago, so I've tested them and we can remove the setup.py safeguard: PR https://github.com/mu-editor/mu/pull/678

carlosperate:
The imposed restriction has been removed with PR #678, so I'll close this issue.

There is still more ongoing work to improve how Mu deals with dependencies in Raspbian, discussion can be found in issue https://github.com/mu-editor/mu/issues/671.

bennuttall:
Mu now depends on Black, which requires Python 3.6 and above.

ntoll:
[...]

PR #680, opened 2018-10-19
Add environmental markers to setup.py and requirements.txt

carlosperate:
From the discussion in https://github.com/mu-editor/mu/issues/671.

This also removes the need to have a separate requirements_pi.txt.
If accepted we also need to update this page as well to change the requirements.txt mention: https://mu.readthedocs.io/en/latest/setup.html

I've tested `pip install -e .` and `pip install -r requirements.txt` in macOS and Raspberry Pi.
I've also created a wheel in macOS, transferred it to raspberry pi and that works as well.


carlosperate:
The current travis failure is due to an issue in briefcase: https://github.com/pybee/briefcase/pull/169

carlosperate:
Briefcase has been updated and the CI passes.

bennuttall:
:+1:

ntoll:
OK... so regarding docs. The "blessed" way to install Mu on a Raspberry Pi is via the OS package and NOT via pip. I'm just updating the dev docs via this branch to highlight this problem, although the instructions found therein should still just work (since step 1 is to install the PyQt dependencies via ``apt-get``).

carlosperate:
Great, thanks Nicholas! And thanks for updating the docs as well.

About what I mentioned in my last message in https://github.com/mu-editor/mu/issues/669#issuecomment-464752603 about adding a check on import to detect missing PyQt5 dependencies. Taking a pragmatic look, I agree that we should leave that as it is right now as the docs do have all the installation methods. If we start getting users hitting the same error via pip install on the Raspberry Pi we can then reconsider, but until then no need to guess, as alpha/beta testing we'll let us know.

ntoll:
@carlosperate exactly... alpha/beta testing will highlight how visible this `pip` related problem may be for users on Raspberry Pi.

Issue #683, opened 2018-10-21
Blank line appears in program output


tjguk:
(This was spotted by a Year 8 group I was teaching who took the trouble to make sure it could be reproduced and in the same place).

Take the code in https://gist.github.com/tjguk/61bd3e38c56fb5e1467381e5a9aa117e and run it in Mu. Between number 57 & 58 a blank line appears. It always appears and it always appears there. If I change the program to run without fizzing or buzzing, no blank line appears. Which leads me to think it's a buffer-length thing.

I've only tried on Win10 / Mu 1.0.1 from the installer

![image](https://user-images.githubusercontent.com/580948/47265528-194e5080-d521-11e8-996f-d0e81ad9922e.png)

tjguk:
Relevant logs attached though I'm not seeing anything especially useful

[mu-logs.txt](https://github.com/mu-editor/mu/files/2499045/mu-logs.txt)

tim-mccurrach:
I'm not 100% sure, I can't run the code cos not with laptop at the moment, but this looks like it might be related to issue #674.

I think that the '\r\n' that's appended to whatever is printing, gets split across the 256 byte chunk in `read_std_out` and you end up with `... \r` and `\n...` giving two new lines instead of one. I'm nearly done with a pull request to fix this, that I'll probably submit this evening :)

tim-mccurrach:
I think this is related to issue #674 

The `\r\n` on the end of a print gets split (by the 256 byte chunks) into a `...\r` and a `\n...` giving two new lines instead of one. I'm working on a PR for this which I'll submit this evening, I've just got to write the tests :)

sorry, if this is a duplicate reply. I just replied but it didn't seem to go through??

tjguk:
Thanks @tim-mccurrach that certainly sounds plausible and along the lines of what I expected. I'll look out for your PR whenever and test it against our code.

carlosperate:
Looks like #723 has fixed the issue as I am not able to replicate any more :)

Issue #674, opened 2018-10-14
Fix byte decoding of stdout from subprocess


ntoll:
Try this in Python3 mode:

```
print(b'\xe2\x94\x80\xe2'.decode('utf-8'))
```

The code which led to this problem is:

```
def menu():
    print('┌──────────────────────────┐')
    print('│ The Treasure Hunt │')
    print('│ 1. Play │')
    print('│ 2. Help │')
    print('│ 3. Quit │')
    print('└──────────────────────────┘')

menu()
```

The problem is byte stream from the subprocess only reads 256 bytes at a time and the byte is thus corrupted.

tim-mccurrach:
I’ve written a fix for this problem, but in the process discovered a new related bug. Potential fixes for the new bug, might circumnavigate the original bug so I’ve held off on a pull request for now. (I also wasn’t sure about opening a separate issue, as they both relate to the same fundamental issue, and this would risk fixing 2 separate symptoms rather than the original cause.)

**The new bug** 
The following code produces differing results when run multiple times (on v1.01, on windows 64bit both production and development versions):
```
a="a"*300
print(a)
```
Some (most) of the time it will print the full string, along with the new line plus ">>>" prompt. However, other times, it will only print a certain section of the string and stop short (presumably the first 256 characters). In such cases pressing enter again will cause the remainder (or for really long strings, the next section of the string) to be printed, and there will be multiple ">>>" prompts appended to the end of the string. Behaviour becomes increasingly unpredictable when we replace 300 with larger numbers.

It seems like `read_from_stdout` does not always get called when there is still data ready to be read. For what it’s worth, the code below fixes the original problem, but might not be necessary if a fix to the above didn’t involve reading in chunks:

```
def detect_missing_utf8_bytes(self, args):
    """
    Return the number of bytes missing from the end of a utf-8 encoded
    byte string. Return -1 if the byte string is ill-formed.
    """
    if args[-1] >> 7 == 0:  # Single byte character
        return 0
    for continuation_bytes in range(min(3, len(args))):
        if not args[-1 * continuation_bytes - 1] >> 6 == 2:
            break
    else:
        return -1
    initial_byte = args[-1 * continuation_bytes - 1]
    if initial_byte >> 5 == 6:  # 2-byte character
        character_length = 2
    elif initial_byte >> 4 == 14:  # 3-byte character
        character_length = 3
    elif initial_byte >> 3 == 30:  # 4-byte character
        character_length = 4
    else:
        return -1
[...]

PR #685, opened 2018-10-23
Fix 674, 683 and problem with 256 byte chunks

tim-mccurrach:
Addresses #674 and #683, as well as the fact that really long strings are not guaranteed to print to completion. `readyRead` is not emitted recursively, only when the data  is first added to stdout, so if `read_from_stdout` is still working whilst this signal is emitted, or if it is emitted following a block of data of more than 256 bytes being added to stdout, not all of the data is subsequently read. 

To address this, if more than 256 bytes is written, then after reading and appending, a queued data-flood signal is emitted if there are still bytes available to be written, which then reads the next 256 byte chunk. Whilst this process is going on, normal signals from `readyRead` are ignored for performance.

I guess this is quite a lot of changes, and I am aware we want to keep the code for mu simple, and that there might be better ways of doing this? 
Other questions:

- The function `detect_missing_utf8_bytes` doesn’t really have to belong to `PythonProcessPane`, is this an okay place to put it?

- At the moment if the end of a 256 byte string is genuinely ill-formed, as apposed to just cut in half in an inconvenient place, it is logged, and the whole ‘chunk’ is ignored. Is this an unnecessary complication? Especially as it doesn't pick up all ill-formed byte sequences – I know there is an aim to keep the codebase simple.

- The code:

```
            self.append(data)
            self.on_append_text.emit(data)
            cursor = self.textCursor()
            self.start_of_current_line = cursor.position()
```
is now repeated in the function `read_from_stdout`, `on_data_flood`, and `on_process_halt`. Would it be better to put it in a separate function?

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -915,6 +915,56 @@ class PythonProcessPane(QTextEdit):
             history_item = self.input_history[history_pos]
             self.replace_input_line(history_item)
 
+    def detect_missing_utf8_bytes(self, byte_string):
+        """
+        Return the number of bytes missing from the end of a utf-8 encoded
+        byte-string. Return -1 if the byte-string is ill-formed.
+        """
+        if byte_string[-1] >> 7 == 0:  # Single byte character
+            return 0
+        for continuation_bytes in range(min(4, len(byte_string))):
+            if not byte_string[-1 * continuation_bytes - 1] >> 6 == 2:
+                break
+        else:
+            return -1
+        initial_byte = byte_string[-1 * continuation_bytes - 1]
+        if initial_byte >> 5 == 6:  # 2-byte character
+            character_length = 2
+        elif initial_byte >> 4 == 14:  # 3-byte character
+            character_length = 3
+        elif initial_byte >> 3 == 30:  # 4-byte character
+            character_length = 4
+        else:
+            return -1
+        result = character_length - continuation_bytes - 1
+        if result >= 0:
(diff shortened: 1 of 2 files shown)


carlosperate:
Hi @tim-mccurrach,
Quick note, I had a quick look at the travis failure and I believe it's because `assert_called_once` was introduced in Python 3.6 and so it fails in the 3.5: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

tim-mccurrach:
@carlosperate thankyou very much for your help :) I was wondering why it wasn't passing for 3.5, but couldn't quite figure it out. I've updated that and hopefully it'll pass properly this time :)

treese:
This problem may be related (if not, I can open a new issue): With Mu 1.0.1 on Mac OS X 10.13.6 (High Sierra), the following loop only prints 256 characters before pausing. Hitting return continues the printing.
```
for x in range(0,100):
    print("{0:03d}".format(x))
```
I don't have a convenient way to try building the patched code to test it. Does this still happen with the pull request? Is it a different issue?

tim-mccurrach:
@treese It certainly sounds like its related (especially given that it prints 256 characters). I won't be able to test your code against the patch until quite late this evening, but I should imagine this is caused by the same thing, yes.

tim-mccurrach:
@treese I have run your code in the branch for this PR, and it does indeed fix the problem :)

PR #687, opened 2018-10-23
Make pretty with Black

ambv:
Adds a new button that uses [Black](https://github.com/ambv/black/) to format code.

## Benefits
* Kids will learn good Python code style by example.
* Kids don't have to be distracted by manual tidy formatting while they are trying to understand fundamentals of programming.
* Teachers are able to look at all code the same way.
* Teachers and students are able to reflow code to match their chosen zoom level.
* Formatting can sometimes expose bugs (a line like `a = 1,` will get formatted to `a = (1,)`; or `.1` will get formatted to `0.1`).
* World domination of the Black code style by teaching the next generation of programmers the One True Style.

## Risks
The following are speculative problems, up to you to consider if they are serious issues and whether there could be ways to mitigate them.

* Kids will get reliant on auto-formatting.
* Some of applied automatic formatting will not be as beautiful as hand-formatting.
* Some formatting will raise questions from students (like number and string quote normalization).
* Black requires Python 3.6+ to run.
* On a 1280x800 screen the "Make Pretty" button used up the last available slot in the Pygame Zero mode.

## Disappointments

I started out with a black diamond icon which had the following properties:
* it was, well, **black**
* it was a diamond (→ pretty)
* "Black Diamond" is one of my favorite jazz piano standards

However, the icon stood out like a sore thumb. It brought too much extra attention. So I toned it down. There's two darker versions of the button in the pull request (make-pretty-light and make-pretty-dark) for your consideration.

## Demo
https://twitter.com/llanga/status/1054750407468163072

## This is a work in progress

As comments in code state, this is missing important things:
- [ ] user has to be able to undo the formatting
- [ ] max line length should be dependent on the available character width in the editor
- [ ] the cursor needs to be read before the formatting and placed in the right spot after the formatting
- [ ] there are no translations of the button and the tooltip

I will need help with the above. I'll deal with the Black-side of providing information what the updated coordinates of the cursor should be. But I don't know how to alter the cursor position programatically.

I'm more confused at how to get available character width from the editor pane.


mu/logic.py
```diff
@@ -410,7 +410,7 @@ def check_pycodestyle(code):
     os.close(code_fd)
     save_and_encode(code, code_filename)
     # Configure which PEP8 rules to ignore.
-    ignore = ('E121', 'E123', 'E126', 'E226', 'E302', 'E305', 'E24', 'E704',
+    ignore = ('E121', 'E123', 'E126', 'E226', 'E203', 'E302', 'E305', 'E24', 'E704',
```
ambv: This is necessary since Black formats slices according to PEP 8, and not according to what pycodestyle prefers:

screenshot 2018-10-17 12 02 29

ambv:
This is now green except for Python 3.5 which is unable to install Black. I will let you decide what to do with that.

ntoll:
@ambv I'm going to make some minor changes to this before merging:

* Fix the crash when black gets mal-formed Python so there's a friendly message.
* "Make Pretty" -- hmm... for readability, how about just "Tidy"..? (As in it'll tidy up your messy looking code). ;-) [In general, single words are better for kids when it comes to reading a UI and "pretty" is also, sadly, a loaded term].
* I'll try to address / answer / work around the missing/problematic things you list.
* If we move to "Tidy"... then the diamond icon probably won't make much sense. I'll suggest some others and we can work it out. :-)
* As always, all the above is up for grabs and ideas, constructive critique and feedback most welcome (basically, I'm open minded and up for positive discussions).

ntoll:
Hmm... so Python 3.5, we support it only because PyQt do. I'll ask the wider user base / community what they think about dropping it (I think we should if we can).

ntoll:
https://twitter.com/ntoll/status/1096813376171360256

ntoll:
OK... so I've managed to make it work with undo, added some guards for invalid code and made sure there's a message when black completes that also mentions using CTRL-Z to undo the changes.

I've spent a LOT of this evening trying to figure out how to get the current character width for the widget. I've thrown in the towel since there's no built in function in the QScintilla widget, nor do any of the attributes/methods that have anything to do with "width" (and most of them are nothing to do with the actual width of the window) work in any other unit of measurement than pixels. :-/ As Homer Simpson once said, "If something is hard to do, it's probably not worth doing..." ;-) In this instance I believe your default of 88 characters width is a good decision and most pragmatic in this situation.

As for moving the cursor to the correct token in the code, as your comment in the code states, it would require changes in black.

In the morning I'm going to ensure the tests and code work correctly in Python 3.5 (i.e. this feature is ignored) and 3.6+. Once fixed I'll merge into master in preparation for the 1.1.0-alpha.1 release.

ntoll:
OK... I've merged this with the changes I described above..! :tada:

This is very much a feature that can be updated / modified and changed as we see fit over the alpha/beta process. Right now, I'm anxious to get the code in front of users so they can test it and give feedback.

@ambv thank you for your work on this. The teachers I've shown this to all love it..! :-)

Issue #707, opened 2018-11-14
stdout is truncated at 512 characters


bennuttall:
I've found that and stdout is limited to 512 characters.

```python
print('|'*511 + 'x')
```

works as expected. I see 511 pipes followed by `x`:

```
|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||x
```

However:

```python
print('|'*512 + 'x')
```

I only see 512 pipes, no `x`:

```
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
```

This seems to be the cause of the pgzero issue I opened earlier: #706.

tjguk:
Thanks, Ben. This seems to be a popular issue today -- and there's a PR #685 currently waiting to be tested to fix this

ntoll:
Fixed by #723

Issue #706, opened 2018-11-14
pgzero mode error shows truncated stack trace


bennuttall:
The pgzero mode seems to truncate the stack trace when outputting errors which lie within pgzero.

The following one-line example is enough to demonstrate:

```python
actor = Actor('alien')
```

This will work fine if `images/alien.png` exists. If there's no `images` directory, then `pgzrun` from the terminal will give a meaningful error message:

```python
ben@gunter:~/Projects/bennuttall/pgz-campug {master} $ pgzrun game.py 
/usr/lib/python3/dist-packages/apport/report.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import fnmatch, glob, traceback, errno, sys, atexit, locale, imp
Traceback (most recent call last):
  File "/usr/local/bin/pgzrun", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/pgzero/runner.py", line 92, in main
    exec(code, mod.__dict__)
  File "game.py", line 4, in <module>
    alien = Actor('alien')
  File "/usr/local/lib/python3.6/dist-packages/pgzero/actor.py", line 88, in __init__
    self.image = image
  File "/usr/local/lib/python3.6/dist-packages/pgzero/actor.py", line 103, in __setattr__
    return object.__setattr__(self, attr, value)
  File "/usr/local/lib/python3.6/dist-packages/pgzero/actor.py", line 218, in image
    self._orig_surf = self._surf = loaders.images.load(image)
  File "/usr/local/lib/python3.6/dist-packages/pgzero/loaders.py", line 120, in load
    self.validate_root(name)
  File "/usr/local/lib/python3.6/dist-packages/pgzero/loaders.py", line 102, in validate_root
    subpath=self.subpath, type=self.TYPE, name=name
KeyError: "No 'images' directory found to load image 'alien'."
```

However, Mu gives a truncated stack trace which contains no useful information for the user:

```python
/usr/lib/python3/dist-packages/apport/report.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import fnmatch, glob, traceback, errno, sys, atexit, locale, imp
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
[...]

PR #708, opened 2018-11-15
Alternative cleaner approach to #685

tim-mccurrach:
Addresses the same issues as #685 (#707 ,#706 etc.) but does it in a much cleaner fashion.

The whole problem of having to deal with split utf-8 strings (issue #674 ), or other sequences such as `\r\n` (issue #683 ) can be avoided completely if we read the data line by line rather than in 256 byte chunks. This turned out to be significantly slower for actually printing something like a for loop, so I opted for reading a 256 byte chunk, and then some more up until the end of the current line. This makes the solution a lot shorter and simpler, and the code-base more accessible for beginners.

Since it relies on new-line characters, mu would still freeze in the following situation:
```
while True:
    print('Hello', end='')
```
However, I’m struggling to think of a situation where this could happen by accident. I think any beginner who is advanced enough to think about using the `end` keyword argument, would also be advanced enough to identify an infinite while loop, and understand what will likely happen. So whilst it doesn't cover every single case, might be a better solution than #685 . Simple is better than complex?

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -596,6 +596,7 @@ class PythonProcessPane(QTextEdit):
     """
 
     on_append_text = pyqtSignal(bytes)
+    read_buffer = pyqtSignal()
 
     def __init__(self, parent=None):
         super().__init__(parent)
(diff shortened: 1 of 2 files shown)


tjguk:
Well I can confirm that it solves the problem where the output pane of a program is truncated short. The problem showed up in a session today when we were displaying a long list of words. Without the fix, the output stopped after a few hundred characters; with the fix -- after a delay -- all 360,000+ words were apparently displayed!

tjguk:
Thanks @tim-mccurrach. From my point of view, this really isn't a showstopper. It's the common experience that outputting a large amount of text will tend to delay things, and this certainly wasn't an unconscionable delay -- just a few seconds.

(Of course if you can come up with something better, that would be good too!)

ntoll:
Hi.

Thank you for your contribution and apologies for taking so long to respond.

I think I have a single line fix for this which fixes all the related issues and I've stuck it into #723 which I'll merge immediately so it'll get into the 1.0.2 point release.

Thanks, once again for all the work that went into this.

tim-mccurrach:
👍  glad to. Learnt a lot doing so :)

PR #710, opened 2018-11-20
Fix typos for locale de_DE

wayoda:
Most of the changes refer to misspellings. 
Line `mu/logic.py:79` is from the file that defines the MOTD. I guess __import this__ should not be translated in any language.


ntoll:
Danke / Thank you! :-)

PR #723, opened 2018-12-11
Fix reading of stdout from Python subprocess.

ntoll:
This single line change fixes #685, #683, #708, #707, #706, #674 (I've tested all the error conditions, and they now work with this change). I've also updated the tests to ensure the single line change is checked.

How does it work?

As @tim-mccurrach correctly pointed out, the `readyRead` signal is only emitted when *new* data is added to the buffer. Since the `read_from_stdout` method connected to this signal only reads 256 bytes, there may be more data left on the buffer and this will never get displayed since the process may have exited and there's no new data to trigger `readyRead`.

This single line creates a new QTimer which schedules a single shot event in the immediate future (i.e. next time round the event loop) to read again from stdout. If there's no data left in the buffer, the timer won't be rescheduled. If there is, it'll keep scheduling itself to read the next 256 bytes.

Nice and simple..! :-)

I'm going to merge this *right now* since it fixes a large number of bugs and I'd like to get it out for 1.0.2.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -611,6 +611,7 @@ class PythonProcessPane(QTextEdit):
         self.input_history = []  # history of inputs entered in this session.
         self.start_of_current_line = 0  # start position of the input line.
         self.history_position = 0  # current position when navigation history.
+        self.stdout_buffer = b''  # contains non-decoded bytes from stdout.
 
     def start_process(self, script_name, working_directory, interactive=True,
                       debugger=False, command_args=None, envars=None,
@@ -783,10 +784,16 @@ class PythonProcessPane(QTextEdit):
         """
         data = self.process.readAll().data()
         if data:
-            self.append(data)
-            self.on_append_text.emit(data)
-            cursor = self.textCursor()
-            self.start_of_current_line = cursor.position()
+            while True:
+                try:
+                    self.append(data)
+                    self.on_append_text.emit(data)
+                    self.set_start_of_current_line()
+                    break
+                except UnicodeDecodeError:
+                    # Discard problematic start byte and try again.
+                    # (This may be caused by a split in multi-byte characters).
+                    data = data[1:]
 
     def parse_input(self, key, text, modifiers):
         """
@@ -818,6 +825,7 @@ class PythonProcessPane(QTextEdit):
                 if halt_flag:
                     # Clean up from kill signal.
                     self.process.readAll()  # Discard queued output.
+                    self.stdout_buffer = b''
                     # Schedule update of the UI after the process halts (in
                     # next iteration of the event loop).
                     QTimer.singleShot(1, self.on_process_halt)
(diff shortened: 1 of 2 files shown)


tim-mccurrach:
@ntoll I may well have missed something here, but I think there will still be problems with utf-8 encodings, since the 256 byte chunk might split a 2,3 or 4 byte character in two and result in an encoding error.

However, a fix to this would be after the line `data = self.process.read(256)` you could add on up to the rest of the line. I think (I can't remember precisely off the top of my head) QProcess has a readline method. So you could add something like data += self.process.readLine(), and then this will avoid this error.

ntoll:
@tim-mccurrach ooh... good point. I've been trying (and failing) to make the split-a-2-byte-character problem happen (like I said, I ran all the examples of problems I could find, and they all worked for me).

tim-mccurrach:
hmm, maybe "a"*254 + some 4 byte character

tim-mccurrach:
I'm at work atm, so can't test this, but I think this won't address #683 since a split `\r\n` won't raise a UnicodeDecodeError. I guess that's not the worst problem in the world though, and won't come up that frequently.

ntoll:
@tim-mccurrach +1. I'm in London at the moment and won't be able to check this code on a Windows machine until tomorrow. However, when I tried @tjguk's fizzbuzz example on my Linnux based laptop it worked fine for me.

tim-mccurrach:
Trying the fizzbuzz example, I still get the additional line (although only 90% of the time).

On a seperate note: on my (2 year old, but not super slow) laptop - it can be quite clunky in an infinite/long loop, and although the interface doesn’t become unresponsive but definitely becomes less responsive. I daresay the effect may be worse on the old school computers we have.

I think this is because when there is a sufficiently big backlog each `readyRead` signal starts a new string of `QTimer. singleShot`’s and these build up meaning lots of calls to `read_from_stdout` all at the same time clogging up the event loop.

Blocking all calls to `read_from_stdout` that come via readyRead, whilst the function is already calling itself via the `QTimer` event alleviated this. There may well be a cleaner way to do it but something like the following made things a lot smoother:

```
def attempt_read_from_stdout(self):
    """
    Check that data from stdout is not already being processed
    """
    if not self.currently_reading_from_stdout:
        self.currently_reading_from_stdout = True
        self.read_from_stdout()
        
def read_from_stdout(self):
    """
    Process incoming data from the process's stdout.
    """
    data = self.stdout_buffer + self.process.read(256)
    if data:
        self.something = True
        try:
            self.append(data)
            self.on_append_text.emit(data)
            self.set_start_of_current_line()
            self.stdout_buffer = b''
        except UnicodeDecodeError:
            self.stdout_buffer = data
        QTimer.singleShot(2, self.read_from_stdout)
    else:
        self.currently_reading_from_stdout = False
```
(obviously then connecting readyRead to attempt_read_from_stdout)

I appreciate that this since the interface doesn't become unresponsive this is not a critical issue, and perhaps cleaner simpler code should be a greater priority.

ntoll:
@tim-mccurrach yes... you suggestion to mitigate an explosion of QTimer single shots is a good one and I'll add it to the PR.

ntoll:
OK... merging this and will fix Windows things in further code.

Issue #697, opened 2018-11-01
Unable to import file from mu_code folder (Windows)


LargeKnome:
[Log.txt](https://github.com/mu-editor/mu/files/2538550/Log.txt)
Here is my full log, as requested.

carlosperate:
Hi @LargeKnome,

Thank you for the report, but without a bit more info we don't really know what to look for.

Could you explain step by step what you were trying to do, what did you expect to happen, and what did Mu do?

tjguk:
@carlosperate for info this was the outcome from a discussion over on Gitter starting around here: https://gitter.im/mu-editor/general?at=5bdae32ea9c55b25748d40c1

It's not an awful lot clearer there as it's difficult to get a clear picture on an online to-and-fro. But it does seem to be somehow tied to Mu having been installed in an unusual location to start with and/or the files being picked up from somewhere unusual and then moved over.

I suspect (in fact I half remember from when I was doing the load/save testing) that we have odd corner cases where two files have the same name but in different directories. But I haven't put any time into trying to reproduce that.

carlosperate:
Thank you @tjguk.
This is definitely https://github.com/mu-editor/mu/issues/662, we need to fix including `mu_code` into the path for Windows. It works in debug, but not in run.

carlosperate:
As #662 has been fixed this should be resolved, so I'll close this issue.
If anybody is still experiencing this problem please reopen this issue or #662.
Thanks for the report @LargeKnome!

Issue #662, opened 2018-09-24
mu_code not included in sys.path on Windows 10


malbarbo:
Running

```python
import sys

for p in sys.path:
    print(p)
```

on Linux prints

```
/home/malbarbo/mu_code
/usr/lib/python35.zip
/usr/lib/python3.5
/usr/lib/python3.5/plat-x86_64-linux-gnu
/usr/lib/python3.5/lib-dynload
/home/malbarbo/.local/lib/python3.5/site-packages
/usr/local/lib/python3.5/dist-packages
/usr/lib/python3/dist-packages
```

but on Windows 10 it prints

```
C:\Users\Marco\AppData\Local\Mu\Python\python36.zip
C:\Users\Marco\AppData\Local\Mu\Python\
C:\Users\Marco\AppData\Local\Mu\pkgs
```

The `mu_code` path is not included on windows, which means that files in `mu_code` dir cannot be imported.

ZanderBrown:
Duplicate of #652

carlosperate:
Also, as mentioned in https://github.com/mu-editor/mu/issues/612 there is a difference between debug and run.

Run:
```
C:\Program Files\Mu\Python\python36.zip
C:\Program Files\Mu\Python
C:\Program Files\Mu\pkgs
```

Debug:
```
c:\users\microbit-carlos\mu_code
C:\Program Files\Mu\Python
C:\Program Files\Mu\pkgs
C:\Program Files\Mu\pkgs\IPython\extensions
```

ntoll:
@carlosperate I can't see anything obvious in Mu code that adds the `mu_code` directory to sys.path when in the debugger. I guess this must be something Python does when it's being driven by the `bdb` (base debugger) class.

In any case, I've checked the above and can reproduce, so I'll apply a quick fix to do two things:

* Add the `mu_code` directory to the PYTHONPATH
* If the directory containing the Python script isn't `mu_code` add it to PYTHONPATH too.

This should ensure code in both `mu_code` and the current location of the script can be imported.

I have a feeling this is down to the way `pynsist` packages/configures Mu on Windows. I believe the Python instance is run in a locked down mode, thus excluding the more obvious directories from appearing on the path.

carlosperate:
Yes, when I had a look I didn't see anything obvious either that would explain the difference on Windows.
The fix you describe sounds good, and you also bring a good point about testing in both the pip installed Mu and the pynsist installer version as well, in case there are more differences.

[...]

Issue #612, opened 2018-08-15
Debugger Failes if numpy is importet


jrast:
The import of numpy failes if the file is debugged instead of run.

Content of Testfile:

````testfile.py
import numpy as np
print(np.arange(10))
````

If i run this file by clicking "Run", I get the expected output:
```
[0 1 2 3 4 5 6 7 8 9]
````

But if I run the file by clicking "Debug", followed by "Continue", I get the following output:
````
exception: Traceback (most recent call last):
  File "D:\MuEditor\pkgs\mu\debugger\runner.py", line 494, in run
    debugger._runscript(filename)
  File "D:\MuEditor\pkgs\mu\debugger\runner.py", line 469, in _runscript
    self.run(e)
  File "bdb.py", line 431, in run
  File "<string>", line 1, in <module>
  File "c:\users\jrast\mu_code\test1.py", line 1, in <module>
    import numpy as np
  File "D:\MuEditor\pkgs\numpy\__init__.py", line 142, in <module>
    from . import add_newdocs
  File "D:\MuEditor\pkgs\numpy\add_newdocs.py", line 13, in <module>
    from numpy.lib import add_newdoc
  File "D:\MuEditor\pkgs\numpy\lib\__init__.py", line 24, in <module>
    from .financial import *
  File "D:\MuEditor\pkgs\numpy\lib\financial.py", line 15, in <module>
    from decimal import Decimal
ModuleNotFoundError: No module named 'decimal'


---------- FINISHED ----------
exit code: 0 status: 0
````

Technical Information:
Mu: 1.0.0, fresh install (64bit)
Operating System: Windows 7, 64bit

Complete Log: [mu.log](https://github.com/mu-editor/mu/files/2290433/mu.log)

carlosperate:
Hi @jrast, thank you for the bug report. I wasn't able to replicate, this in macOS. I'll see if next time I'm using Windows I remember to try to replicate there.

In the mean-time, would it be possible you have a `decimal.py` file in the same directory than your test file? or in the mu_code directory?
If there is a file named like that in any of the directories that Python looks for modules to import it's possible it's trying to import that one instead of the one provided by numpy.

tim-mccurrach:
Not a solution I'm afraid, but perhaps this might point someone who knows more than me in the right direction. I ran the code (on windows 10) and it also works fine, when just run, but get the same error when run with debug mode. Interestingly though, it works absolutely fine when run from my development version on the same machine. 

[...]

Issue #581, opened 2018-07-27
Debug not working with requests module


g70104sm:
When debugging with Mu I've encountered problem while debugging program using 'requests' module.
To reproduce:
1. Write a program in Python3 mode:
import requests
print("hello")
2. Click debug, then Step Over

Expected: 
debug moves on to next line

Actually Happened (debug window output):
exception: Traceback (most recent call last):
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\mu\debugger\runner.py", line 494, in run
    debugger._runscript(filename)
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\mu\debugger\runner.py", line 469, in _runscript
    self.run(e)
  File "bdb.py", line 431, in run
  File "<string>", line 1, in <module>
  File "c:\users\smiotk\mu_code\hello.py", line 1, in <module>
    import requests
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\requests\__init__.py", line 43, in <module>
    import urllib3
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\urllib3\__init__.py", line 8, in <module>
    from .connectionpool import (
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\urllib3\connectionpool.py", line 11, in <module>
    from .exceptions import (
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\urllib3\exceptions.py", line 2, in <module>
    from .packages.six.moves.http_client import (
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\urllib3\packages\six.py", line 203, in load_module
    mod = mod._resolve()
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\urllib3\packages\six.py", line 115, in _resolve
    return _import_module(self.mod)
  File "C:\Users\smiotk\AppData\Local\Mu\pkgs\urllib3\packages\six.py", line 82, in _import_module
    __import__(name)
ModuleNotFoundError: No module named 'http'

---------- FINISHED ----------
exit code: 0 status: 0

ENVIRONMENT:
Win 10 
Mu 1.0.0

Thanks for creating Mu, I love its simplicity!

tjguk:
Hand-wavingly, I suspect that this may be relevant: http://mitogen.readthedocs.io/en/stable/shame.html#six

ntoll:
@g70104sm thank you for reporting this. Ouch.

[...]

PR #726, opened 2018-12-21
Ensure correct sys.path contents in Windows, for sub-processes launched by Mu

ntoll:
This PR introduces three changes.

* A fix for #697, #662 so the expected / correct paths are included when Python 3 mode launches / debugs a user's script. This works by attempting to drop a `mu.pth` file in the directory Python/Windows reports as the `USER_SITE` (a directory that's usually writeable by the user, but which Python checks for configuration files when it starts up). If this step fails, Mu logs the error and continues as it used to behave (but the user will thus encounter the import errors reported in the referenced issues). The problem was caused by the version of Python used by the Windows installer for Mu running in "isolated" mode (i.e. protected from other versions of Python that may be installed on the system).
* A fix for #612 and #581 so the path to `pythonXX.zip` (where XX is Major/Minor numbers of the Python version) is included in the `sys.path`. This is simply an injection of the location of `python36.zip` into `sys.path` before the debugger kicks off the script-to-be-debugged.
* A minor fix to ensure the zoom-level from the last session is restored on application start. This is a trivial update.

While a relatively small PR, the path related issues took a while to figure out since there were various (and ultimately unusable) potential approaches which needed to be checked before I could be sure I had a "good" solution. I believe the approach taken here meets both the "it's simple" and the "it works" heuristics. However, I'm painfully aware that I'm no Windows expert nor do I have a Windows machine available to me which would allow me to check some of the more "weird" configurations Mu is likely to encounter (especially in school based environments). 

As a result, I'd love some feedback from Windows-related folks before merging. :-)

cc/@tjguk and @tim-mccurrach

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -194,6 +194,12 @@ class Window(QMainWindow):
     load_theme = pyqtSignal(str)
     previous_folder = None
 
+    def set_zoom(self):
+        """
+        Sets the zoom to current zoom_position level.
+        """
+        self._zoom_in.emit(self.zooms[self.zoom_position])
+
     def zoom_in(self):
         """
         Handles zooming in.
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -676,6 +676,8 @@ class PythonProcessPane(QTextEdit):
             # working_directory, the directory containing the script to run.
             try:
                 if site.ENABLE_USER_SITE:
+                    # Ensure the USER_SITE directory exists.
+                    os.makedirs(site.getusersitepackages(), exist_ok=True)
                     site_path = site.USER_SITE
                     path_file = os.path.join(site_path, 'mu.pth')
                     logger.info('Python paths set via {}'.format(path_file))
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -17,6 +17,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import sys
+import site
 import os
 import re
 import platform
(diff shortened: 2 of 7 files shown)


tim-mccurrach:
With regard issues #612 and #581 everything worked well in all of the situations I tried.

With regard importing a module for `mu_code` I found that running a development version inside a virtualenv, everything worked well in both debug mode and running the file as normal. However, when I ran `make win64` and tried the same thing on the version I installed, I found that in debug mode I could import from `mu_code`, but when I ran the code normally I could not. (Both the script I was running and the module I tried to import were in `mu_code`).

I'm not exactly sure why this is the case. Printing out `sys.path` in each of the various situations gave the following:

- In the dev version inside the virtualenv, the only difference in `sys.path` was that in debugger mode there was an additional path `...lib\site-packages\IPython\extensions`

- In the installed version, again the only difference between debug and normal mode was the additional IPython path.

- `mu_code` was in sys.path for the dev version.

- In the installed version `mu_code` was not in `sys.path`, for debug or normal. Although I could import from `mu_code` in debug mode. 

I'm sorry this probably isn't the best news for your Saturday morning. If there's anything you want me to try on my machine just let me know.

ntoll:
@tim-mccurrach can you run the following script and copy the full output into a comment on this issue. Can you also attach your log file too please..?

```
import sys
import os
import site

for p in sys.path:
    print(p)

print(site.ENABLE_USER_SITE)
print(site.USER_SITE)
print(os.path.exists(os.path.join(site.USER_SITE, 'mu.pth')))
```

Thank you!

tjguk:
@ntoll  Just from a quick perusal of the code... At least on my system, site.USER_SITE doesn't exist by default. (You can actually see this if you run python -msite). And AFAICT we're not creating it. I'm actually surprised we're not seeing some kind of filesystem error.

I'll try to have more or a look later today or tomorrow

tim-mccurrach:
Running the installed version of v1.0.2 I get the following output:
```
C:\Users\timmc\AppData\Local\Mu\Python\python36.zip
C:\Users\timmc\AppData\Local\Mu\Python
C:\Users\timmc\AppData\Local\Mu\pkgs
True
C:\Users\timmc\AppData\Roaming\Python\Python36\site-packages
False
```
When running in debugger mode:
```
c:\users\timmc\mu_code   # Note: This is the location of the file, and won't necessarily be mu_code
C:\Users\timmc\AppData\Local\Mu\Python
C:\Users\timmc\AppData\Local\Mu\pkgs
C:\Users\timmc\AppData\Local\Mu\pkgs\IPython\extensions
c:\users\timmc\appdata\local\mu\python\python36.zip
True
C:\Users\timmc\AppData\Roaming\Python\Python36\site-packages
False
```
Last part of log file attached:
[logs.txt](https://github.com/mu-editor/mu/files/2705575/logs.txt)

tjguk:
Well that was an interesting exploration into the somewhat messiness ofo the site/isolation flags. I was particularly confused by the fact that the virtual environment I was working in for development turned off 
user site support, while the NSIS-packaged Python didn't. While I had expected the opposite.

In case it helps anyone, I've been using this code to try out different environments:

https://gist.github.com/tjguk/babef6799c87b3a5e2c240ca8add9c45

The business with the "USERPROFILE" is because I use a different user for installed as opposed to development Mu.

Anyhow, the long and short seems to be that adding a simple os.makedirs(site.getusersitepackages(), exist_ok=True) is likely to set things right. I'll experiment a bit more but I'll probably just mod the PR with that.

ntoll:
@tjguk @tim-mccurrach Aha... this feedback is invaluable, and I think I can see what the problem is.

@tim-mccurrach the reason it's not working for you is because the `mu.pth` file isn't created so the runner and debugger child processes don't pick it up and add the paths that are supposed to be found therein into `sys.path` for the child process. That's obvious from your `log.txt` file (see the bit which says, `[Errno 2] No such file or directory: 'C:\\Users\\timmc\\AppData\\Roaming\\Python\\Python36\\site-packages\\mu.pth'`). This also chimes with @tjguk's comments too -- the USER_SITE directory doesn't exist so where's the filesystem error? (It's in the logs).

So here's where I've slipped up. I've used the `site.USER_SITE` constant to get the path I need. However, what I should be doing is calling the `site.getusersitepackages` function (see: https://docs.python.org/3/library/site.html#site.getusersitepackages) which will both return the correct path and, as the documentation tells us, will also create the directory if it doesn't already exist. The reason the `mu.pth` file isn't written on your machines is probably (my hunch) because the directory path hasn't been created. Why does it work on my machine? I think it's because I was experimenting with the `site` module in the REPL and must have called the method, thus creating the directory, before settling on the solution I pushed as this PR.

Ergo, the simplest solution is to replace assignments from the `site.USER_SITE` to the return value of `site.getusersitepackages`.

I'll push that change later today. Please re-test. :-)

ntoll:
@tjguk @tim-mccurrach OK... thanks to Tim for the heads up on the actual behaviour needed. I've just pushed a small change to fix this (and updated the tests).

Please test!

I'm off to play Christmas carols with the local brass band and will catch up later today... :trumpet: :musical_note: :christmas_tree: :+1:

ntoll:
@tjguk @tim-mccurrach could you please confirm things work with this PR? It's one of the final blockers for a 1.0.2 release. Thanks.

tjguk:
Sorry; setting time aside this evening

ntoll:
Thank you!

tjguk:
Ok; I can confirm that, on an installed version of Mu, I was able to import from another module within the mu_code directory.

tjguk:
... and that, if I load a file from another directory (ie not mu_code), then I am able to import files from *that* directory as well as from the mu_code directory.

ntoll:
Thanks for the heads up. Merging for 1.0.2 release.

PR #728, opened 2019-01-01
Restore test randomisation

knowledgejunkie:
Since pytest-random-order 1.0.0, test randomisation is disabled by
default.

Adding `--random-order` to pytest should restore the desired behaviour.


ntoll:
Thanks for the spot! :-)

PR #739, opened 2019-01-14
File dialog will remember path last used.

### LINKED ISSUE #480: mu_code path issues
Currently, when you click load, and want to load a few files from a path other than `mu_code`, you have to browse to the file location each time. It would be nice if Mu remembered the last path you browsed to, rather than defaulting to `mu_code`. Alternatively, if you could select multiple files to open together, that would be handy.

Similarly, it seems the path `~/mu_code/` is hard-coded as the pwd, so if your file is located in `~/Projects/project_bob/src/`, the following code:

```python
with open('file.txt') as f:
    print(f.readline())
```

will attempt to open `~/mu_code/file.txt`. But if you run the script from the terminal from within the `src` directory it would use `~/Projects/project_bob/src/file.txt`. I think the pwd should be the location of the file run, not always `mu_code`.

ntoll:
This PR fixes #480. @bennuttall please test this behaviour works as expected for you. I'll merge once you reply. This is the final blocker for 1.0.2.

Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
 
     # To maximise compatibility pick earliest image, OS X 10.10 Yosemite
     - os: osx
-      osx_image: xcode8
+      osx_image: xcode8.3
       sudo: required
       language: generic
       python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
 
     # To maximise compatibility pick earliest image, OS X 10.10 Yosemite
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode8
       sudo: required
       language: generic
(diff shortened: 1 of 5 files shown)


bennuttall:
Awesome. Works for me on my PC :+1:

I tried on a Pi and for some reason it changed the `mu-editor` executable to launch with Python 2.7. I'll look into it. (Certainly not related to this PR)

tjguk:
@bennuttall it's a cunning ploy by the Society for the Preservation of Python 2.7 (SPPy2.7). They're everywhere...

ntoll:
Cool... once CI is green I'll merge later tonight. Tomorrow I'll be testing on Windows, OSX and Linux (including Pi) then releasing.

Thanks for the input.

Issue #725, opened 2018-12-14
Text selecting with UTF-8 characters - problem.


abixadamj:
Mu-Editor 1.0, Linux Mint 19 64-bit.
When selecting text, seems to be OK...

![obraz](https://user-images.githubusercontent.com/5088643/50017111-820af400-ffcb-11e8-8947-0507df367f0c.png)

but when there is some character with UTF-8 (Polish special signs).. there is something wired..

![obraz](https://user-images.githubusercontent.com/5088643/50017194-ba123700-ffcb-11e8-9d4d-e0d2569dbce9.png)

It seems that if there is more special signs, then select is shifted...

![obraz](https://user-images.githubusercontent.com/5088643/50017288-efb72000-ffcb-11e8-96d8-dcc8e46a1246.png)

tjguk:
Thanks for the description. For reproducibility would you mind posting a small example which fails in the way you describe, please? Either link to a gist or (if GH allows it) attach a code file here. Thanks

abixadamj:
```
from PyTechBrain import *
from time import sleep

'''
Polish letter in UTF-8: ą
'''

q = PyTechBrain()
```

tjguk:
Perfect; thanks. I'll dig into it

ntoll:
If you're quick... this'll make it into 1.0.2. ;-)

tjguk:
How quick do I have to be? (I don't think it's worth holding anything back for).

ntoll:
By the end of the week? My plan to cut a release today has been shot to pieces by other factors and so I'm just expanding the scope of what to include and hope to prepare for a Friday release on Thursday.

tjguk:
Ok; I'll see how that works out. But this is a minor malfunction of a little-used feature so...

ntoll:
Ack. ;-)

tjguk:
(laughs) https://github.com/mu-editor/mu/blob/master/mu/interface/editor.py#L348

ntoll:
Hah! :-)

tmontes:
Chiming in with my analysis:

* Culprit seems to be in `range_from_positions` called from `highlight_selected_matches` in https://github.com/mu-editor/mu/blob/master/mu/interface/editor.py where a `FIXME` note by @ntoll says:

  > Not clear whether the Scintilla conversions are expecting bytes or characters (ie codepoints).

[...]

PR #741, opened 2019-01-15
Fix selection match highlights with non-ASCII editor contents.

tmontes:
Addresses issue #725, please see my [notes there](
https://github.com/mu-editor/mu/issues/725#issuecomment-454389761).

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -29,7 +29,7 @@ from mu.logic import NEWLINE
 
 
 # Regular Expression for valid individual code 'words'
-RE_VALID_WORD = re.compile('^[A-Za-z0-9_-]*$')
+RE_VALID_WORD = re.compile(r'^\w+$')
 
 
 logger = logging.getLogger(__name__)
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -362,8 +362,7 @@ class EditorPane(QsciScintilla):
         return the corresponding Scintilla line-offset pairs which are
         used for searches, indicators etc.
 
-        FIXME: Not clear whether the Scintilla conversions are expecting
-        bytes or characters (ie codepoints)
+        NOTE: Arguments must be byte offsets into the underlying text bytes.
         """
         start_line, start_offset = self.lineIndexFromPosition(start_position)
         end_line, end_offset = self.lineIndexFromPosition(end_position)
@@ -429,8 +428,10 @@ class EditorPane(QsciScintilla):
         # to the current theme.
         #
         indicators = self.search_indicators['selection']
-        text = self.text()
-        for match in re.finditer(selected_text, text):
+        encoding = 'utf8' if self.isUtf8() else 'latin1'
+        text_bytes = self.text().encode(encoding)
+        selected_text_bytes = selected_text.encode(encoding)
+        for match in re.finditer(selected_text_bytes, text_bytes):
             range = self.range_from_positions(*match.span())
             #
             # Don't highlight the text we've selected
--- a/tests/interface/test_editor.py
+++ b/tests/interface/test_editor.py
@@ -452,6 +452,42 @@ def test_EditorPane_highlight_selected_matches_with_match():
     assert ep.search_indicators['selection']['positions'] == expected_ranges
 
 
+def test_EditorPane_highlight_selected_matches_with_match_in_nonASCII_text():
+    """
+    Ensure that if the current selection is a single word then it causes the
+    expected search/highlight call when the text is not ASCII only.
+
+    There appears to be no way to iterate over indicators within the editor.
+    So we're using the search_indicators structure as a proxy
+    """
+    text = "résumé foo bar foo baz foo"
+    search_for = "foo"
+
+    ep = mu.interface.editor.EditorPane(None, 'baz')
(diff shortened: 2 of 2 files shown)


tmontes:
Progress:
* Included one related fix: highlighting matched, non-ASCII selections as well.
  * Simplified the `RE_VALID_WORD` regular expression to `^\w+$`.
  * Per [Python re's module docs](https://docs.python.org/3/library/re.html), the `\w` class *"includes most characters that can be part of a word in any language, as well as numbers and the underscore"*, which suits the use case perfectly, AFAICT.
  * For some reason, the previous regular expression allowed for empty strings: am I missing something?
* Implemented two tests:
  * Test that selection match highlights work for ASCII only words in a document with non-ASCII characters (as reported in #725).
  * Test that the selection match highlights work for non-ASCII only words.
  * Both passing on my dev system.

Let's see what Travis and AppVeyor have to say...

ntoll:
I know how to fix the CI fails (basically, rebase from current master).

I've taken a look with eyeball Mk.1 and this looks good. I'll try to take a proper look by the end of the weekend, and hopefully merge.

Thank you for the contribution @tmontes..!

tjguk:
If I can I'll try to look at this this afternoon (since it's my code which is failing :) ). But if you're happy to go ahead and merge please do so without waiting for me. I can always pick up issues afterwards if there are any.

tmontes:
@ntoll, @tjguk, thanks for taking a look at this. :)

I "blindly" went for the "update branch with current master" but AppVeyor and Travis are still failing. I then went digging deeper:

* Travis failure analysis:
  * `pycodestyle` is complaining about lines [142](https://github.com/mu-editor/mu/blob/3eaa305ba62e5f4c51057eb5fb84270bdee033f2/mu/debugger/runner.py#L142) and [144](https://github.com/mu-editor/mu/blob/3eaa305ba62e5f4c51057eb5fb84270bdee033f2/mu/debugger/runner.py#L144) which it says are over-indented (they are, indeed, and have been so since mid 2018, commit af47bc1b).
  * It didn't fail before on CI neither on my local check though. Motive: `pycodestyle` has recently been updated and Travis brought in the newer version, 2.5.0, just released on 2019-01-29, that now includes a new check for "E117: Over-indented code blocks" ([change log here](https://pycodestyle.readthedocs.io/en/latest/developer.html#changes)).

* AppVeyor failure analysis:
  * Seems like it fails complaining that `tests\debugger\test_client.py:579: use ==/!= to compare str, bytes, and int literals` [here](https://ci.appveyor.com/project/carlosperate/mu/builds/22074452#L394), triggered by the `make win64` which will run the equivalent of a `make check` as a dependency (am I reading this right?). The log omits further details that I could use to pinpoint why this hasn't failed before.
  * Interestingly, [line 579](https://github.com/mu-editor/mu/blob/3eaa305ba62e5f4c51057eb5fb84270bdee033f2/tests/debugger/test_client.py#L579) in `tests/debugger/test_client.py` does an assertion / comparison which I find odd. The code reads `assert bp.ignore is 5` and the message makes sense. Why is it not using `==`?

Way forward:
* Fixing Travis should be handled outside of this PR, with a single-purpose issue+PR (PR only?) that fixes the over-indented code blocks in `debugger/runner.py`.
* With regards to AppVeyor, my guess is, again, something outside of this PR's scope needs work. Even though this could probably be addressed by changing the `is` operator to an `==` one, I'm not sure i) why the `is` is being used, and, ii), more importantly, why this hasn't failed before. (or maybe I'm reading the AppVeyor log wrongly!).

Summary:
* Both these CI failures seem unrelated to the PR itself.
* Let's hold on until both are fixed at the "project level", I say...

Hopefully someone can chime in with more feedback on this. Thanks!

tmontes:
PS: @ntoll, I see you fixed the indentation the newer 2.5.0 `pycodestyle` is complaining about in PR #749, commit 54c50d55a865b250a3b9845e80ff292b63f2e055. Maybe that's why you suggested rebasing from master?

carlosperate:
>I thought the RE_VALID_WORD might have been mine as well, but I simply inherited and didn't look too closely. I think the original was @carlosperate -- so he might have some notion of why zero-length strings were allowed.

It's been way too long since I wrote that, so honestly cannot remember if there was a special reason why zero-length strings should be allowed, but it doesn't sound like there is, so the new regex looks good.

tmontes:
Regarding the AppVeyor failure I commented on, from the just released Python 3.8.0a1 change log:

> [bpo-34850](
https://bugs.python.org/issue34850): The compiler now produces a [`SyntaxWarning`](
https://docs.python.org/3.8/library/exceptions.html#SyntaxWarning) when identity checks (`is` and `is not`) are used with certain types of literals (e.g. strings, ints). These can often work by accident in CPython, but are not guaranteed by the language spec. The warning advises users to use equality tests (`==` and `!=`) instead.

So, yes, maybe the [`bp.ignore is 5` assertion in `tests/debugger/test_client.py`:579](
https://github.com/mu-editor/mu/blob/3eaa305ba62e5f4c51057eb5fb84270bdee033f2/tests/debugger/test_client.py#L579) should be changed.

Thoughts?

ntoll:
Merged! Thank you for fixing this folks!

PR #748, opened 2019-01-26
Added support for SparkFun SAMD21 Dev and Mini Breakout boards

ShawnHymel:
...and updated test_client.py to check for equality on bp.ignore.

### DIFF
--- a/mu/modes/adafruit.py
+++ b/mu/modes/adafruit.py
@@ -64,6 +64,10 @@ class AdafruitMode(MicroPythonMode):
         (0x239A, 0x803C),  # future board
         (0x239A, 0x803E),  # future board
         (0x239A, None),    # Any Adafruit Boards
+        
+        # Non-Adafruit boards
+        (0x1B4F, 0x8D22),  # SparkFun SAMD21 Mini Breakout
+        (0x1B4F, 0x8D23),  # SparkFun SAMD21 Dev Breakout
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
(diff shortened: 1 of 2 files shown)


mu/modes/adafruit.py
```diff
@@ -64,6 +64,10 @@ class AdafruitMode(MicroPythonMode):
         (0x239A, 0x803C),  # future board
         (0x239A, 0x803E),  # future board
         (0x239A, None),    # Any Adafruit Boards
+
+        # Non-Adafruit boards
+        (0x1B4F, 0x8D22),  # SparkFun SAMD21 Mini Breakout
+        (0x1B4F, 0x8D23),  # SparkFun SAMD21 Dev Breakout
```
carlosperate: Is this Vendor ID specific to Sparkfun boards?
(i.e. although probably not applicable for this case, some boards use the same type of UART-to-USB bridges, so they could end up with the same VID/PID).

ShawnHymel: Yes, the VID is unique to SparkFun. I had @jimblom from SparkFun pull the PIDs for the boards specifically for CircuitPython.

carlosperate: Great, thanks @ShawnHymel.

ntoll:
@ShawnHymel all merged. Thank you very much for your contribution..! BTW, since there are several non-Adafruit boards for CircuitPython, we just changed the "Adafruit Mode" to "CircuitPython Mode" so everyone is equally represented. :-)

ShawnHymel:
Good to know, thanks!

Issue #714, opened 2018-11-29
Feature - where is plotter?


rhubarbdog:
The screen shot at codewith.mu shows a copy of mu with a plotter button.
Is there a version with a plotter?
Is there any documetation and examples of graphing data?

wayoda:
There is a [Tutorial for the Plotter](https://codewith.mu/en/tutorials/1.0/plotter) on the[Code with Mu Tutorials](https://codewith.mu/en/tutorials/) page.

wayoda:
Sorry, I never installed it on an RPi. Can't help on this one.

ukBaz:
There are issues with the plotter on an RPi. See:
https://github.com/mu-editor/mu/issues/689#issuecomment-433419161

carlosperate:
We are going through the process to create PyQtChart (the library that provides the plotted) deb packages, which can then be submitted for inclusion in Debian and/or Raspbian.

Issue  https://github.com/mu-editor/mu/issues/441 is the current thread discussing this work, so I'll close this one as a duplicate. Thanks @wayoda  and @ukBaz for all your help as well.

Issue #694, opened 2018-10-30
Mu on debian pixel Raspberry Pi OS


DavidWhaleMEF:
I tried to install Mu on the Raspberry Pi OS Debian Pixel
![screen shot 2018-10-30 at 10 45 40](https://user-images.githubusercontent.com/25079586/47713024-2958e500-dc31-11e8-88ae-cd875313c2ff.png)
 (running in a virtual box on Mac OS 10.10.5) and it seems to (incorrectly) install the older version of Mu...

![screen shot 2018-10-30 at 10 41 36](https://user-images.githubusercontent.com/25079586/47712798-9b7cfa00-dc30-11e8-9733-a164475150b2.png)

![screen shot 2018-10-30 at 10 43 45](https://user-images.githubusercontent.com/25079586/47712904-da12b480-dc30-11e8-8a04-d7e36e8229f5.png)

carlosperate:
Hi David,

Is it possible you are running Raspbian Jessie? Try running `cat /etc/os-release`.
If you are, it's possible Mu v1.0 is only available in Raspbian Stretch, can you upgrade?

I am not really sure if the Raspbian release is the problem, although it's quite possible since Jessie is running Python 3.4 and old PyQt5 releases that might not be compatible with current Mu.

carlosperate:
Running on the latest Stretch image should work.
It would also be interesting to see if the plotter is available on x86 Raspbian, as it's not currently available in the ARM release, but upstream debian packages for PyQtChart should be available.

carlosperate:
Maybe it can be installed via apt? There is this package in stretch: https://packages.debian.org/stretch/python3-pyqtgraph

As you have raised in a different issue the current mu wheel might not work, but installing mu-editor via apt should install everything necessary for Mu to work without charts, and manually installing that package should be able to unlock it.

Issue #689, opened 2018-10-26
Plotter capability not available in 1.0.1 on Raspberry Pi (latest Stretch raspbian)


topshed:
After complicated install (see issue #688), I got Mu 1.0.1 installed on a PI, but Plotter button is missing.

2018-10-26 11:48:09,456 - root:112(run) INFO: 

-----------------

Starting Mu 1.0.1
2018-10-26 11:48:09,457 - root:113(run) INFO: uname_result(system='Linux', node='raspberrypi', release='4.14.71-v7+', version='#1145 SMP Fri Sep 21 15:38:35 BST 2018', machine='armv7l', processor='')
[...]

PR #749, opened 2019-01-29
Simple support for third party packages.

### LINKED ISSUE #675: Facilitate installation of 3rd party packages for use with Mu
As a (beginner or teacher) user, I want to be able to install 3rd party packages from PyPI in an easy to use/configure way. Furthermore, the packages I install must be isolated from any other user of Mu on the system.

This issue has been created because of the "rule of three" (3+ users have asked for this sort of feature). A typical request comes via Al Sweigart (@asweigart): https://twitter.com/AlSweigart/status/1051960899982872576 

My initial thought is to add a pane to the admin widget containing a textarea into which users write the names of the packages they want in their environment, much like a `requirements.txt` lite. Upon confirmation of any changes we use `pip` (perhaps like this: https://pip.pypa.io/en/stable/user_guide/#user-installs) to install the packages in a known isolated-but-for-mu path.

I'd like this to be in the Autumn/Winter 1.1 release, I also believe making the simplest possible solution (and then iterating) is the way to go... i.e. this feature is considered "beta" upon release (and could be subject to change upon user feedback).

ntoll:
This PR fixes #675.

It has been tested on Linux (Ubuntu 18.04) but not Windows, OSX or Raspberry Pi. **Please give feedback from testing in these environments**.

# What does it do?

This is the simplest possible third party package manager for a beginners' editor. Until now we have bundled a bunch of "third party" modules with Mu in the official installers, thus bloating Mu's size. Furthermore, our bundled modules go out of date and/or miss some important packages for folks. This new functionality allows users to simply specify packages to be installed from PyPI. It also allows users to remove packages.

It is accessed via the "admin" dialog (accessed via the cog on the bottom right of the UI) and is one of the tabs found therein. The following GIF demonstrates it in action:

mu_modules

The user simply ensures the list of module names (one per line) reflects their requirements. To remove a module from the list, just remove its line. Python's `pip` command (which does the heavy lifting under the hood) will, of course, install dependencies as required. To keep things simple (this is a beginners' editor!) the notion of version number is ignored, and the latest version will be installed by default. Ergo, if a user needs to update a package, they should remove and re-install it.

Currently, there is no attempt to handle conflicts between pre-packaged (in the installer) modules or the same module installed via this method. My hunch is to remove these pre-installed modules in favour of this mechanism. **I would love to hear thoughts on this from folks!**

Packages installed in this way are available in both scripts run or debugged in Python 3 mode, or from within the Python 3 REPL (iPython shell).

# How Does it Work?

There were several ways to skin this cat, and I've gone for the simplest I can think of.

* The new tab for third party packages is added to the admin dialog. If, when the dialog is closed via "OK", there is a change in package state, then a new modal dialog launches to provide visibility of the package syncing process.
* This new package related dialog simply contains a text area into which is piped information from `pip` and other functionality.
* The **recommended way to use pip is by launching it in a subprocess**. The `pip` developers will not guarantee the internal API of the `pip` module. This is why things work the way they do: new packages are added to a list. The first new package is popped from the list and installed via `pip` in a subprocess. When the subprocess completes, if there are any more candidate packages to install, the process is repeated.
* The `--target` flag is used to specify an OS specific writeable directory into which these packages will be installed. This directory is added to the Python path when running or debugging Python 3 scripts or using the Python 3 REPL.
* Sadly, the `pip` command doesn't allow for targetting a "package directory" from which to remove packages. Ergo, Mu uses the list of packages that have been removed in the admin dialog to discover the metadata for them in the directory Mu uses for such packages. Happily, the metadata contains a list of all files related to each package. Mu simply iterates over these, deletes them and deletes any parent directories related to the module to be removed.

That's it! All feedback most welcome!

cc/@asweigart

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -21,6 +21,7 @@ import sys
 import logging
 import csv
 import shutil
+import platform
 from PyQt5.QtCore import QSize, QProcess, QTimer
 from PyQt5.QtWidgets import (QVBoxLayout, QListWidget, QLabel, QListWidgetItem,
                              QDialog, QDialogButtonBox, QPlainTextEdit,
@@ -224,7 +225,11 @@ class AdminDialog(QDialog):
         self.tabs.addTab(self.microbit_widget, _('BBC micro:bit Settings'))
         self.package_widget = PackagesWidget()
         self.package_widget.setup(packages)
-        self.tabs.addTab(self.package_widget, _('Third Party Packages'))
+        if 'armv' not in platform.platform():
+            # Only allow third party package management if NOT running on a
+            # Raspberry Pi. See:
+            # https://github.com/mu-editor/mu/pull/749#issuecomment-459051823
+            self.tabs.addTab(self.package_widget, _('Third Party Packages'))
 
     def settings(self):
         """
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -314,6 +314,7 @@ class PackageDialog(QDialog):
         self.to_add = to_add
         self.module_dir = module_dir
         self.pkg_dirs = {}  # To hold locations of to-be-removed packages.
+        self.process = None
         # Basic layout.
         self.setMinimumSize(600, 400)
         self.setWindowTitle(_('Third Party Package Status'))
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -176,18 +176,31 @@ class PackagesWidget(QWidget):
     def setup(self, packages):
         widget_layout = QVBoxLayout()
         self.setLayout(widget_layout)
-        label = QLabel(_('The packages shown below will be available to '
-                         'import in Python 3 mode. Delete a package from the '
-                         'list to remove its availability.\n\n'
-                         'Each separate package name should be on a new '
-                         'line. Packages are installed from PyPI '
-                         '(see: https://pypi.org/).'))
-        label.setWordWrap(True)
-        widget_layout.addWidget(label)
         self.text_area = QPlainTextEdit()
         self.text_area.setLineWrapMode(QPlainTextEdit.NoWrap)
-        self.text_area.setPlainText(packages)
-        widget_layout.addWidget(self.text_area)
+        if 'armv' in platform.platform():
+            # Only allow third party package management if NOT running on a
+            # Raspberry Pi. See:
+            # https://github.com/mu-editor/mu/pull/749#issuecomment-459051823
+            label = QLabel(_('Third party packages are not supported on the '
+                             'Raspberry Pi. Please use the operating '
+                             "system's package manager instead (refer to the "
+                             "operating system's documentation for how to do "
+                             "this)."))
+            label.setWordWrap(True)
+            widget_layout.addWidget(label)
+            self.text_area.setPlainText('')
+        else:
+            label = QLabel(_('The packages shown below will be available to '
+                             'import in Python 3 mode. Delete a package from '
+                             'the list to remove its availability.\n\n'
+                             'Each separate package name should be on a new '
+                             'line. Packages are installed from PyPI '
+                             '(see: https://pypi.org/).'))
+            label.setWordWrap(True)
+            widget_layout.addWidget(label)
+            self.text_area.setPlainText(packages)
+            widget_layout.addWidget(self.text_area)
 
 
 class AdminDialog(QDialog):
@@ -225,11 +238,7 @@ class AdminDialog(QDialog):
         self.tabs.addTab(self.microbit_widget, _('BBC micro:bit Settings'))
         self.package_widget = PackagesWidget()
         self.package_widget.setup(packages)
-        if 'armv' not in platform.platform():
-            # Only allow third party package management if NOT running on a
-            # Raspberry Pi. See:
-            # https://github.com/mu-editor/mu/pull/749#issuecomment-459051823
-            self.tabs.addTab(self.package_widget, _('Third Party Packages'))
+        self.tabs.addTab(self.package_widget, _('Third Party Packages'))
 
     def settings(self):
         """
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -348,7 +348,7 @@ class PackageDialog(QDialog):
         self.pkg_dirs = {}
         for pkg in self.to_remove:
             for d in dirs:
-                if os.path.basename(d).lower().startswith(pkg + '-'):
+                if os.path.basename(d).lower().startswith(pkg.lower() + '-'):
                     self.pkg_dirs[pkg] = d
         if self.pkg_dirs:
             # If there are packages to remove, schedule removal.
@@ -373,11 +373,26 @@ class PackageDialog(QDialog):
                         logger.error('Unable to remove: {}'.format(to_delete))
                         logger.error(ex)
             shutil.rmtree(dist, ignore_errors=True)
-            shutil.rmtree(os.path.join(self.module_dir, package),
-                          ignore_errors=True)
+            # Some modules don't use the module name for the module directory
+            # (they use a lower case variant thereof). E.g. "Fom" vs. "fom".
+            normal_module = os.path.join(self.module_dir, package)
+            lower_module = os.path.join(self.module_dir, package.lower())
+            shutil.rmtree(normal_module, ignore_errors=True)
+            shutil.rmtree(lower_module, ignore_errors=True)
             self.append_data('Removed {}\n'.format(package))
             QTimer.singleShot(2, self.remove_package)
         else:
+            # Clean any directories not containing files.
+            dirs = [os.path.join(self.module_dir, d)
+                    for d in os.listdir(self.module_dir)]
+            for d in dirs:
+                keep = False
+                for entry in os.walk(d):
+                    if entry[2]:
+                        keep = True
+                if not keep:
+                    shutil.rmtree(d, ignore_errors=True)
+            # Check for end state.
             if not (self.to_add or self.process):
                 self.end_state()
 
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -16,11 +16,16 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+import os
+import sys
 import logging
-from PyQt5.QtCore import QSize
+import csv
+import shutil
+from PyQt5.QtCore import QSize, QProcess, QTimer
 from PyQt5.QtWidgets import (QVBoxLayout, QListWidget, QLabel, QListWidgetItem,
                              QDialog, QDialogButtonBox, QPlainTextEdit,
                              QTabWidget, QWidget, QCheckBox, QLineEdit)
+from PyQt5.QtGui import QTextCursor
 from mu.resources import load_icon
 
 
@@ -161,24 +166,48 @@ class MicrobitSettingsWidget(QWidget):
         widget_layout.addStretch()
 
 
+class PackagesWidget(QWidget):
+    """
+    Used for editing and displaying 3rd party packages installed via pip to be
+    used with Python 3 mode.
+    """
+
+    def setup(self, packages):
+        widget_layout = QVBoxLayout()
+        self.setLayout(widget_layout)
+        label = QLabel(_('The packages shown below will be available to '
+                         'import in Python 3 mode. Delete a package from the '
+                         'list to remove its availability.\n\n'
+                         'Each separate package name should be on a new '
+                         'line.'))
+        label.setWordWrap(True)
+        widget_layout.addWidget(label)
+        self.text_area = QPlainTextEdit()
+        self.text_area.setLineWrapMode(QPlainTextEdit.NoWrap)
+        self.text_area.setPlainText(packages)
+        widget_layout.addWidget(self.text_area)
+
+
 class AdminDialog(QDialog):
     """
     Displays administrative related information and settings (logs, environment
-    variables etc...).
+    variables, third party packages etc...).
     """
 
     def __init__(self, parent=None):
         super().__init__(parent)
 
-    def setup(self, log, settings):
+    def setup(self, log, settings, packages):
         self.setMinimumSize(600, 400)
         self.setWindowTitle(_('Mu Administration'))
         widget_layout = QVBoxLayout()
         self.setLayout(widget_layout)
         self.tabs = QTabWidget()
         widget_layout.addWidget(self.tabs)
-        button_box = QDialogButtonBox(QDialogButtonBox.Ok)
+        button_box = QDialogButtonBox(QDialogButtonBox.Ok |
+                                      QDialogButtonBox.Cancel)
         button_box.accepted.connect(self.accept)
+        button_box.rejected.connect(self.reject)
         widget_layout.addWidget(button_box)
         # Tabs
         self.log_widget = LogWidget()
@@ -192,6 +221,9 @@ class AdminDialog(QDialog):
         self.microbit_widget.setup(settings.get('minify', False),
                                    settings.get('microbit_runtime', ''))
         self.tabs.addTab(self.microbit_widget, _('BBC micro:bit Settings'))
+        self.package_widget = PackagesWidget()
+        self.package_widget.setup(packages)
+        self.tabs.addTab(self.package_widget, _('Third Party Packages'))
 
     def settings(self):
         """
@@ -203,6 +235,7 @@ class AdminDialog(QDialog):
             'envars': self.envar_widget.text_area.toPlainText(),
             'minify': self.microbit_widget.minify.isChecked(),
             'microbit_runtime': self.microbit_widget.runtime_path.text(),
+            'packages': self.package_widget.text_area.toPlainText(),
         }
 
 
@@ -262,3 +295,138 @@ class FindReplaceDialog(QDialog):
         Return the value of the global replace flag.
         """
         return self.replace_all_flag.isChecked()
+
+
+class PackageDialog(QDialog):
+    """
+    Display a dialog to indicate the status of the packaging related changes
+    currently run by pip.
+    """
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+    def setup(self, to_remove, to_add, module_dir):
+        """
+        Create the UI for the dialog.
+        """
+        self.to_remove = to_remove
+        self.to_add = to_add
+        self.module_dir = module_dir
+        self.pkg_dirs = {}  # To hold locations of to-be-removed packages.
+        # Basic layout.
+        self.setMinimumSize(600, 400)
+        self.setWindowTitle(_('Third Party Package Status'))
+        widget_layout = QVBoxLayout()
+        self.setLayout(widget_layout)
+        # Text area for pip output.
+        self.text_area = QPlainTextEdit()
+        self.text_area.setReadOnly(True)
+        self.text_area.setLineWrapMode(QPlainTextEdit.NoWrap)
+        widget_layout.addWidget(self.text_area)
+        # Buttons.
+        self.button_box = QDialogButtonBox(QDialogButtonBox.Ok)
+        self.button_box.button(QDialogButtonBox.Ok).setEnabled(False)
+        self.button_box.accepted.connect(self.accept)
+        widget_layout.addWidget(self.button_box)
+        # Kick off processing of packages.
+        if self.to_remove:
+            self.remove_packages()
+        if self.to_add:
+            self.run_pip()
+
+    def remove_packages(self):
+        """
+        Work out which packages need to be removed and then kick off their
+        removal.
+        """
+        dirs = [os.path.join(self.module_dir, d)
+                for d in os.listdir(self.module_dir)
+                if d.endswith("dist-info")]
+        self.pkg_dirs = {}
+        for pkg in self.to_remove:
+            for d in dirs:
+                if os.path.basename(d).lower().startswith(pkg + '-'):
+                    self.pkg_dirs[pkg] = d
+        if self.pkg_dirs:
+            # If there are packages to remove, schedule removal.
+            QTimer.singleShot(2, self.remove_package)
+
+    def remove_package(self):
+        """
+        Take a package from the pending packages to be removed, delete all its
+        assets and schedule the removal of the remaining packages. If there are
+        no packages to remove, move to the finished state.
+        """
+        if self.pkg_dirs:
+            package, dist = self.pkg_dirs.popitem()
+            record = os.path.join(dist, 'RECORD')
+            with open(record) as f:
+                files = csv.reader(f)
+                for row in files:
+                    to_delete = os.path.join(self.module_dir, row[0])
+                    try:
+                        os.remove(to_delete)
+                    except Exception as ex:
+                        logger.error('Unable to remove: {}'.format(to_delete))
+                        logger.error(ex)
+            shutil.rmtree(dist, ignore_errors=True)
+            shutil.rmtree(os.path.join(self.module_dir, package),
+                          ignore_errors=True)
+            self.append_data('Removed {}\n'.format(package))
+            QTimer.singleShot(2, self.remove_package)
+        else:
+            if not (self.to_add or self.process):
+                self.end_state()
+
+    def end_state(self):
+        """
+        Set the UI to a valid end state.
+        """
+        self.append_data('\nFINISHED')
+        self.button_box.button(QDialogButtonBox.Ok).setEnabled(True)
+
+    def run_pip(self):
+        """
+        Run a pip command in a subprocess and pipe the output to the dialog's
+        text area.
+        """
+        package = self.to_add.pop()
+        args = ['-m', 'pip', 'install', package, '--target',
+                self.module_dir]
+        self.process = QProcess(self)
+        self.process.readyRead.connect(self.read_process)
+        self.process.finished.connect(self.finished)
+        self.process.start(sys.executable, args)
+
+    def finished(self):
+        """
+        Called when the subprocess that uses pip to install a package is
+        finished.
+        """
+        if self.to_add:
+            self.process = None
+            self.run_pip()
+        else:
+            if not self.pkg_dirs:
+                self.end_state()
+
+    def read_process(self):
+        """
+        Read data from the child process and append it to the text area. Try
+        to keep reading until there's no more data from the process.
+        """
+        data = self.process.readAll()
+        if data:
+            self.append_data(data.data().decode('utf-8'))
+            QTimer.singleShot(2, self.read_process)
+
+    def append_data(self, msg):
+        """
+        Add data to the end of the text area.
+        """
+        cursor = self.text_area.textCursor()
+        cursor.movePosition(QTextCursor.End)
+        cursor.insertText(msg)
+        cursor.movePosition(QTextCursor.End)
+        self.text_area.setTextCursor(cursor)
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -179,7 +179,8 @@ class PackagesWidget(QWidget):
                          'import in Python 3 mode. Delete a package from the '
                          'list to remove its availability.\n\n'
                          'Each separate package name should be on a new '
-                         'line.'))
+                         'line. Packages are installed from PyPI '
+                         '(see: https://pypi.org/).'))
         label.setWordWrap(True)
         widget_layout.addWidget(label)
         self.text_area = QPlainTextEdit()
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -348,7 +348,9 @@ class PackageDialog(QDialog):
         self.pkg_dirs = {}
         for pkg in self.to_remove:
             for d in dirs:
-                if os.path.basename(d).lower().startswith(pkg.lower() + '-'):
+                # Assets on the filesystem use a normalised package name.
+                pkg_name = pkg.replace('-', '_').lower()
+                if os.path.basename(d).lower().startswith(pkg_name + '-'):
                     self.pkg_dirs[pkg] = d
         if self.pkg_dirs:
             # If there are packages to remove, schedule removal.
@@ -412,6 +414,7 @@ class PackageDialog(QDialog):
         args = ['-m', 'pip', 'install', package, '--target',
                 self.module_dir]
         self.process = QProcess(self)
+        self.process.setProcessChannelMode(QProcess.MergedChannels)
         self.process.readyRead.connect(self.read_process)
         self.process.finished.connect(self.finished)
         self.process.start(sys.executable, args)
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -156,10 +156,10 @@ def installed_packages():
                 if d.endswith("dist-info")]
     for pkg in pkg_dirs:
         metadata_file = os.path.join(pkg, 'METADATA')
-        with open(metadata_file) as f:
+        with open(metadata_file, 'rb') as f:
             lines = f.readlines()
-            name = lines[1].rsplit(':')[-1].strip()
-            result.append(name)
+            name = lines[1].rsplit(b':')[-1].strip()
+            result.append(name.decode('utf-8'))
     return sorted(result)
 
 
@@ -1108,25 +1108,25 @@ class Editor:
         with open(LOG_FILE, 'r', encoding='utf8') as logfile:
             new_settings = self._view.show_admin(logfile.read(), settings,
                                                  '\n'.join(packages))
-            if new_settings:
-                self.envars = extract_envars(new_settings['envars'])
-                self.minify = new_settings['minify']
-                runtime = new_settings['microbit_runtime'].strip()
-                if runtime and not os.path.isfile(runtime):
-                    self.microbit_runtime = ''
-                    message = _('Could not find MicroPython runtime.')
-                    information = _("The micro:bit runtime you specified "
-                                    "('{}') does not exist. "
-                                    "Please try again.").format(runtime)
-                    self._view.show_message(message, information)
-                else:
-                    self.microbit_runtime = runtime
-                new_packages = [p for p in
-                                new_settings['packages'].lower().split('\n')
-                                if p.strip()]
-                self.sync_package_state(packages, new_packages)
+        if new_settings:
+            self.envars = extract_envars(new_settings['envars'])
+            self.minify = new_settings['minify']
+            runtime = new_settings['microbit_runtime'].strip()
+            if runtime and not os.path.isfile(runtime):
+                self.microbit_runtime = ''
+                message = _('Could not find MicroPython runtime.')
+                information = _("The micro:bit runtime you specified "
+                                "('{}') does not exist. "
+                                "Please try again.").format(runtime)
+                self._view.show_message(message, information)
             else:
-                logger.info("No admin settings changed.")
+                self.microbit_runtime = runtime
+            new_packages = [p for p in
+                            new_settings['packages'].lower().split('\n')
+                            if p.strip()]
+            self.sync_package_state(packages, new_packages)
+        else:
+            logger.info("No admin settings changed.")
 
     def sync_package_state(self, old_packages, new_packages):
         """
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1124,7 +1124,8 @@ class Editor:
             new_packages = [p for p in
                             new_settings['packages'].lower().split('\n')
                             if p.strip()]
-            self.sync_package_state(packages, new_packages)
+            old_packages = [p.lower() for p in packages]
+            self.sync_package_state(old_packages, new_packages)
         else:
             logger.info("No admin settings changed.")
 
@@ -1137,9 +1138,15 @@ class Editor:
         """
         old = set(old_packages)
         new = set(new_packages)
+        logger.info('Synchronize package states...')
+        logger.info('Old: {}'.format(old))
+        logger.info('New: {}'.format(new))
         to_remove = old.difference(new)
         to_add = new.difference(old)
         if to_remove or to_add:
+            logger.info('To add: {}'.format(to_add))
+            logger.info('To remove: {}'.format(to_remove))
+            logger.info('Site packages: {}'.format(MODULE_DIR))
             self._view.sync_packages(to_remove, to_add, MODULE_DIR)
 
     def select_mode(self, event=None):
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -47,6 +47,9 @@ HOME_DIRECTORY = os.path.expanduser('~')
 WORKSPACE_NAME = 'mu_code'
 # The default directory for application data (i.e., configuration).
 DATA_DIR = appdirs.user_data_dir(appname='mu', appauthor='python')
+# The directory containing user installed third party modules.
+MODULE_DIR = os.path.join(DATA_DIR, 'site-packages')
+sys.path.append(MODULE_DIR)
 # The default directory for application logs.
 LOG_DIR = appdirs.user_log_dir(appname='mu', appauthor='python')
 # The path to the log file for the application.
@@ -142,6 +145,22 @@ ENCODING_COOKIE_RE = re.compile(
     "^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)")
 
 logger = logging.getLogger(__name__)
+
+
+def installed_packages():
+    """
+    List all the third party modules installed by the user.
+    """
+    result = []
+    pkg_dirs = [os.path.join(MODULE_DIR, d) for d in os.listdir(MODULE_DIR)
+                if d.endswith("dist-info")]
+    for pkg in pkg_dirs:
+        metadata_file = os.path.join(pkg, 'METADATA')
+        with open(metadata_file) as f:
+            lines = f.readlines()
+            name = lines[1].rsplit(':')[-1].strip()
+            result.append(name)
+    return sorted(result)
 
 
 def write_and_flush(fileobj, content):
@@ -559,6 +578,9 @@ class Editor:
         if not os.path.exists(DATA_DIR):
             logger.debug('Creating directory: {}'.format(DATA_DIR))
             os.makedirs(DATA_DIR)
+        if not os.path.exists(MODULE_DIR):
+            logger.debug('Creating directory: {}'.format(MODULE_DIR))
+            os.makedirs(MODULE_DIR)
         logger.info('Settings path: {}'.format(get_settings_path()))
         logger.info('Session path: {}'.format(get_session_path()))
         logger.info('Log directory: {}'.format(LOG_DIR))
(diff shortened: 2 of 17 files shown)


bennuttall:
Just testing this on a Pi. It's a great idea and will be really helpful.

UI/UX: :+1: 

It seems to work nicely. I just added a couple of libraries and from the output, it seemed to do what I expected it to. It said it has succeeded.

However I tried to import a newly installed module and it failed (it seems the library was never installed). I went to try again and my list had disappeared. I'll investigate to see what went on, but that's all I know so far.

A few questions:

1. Where does it try to install new modules?
1. How do I upgrade a module I installed in Mu?
1. If pressing "OK" on the list of packages installs them, does it also upgrade new ones? (what if I want to / what if I don't want to?)
1. I assume these installations are at "user" level, i.e. local to the pi user, rather than as root with privilege escalation. Correct?

It might be worth pointing out that this is a pip installer, in case people either assume they can install apt packages, or they enter the apt package name "e.g. python3-abc" instead of "abc".

Really excited about this :1st_place_medal:

romilly:
I like the look of this *lots*. I'll have a play this week-end.

ntoll:
Where does it install packages..? Wherever this points to (from `logic.py`): 

`MODULE_DIR = os.path.join(DATA_DIR, 'site-packages')`.

On Linux that's likely to be: `~/.local/share/mu/site-packages/`

How to upgrade? Uninstall / reinstall.

No upgrades automagically (this really is super simple on purpose).

Yes... these are at user level (see path pasted above).

I've been testing it by poking around in the directory referenced above as I try to do different things. If there's an error, it should display it in package dialog's text area. If not, I've probably missed a setting to include `stderr` or something (likely). In any case, the log file may also contain information too. Under the hood it's just `pip` doing this: 

`python -m pip install foo --target /home/username/.local/share/mu/site-packages`

Where `python` is whatever Mu thinks is `sys.python`.

ntoll:
@bennuttall sorry, missed your apt related comment at the end. If they get confused by `apt` package names they're probably advanced enough to work it out for themselves. However, I agree mentioning "PyPI" will help mitigate problems.

ntoll:
@bennuttall OK... I added the following line to the UI instructions: 

`Packages are installed from PyPI (see: https://pypi.org/).`

If that's not enough of a clue (along with link for further information) then they're in a right pickle. This will, of course, be documented and explained in the tutorials and howtos for 1.1.

bennuttall:
If packages installed within Mu are only available within Mu, a user could find themselves confused when they try to launch their scripts from elsewhere, e.g. "I've finished my code, now let's have cron run it at boot".

What's the rationale for not installing into the (user's) library path?

carlosperate:
This looks great, thanks Nicholas!

I don't have much time to look into this PR properly today, I'll give it a proper go tomorrow or the day after, but I though I should give it a tough first test. So the first thing I tried was to install a module that Mu depends on and see what happens, in this case iPython.

Installing works fine, and doesn't seem to break the REPL, but encountered a couple of issues when uninstalling.
1. No matter which package from the list I remove, it always uninstalls and reinstalls Pygments as well (this happens to be the first item in the list, not sure if related or coincidence).
2. Removing all of them still leaves two packages present (ipython-genutils, prompt-toolkit) and I cannot remove them.
    * No errros given on user interface
    * Few errros shown in log:
    ```
    2019-01-30 00:12:47,612 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/pygmentize
    2019-01-30 00:12:47,612 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/pygmentize'
    2019-01-30 00:13:31,661 - mu.logic:1099(show_admin) INFO: Showing admin with logs from /Users/microbit-carlos/Library/Logs/mu/mu.log
    2019-01-30 00:13:36,899 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../share/man/man1/ipython.1.gz
    2019-01-30 00:13:36,899 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../share/man/man1/ipython.1.gz'
    2019-01-30 00:13:36,900 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/iptest
    2019-01-30 00:13:36,900 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/iptest'
    2019-01-30 00:13:36,900 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/iptest3
    2019-01-30 00:13:36,900 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/iptest3'
    2019-01-30 00:13:36,900 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/ipython
    2019-01-30 00:13:36,900 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/ipython'
    2019-01-30 00:13:36,900 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/ipython3
    2019-01-30 00:13:36,900 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/ipython3'
    2019-01-30 00:13:37,112 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/easy_install
    2019-01-30 00:13:37,112 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/easy_install'
    2019-01-30 00:13:37,112 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/easy_install-3.6
    2019-01-30 00:13:37,112 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/easy_install-3.6'
    2019-01-30 00:13:37,212 - mu.interface.dialogs:373(remove_package) ERROR: Unable to remove: /Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/pygmentize
2019-01-30 00:13:37,212 - mu.interface.dialogs:374(remove_package) ERROR: [Errno 2] No such file or directory: '/Users/microbit-carlos/Library/Application Support/mu/site-packages/../../bin/pygmentize'
```

GIF (heavy, 3 MBs) showing the test:

**Comment by tjguk** [comment]
I've had the same result as @bennuttall: the usual pip output suggested that the install had worked. But "import networkzero as nw0" failed with an ImportError. I'll look into it later to see if I can see what's happening.

Win10, Python 3.6.4, pip 18.0, running from the checked-out branch

**Comment by ntoll** [comment]
@bennuttall @tjguk hmm... I'll play around on my Windows laptop and see what's going on.

**Comment by tjguk** [comment]
Well there's definitely something up with sys.path. This code (run under Mu):

```
import os, sys
for p in sys.path:
    print(p)
```

gives this result:

```
c:\users\goldent\mu_code
c:\users\goldent\mu_code\:C:\Users\goldent\.venvs\mu\Scripts\python36.zip:C:\Program Files\Python36\DLLs:C:\Program Files\Python36\lib:C:\Program Files\Python36:C:\Users\goldent\.venvs\mu:C:\Users\goldent\.venvs\mu\lib\site-packages:C:\Users\goldent\AppData\Local\python\mu\site-packages:C:\Users\goldent\.venvs\mu\lib\site-packages\IPython\extensions
C:\Users\goldent\.venvs\mu\Scripts\python36.zip
C:\Program Files\Python36\DLLs
C:\Program Files\Python36\lib
C:\Program Files\Python36
C:\Users\goldent\.venvs\mu
C:\Users\goldent\.venvs\mu\lib\site-packages
```

Note that second line, which is clearly a mis-munged path. (Looks like the colon-separator as wrongly been used to manipulate rather than the semicolon which is correct on Windows)

**Comment by tjguk** [comment]
For completeness, the path where the packages should be and were installed is `C:\Users\goldent\AppData\Local\python\mu\site-packages` which is somewhere inside that jumbled second line (if you squint a bit)

**Comment by ntoll** [comment]
@tjguk bingo. Cross platform inconsistencies FTW. I'll fix just as soon as I finish paying my tax bill! :-)

**Comment by ntoll** [comment]
OK... so I have it working on Windows if I just `os.pathsep` instead of `:` for generating the `PYTHONPATH` from `sys.path`. I've also updated the unit tests to reflect this change. @tjguk can you please re-check..? Thank you.

**Comment by ntoll** [comment]
@carlosperate aha... this is likely due to name mismatches between the advertised package name (found in the METADATA for the package) and the name used for identifying the module on the filesystem. I'll take a look.

**Comment by ntoll** [comment]
@bennuttall but this is *still* beyond the scope of Mu. This is a learning opportunity for folks to understand that they need to install dependencies for their stuff. The simplest solution being to add the Mu `site-packages` to their system Python PATH. :-)

**Comment by tjguk** [comment]
I think the crux here is that the internal pip functionality being added here is trying to solve what we might call a "walled-garden problem" for users of the bundled Mu distro who have the advantage of not having to install pre-installed packages but the disadvantage of not being able to install anything else!

Where a user is using a non-bundled Mu, they can install whatever they like in the usual way and it will be available to code run under Mu. Or am I wrong?

I mean: I do recognise the force of @bennuttall argument: that code which runs under Mu should not, ideally, fail to run outside Mu. I just don't know how much of an issue it would be. 

Thinking aloud about low-tech solutions: it wouldn't be the most complex thing in the world for Mu to run "pip freeze" and generate a requirements.txt file which could be run outside the Mu environment ("pip install --user -r requirements") to meet the need.

**Comment by ntoll** [comment]
@carlosperate my hunch was correct, some of the package names were not getting normalised to the correct form for detecting their assets on the filesystem. E.g the package named `Foo-bar` has metadata contained in a directory called `foo_bar-1.0.0.dist-info`. I've updated the code and changed the relevant unit test to reflect this requirement. Please try again. Thank you.

**Comment by carlosperate** [comment]
> Thinking aloud about low-tech solutions: it wouldn't be the most complex thing in the world for Mu to run "pip freeze" and generate a requirements.txt file which could be run outside the Mu environment ("pip install --user -r requirements") to meet the need.

This sounds like something that could go into the website, rather than a Mu feature. Goes well with the idea of "graduating" and showing a way for users to move their development environment. The problem would be discoverability, as things that are not straightforward to explain (or infer from the app) are usually hard to search for and hard find.

I do agree that the Raspberry Pi use case is a bit more confusing. For Windows and macOS, Mu runs its own packaged Python and I think it is clear to understand that it is its own "Python bubble".
Raspberry Pi on the other hand uses the system python3, so a user "pip3 installing", or "apt-get install python3-x" packages will have them available within Mu, so it is a bit more complex to understand why it works one way and not the other.

I woudn't be against Mu installing packages in the system python *only* on Raspberry Pi, as that makes that environment easier to understand, but that would require dealing with that specific use-case in the code and adds complexity to the codebase, it's not a lot, but these things do add up.

**Comment by ntoll** [comment]
@tjguk yes, you are correct. If the editor was installed in a non-bundled way, whatever is available to system Python (or within your virtualenv) should just work.

Regarding @bennuttall's excellent point, this is *different* to how some of his own colleagues have discussed how Mu should work (isolated and self-contained with sensible defaults -- i.e. the completely understandable "don't make me think" problem for beginner developers who already have enough on their plate to think about). I believe there's enough of a simple solution (just install the packages you need via system Python's pip or via apt-get), that such a problem is easy to fix. The simplest one-time fix is to add Mu's `site-packages` to sys.path in one of several easy-to-do and well documented ways.

Does this make sense? I'm all ears for alternatives.

**Comment by ntoll** [comment]
@carlosperate I think we're agreeing when I say this is a learning opportunity and you mention this is something for the website.

This feels **VERY** much like the "if you're asking about *that* sort of thing, you're probably too advanced for Mu" situation.  In which case we should try to help (docs on the website etc), but, as you put really well, this is an opportunity for "graduation" to the next level. Mu should encourage this sort of thing.

Finally, I'd feel very uncomfortable introducing yet more OS specific checks. We already do it for the plotter (because Raspbian's version of PyQt is from pre-history), and this just makes the code more complicated. Simple code is good code, and this sort of change doesn't sit well with me.

**Comment by tjguk** [comment]
I think the key question is *how* to graduate in the sort of case which @bennuttall is describing. @ntoll is suggesting adding Mu's site-packages to sys.path in some way. This will (presumably) work but has three drawbacks that I see:

* I should be easy to work out where that is
* It won't work if you want to take a working project to a different machine
* It won't include any of Mu's own pre-installed modules (eg pygame etc.)

My suggestion was, in some way, to generate a requirements.txt. This, I think, overcomes the disadvantages I listed above, assuming that we have a simple way of achieving it from the new admin pane. @carlosperate thinks this should be docs rather than functionality. I'm fine with people learning -- in fact I'm proposing that they do indeed learn how a pip reqts file works etc. But the only way I can see for them to generate the file for themselves is to do something like this in Mu:

```
import subprocess
subprocess.run("pip freeze")
```

Obviously, that could be documented, but seems to me a bit clumsy (and it's not something I'd expect anyone to do in other circs) as opposed to us having an option, or always displaying the current list or something.

Thoughts?

**Comment by carlosperate** [comment]
I also agree this is a learning opportunity. My main concern is that these learning opportunities usually end up in frustration.
No user is going to read the website from a to z, and they won't know how to figure out these things (a lot of users will not even know the website exists, they are just double clicking on a pre-installed application on the school or home computer). So what I think will end up happening is that we will have people not knowing why their code doesn't work outside of Mu.

Having this on the website is fine, but I think it is very important for this content to be discoverable in the app. That might end up being a link on the settings window, but that's an important link to have.

---

About the remove pacakges fix, thanks @ntoll!
Tested the latest commit and that removes the packages correctly.

I has playing with it installing `pyocd` and `ipython`, removed all the packages, and the settings window doesn't have any packages shown (great!).

However, looking at the site_packages folder, these files are still present:

image

It's possible this is because to the packages metadata not listing all the files they install, in which case is not really for Mu to fix, but perhaps there is something to look into? In case there is a missing clean-up step that could be implemented.
If a setup.py file adds an entry point for a console_script, that will create those files in the bin directory, is that something Mu can remove as well? I would have thought pip would deal with these kind of files.

**Comment by ntoll** [comment]
@tjguk hehehe... I love this sort of discussion. I **feels** like we're making progress (even if we may not be).

OK... how about we dump the equivalent of pip freeze into the logs..? That should just be "cut and pastable" and in a reasonably easy place to find.

**Comment by ntoll** [comment]
@tjguk ack. :-) I'm going to take the dog for a walk and think on this.

**Comment by holdenweb** [comment]
This may be a naiive suggestion, but when mu contains its own Python interpreter, would it be possible to provide a `mupython` command that ran the same interpreter in command-line mode? This seems to me like the easiest way to allow run-on-boot commands that use the same environment as mu.

**Comment by tjguk** [comment]
@holdenweb it's a fine line, but I think that's moving Mu from being an editor into being an ecosystem, which I don't think is A Good Thing. [*]

Also, it's only really kicking the can down the road a bit: if someone wants to take that code and run it on a machine without Mu installed at all (or with Mu installed with a completely different set of additional packages) then you're *still* going to have to come up with a solution.

[*] Yes, that was a "1066 And All That" reference for those old enough to think it might be :)

**Comment by bennuttall** [comment]
> but this is still beyond the scope of Mu.

I don't think it should be. Perhaps there's not much of a case for it on a traditional PC, but the Pi is a different beast.

Plus, my mentality for "zero" stuff is that I try to provide a smooth learning curve that helps them get started but let them keep going rather than get stuck at the edges of the library. there was a great blog posts about this concept recently: [Escape Hatches and Ejector Seats](https://anvil.works/blog/escape-hatches-and-ejector-seats). Sure, people will eventually move on from Mu, but I would hope they don't hit the edges too quick!

> This is a learning opportunity for folks to understand that they need to install dependencies for their stuff.

"But I already did"

> The simplest solution being to add the Mu site-packages to their system Python PATH. :-)

Not simple :)

> My main concern is that these learning opportunities usually end up in frustration.

Absolutely! Using IDLE is a learning opportunity...

> it's a fine line, but I think that's moving Mu from being an editor into being an ecosystem, which I don't think is A Good Thing.

Indeed. Simple is better than complex.

When you have multiple Python installations and various package dependencies to think about, you isolate projects into virtual environments With Raspberry Pi, the SD card *is* your isolated environment. There's a system python, and you generally install pip packages system-wide with sudo.

> Regarding @bennuttall's excellent point, this is different to how some of his own colleagues have discussed how Mu should work

I think in this context, my colleagues wouldn't want this side-effect.

I understand the rationale for Mu isolating its environment for Windows users, but the Debian/Raspbian version would use system python and should use system pip. On Raspbian, you know users have access to the terminal and pip. If Mu's package manager installed for the pi user, that would be fine, but if it only installs within Mu, it's just going to cause confusion.

Love the idea - don't want to throw a spanner in the works - just pointing out common use case from my experience!

**Comment by ntoll** [comment]
*sigh* 

> Pi is a different beast

OK... in this case (just like with the plotter) perhaps the simplest thing to do is not show this admin option on the Pi. Folks can just `apt-get` their stuff as per standard Debian practices. Then the problem magically disappears.

**Comment by ntoll** [comment]
@bennuttall on that note... what does `sys.info` return on a Pi so we can identify and make the relevant changes to the admin dialog..?

**Comment by ntoll** [comment]
FWIW... when it comes to the plotter we set a flag on the result of importing the chart library from PyQt, so this approach won't work for detecting a Pi.

**Comment by ntoll** [comment]
Apologies... I think I meant something like:

```import platform

p = platform.platform()```

And using the content of `p` to work out if I'm running in Raspbian... IYSWIM

**Comment by bennuttall** [comment]
> OK... in this case (just like with the plotter) perhaps the simplest thing to do is not show this admin option on the Pi. Folks can just apt-get their stuff as per standard Debian practices. Then the problem magically disappears.

I think that's best - it's a great feature but I think more trouble than it's worth for Pi users. Just trying to protect people from getting a weird experience.

> @bennuttall on that note... what does ~sys.info~platform.platform() return on a Pi so we can identify and make the relevant changes to the admin dialog..?

```
Linux-4.14.79-v7+-armv7l-with-debian-9.6
```

Could be any kernel version, debian version or armv6 or armv7, so a match for "armv" should catch it.

> FWIW... when it comes to the plotter we set a flag on the result of importing the chart library from PyQt, so this approach won't work for detecting a Pi.

When Raspbian Buster arrives, we will magically get pyqtchart and the plotter should just work :+1:

**Comment by ntoll** [comment]
@bennuttall just pushed a fix for this. Can you please re-check?

Folks, I'd like to merge this into master so we can release an alpha 1 next week. Any reason NOT to do this? If I don't hear otherwise, I'll merge on Monday. :-)

File: mu/interface/dialogs.py [hunk 1]
```diff
@@ -224,7 +225,11 @@ def setup(self, log, settings, packages):
         self.tabs.addTab(self.microbit_widget, _('BBC micro:bit Settings'))
         self.package_widget = PackagesWidget()
         self.package_widget.setup(packages)
-        self.tabs.addTab(self.package_widget, _('Third Party Packages'))
+        if 'armv' not in platform.platform():
+            # Only allow third party package management if NOT running on a
+            # Raspberry Pi. See:
+            # https://github.com/mu-editor/mu/pull/749#issuecomment-459051823
+            self.tabs.addTab(self.package_widget, _('Third Party Packages'))
```
carlosperate: I think a missing tab in the settings depending on the platform could cause confusion for online material, tutorials or users trying to help other users.
I've often insisted to other people to go a specific setting in an app just to find out that the instructions I was given them were wrong because the feature was missing or somewhere else in their version they had.

Could we instead show the tab but change the content, a simple one liner saying something along of the lines of "In this platform packages can be installed in using pip3."? That way users would not wonder if they have an old version of Mu, and also will point them to the right direction.

ntoll: Right... so we're just going to go round and round in circles about this. I believe we have a good-enough solution right now, and we should get this functionality in front of users as soon as possible as an alpha release so we get *feedback* from real people. Anything else is "educated" speculation (but speculation nonetheless). 

Only with feedback can we make an properly informed decision. If we don't get feedback, we're probably OK as we are. :-)

carlosperate: I would say the Plotter not being available in Raspberry Pi is a good example of missing features in different platforms causing confusion:
https://github.com/mu-editor/mu/issues/714
https://github.com/mu-editor/mu/issues/689
https://github.com/mu-editor/mu/issues/694

tjguk:
I don't feel strongly about this, and I definitely support @ntoll approach of: let's get it out there. But, turning that back around again: why not "get it out there" with the same pip support enabled for RPi as for everyone else. We (via @bennuttall or elsewhere) can get feedback from real users who are or aren't confused and can decide on that basis.

Anyhow, having said my piece on that I'll leave @ntoll to decide.

mu/logic.py
```diff
@@ -1074,28 +1096,51 @@ def show_admin(self, event=None):
...
+                    message = _('Could not find MicroPython runtime.')
+                    information = _("The micro:bit runtime you specified "
+                                    "('{}') does not exist. "
+                                    "Please try again.").format(runtime)
+                    self._view.show_message(message, information)
+                else:
+                    self.microbit_runtime = runtime
+                new_packages = [p for p in
+                                new_settings['packages'].lower().split('\n')
+                                if p.strip()]
```
carlosperate: If only the `new_packages` list is lower cased (and `packages` left with mixed case) you can end up with false positives when the two lists are compared in `self.sync_package_state()`.

mu/logic.py
```diff
@@ -1074,28 +1096,51 @@ def show_admin(self, event=None):
...
-            runtime = new_settings['microbit_runtime'].strip()
-            if runtime and not os.path.isfile(runtime):
-                self.microbit_runtime = ''
-                message = _('Could not find MicroPython runtime.')
-                information = _("The micro:bit runtime you specified ('{}') "
-                                "does not exist. "
-                                "Please try again.").format(runtime)
-                self._view.show_message(message, information)
+            new_settings = self._view.show_admin(logfile.read(), settings,
+                                                 '\n'.join(packages))
```
carlosperate: Don't think this is really a problem, but everything below this line could be unindented out of the `with` statement, so that the log file handler can be closed.
Line 1129 at the end of this function (inside this `with` statement) does also write to the log, but it seems to work fine.

carlosperate:
I was testing the macOS packaged version and it crashed every time I tried to open the settings.

<details>
<summary>Log</summary>
<pre>Starting Mu 1.0.2
2019-02-01 20:51:20,169 - root:112(run) INFO: uname_result(system='Darwin', node='Carlos-MBP.local', release='18.2.0', version='Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64', machine='x86_64', processor='i386')
2019-02-01 20:51:20,169 - root:113(run) INFO: Python path: ['/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/MacOS', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/python/lib/python36.zip', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/python/lib/python3.6', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/python/lib/python3.6/lib-dynload', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/python/lib/python3.6/site-packages', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/app', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/app_packages', '/Users/microbit-carlos/Library/Application Support/mu/site-packages', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/app_packages/IPython/extensions']
2019-02-01 20:51:20,169 - root:114(run) INFO: Language code: en_GB
2019-02-01 20:51:20,383 - mu.logic:562(__init__) INFO: Setting up editor.
2019-02-01 20:51:20,383 - mu.logic:584(__init__) INFO: Settings path: /Users/microbit-carlos/Library/Application Support/mu/settings.json
2019-02-01 20:51:20,383 - mu.logic:585(__init__) INFO: Session path: /Users/microbit-carlos/Library/Application Support/mu/session.json
2019-02-01 20:51:20,383 - mu.logic:586(__init__) INFO: Log directory: /Users/microbit-carlos/Library/Logs/mu
2019-02-01 20:51:20,383 - mu.logic:587(__init__) INFO: Data directory: /Users/microbit-carlos/Library/Application Support/mu
2019-02-01 20:51:20,396 - mu.logic:600(setup) INFO: Available modes: python, adafruit, microbit, debugger, pygamezero
2019-02-01 20:51:20,509 - mu.logic:1195(change_mode) INFO: Workspace directory: /Users/microbit-carlos/mu_code
2019-02-01 20:51:20,515 - mu.logic:649(restore_session) INFO: Restoring session from: /Users/microbit-carlos/Library/Application Support/mu/session.json
2019-02-01 20:51:20,515 - mu.logic:650(restore_session) DEBUG: {'theme': 'day', 'mode': 'python', 'paths': ['/Users/microbit-carlos/mu_code/packages_test.py'], 'envars': [], 'minify': False, 'microbit_runtime': '', 'zoom_level': 2}
2019-02-01 20:51:20,515 - mu.logic:737(_load) INFO: Loading script from: /Users/microbit-carlos/mu_code/packages_test.py
2019-02-01 20:51:20,515 - mu.logic:293(read_and_decode) DEBUG: Trying to decode with utf-8
2019-02-01 20:51:20,515 - mu.logic:296(read_and_decode) INFO: Decoded with utf-8
2019-02-01 20:51:20,516 - mu.logic:309(read_and_decode) DEBUG: Detected newline '\n'
2019-02-01 20:51:20,516 - mu.logic:811(_load) DEBUG: # Write your code here :-)
print('hello world')
2019-02-01 20:51:20,578 - mu.logic:671(restore_session) INFO: Loaded files.
2019-02-01 20:51:20,578 - mu.logic:675(restore_session) INFO: User defined environment variables: []
2019-02-01 20:51:20,578 - mu.logic:679(restore_session) INFO: Minify scripts on micro:bit? False
2019-02-01 20:51:20,588 - mu.logic:1195(change_mode) INFO: Workspace directory: /Users/microbit-carlos/mu_code
2019-02-01 20:51:22,962 - mu.logic:1099(show_admin) INFO: Showing admin with logs from /Users/microbit-carlos/Library/Logs/mu/mu.log
2019-02-01 20:51:22,962 - root:94(excepthook) ERROR: Unrecoverable error
  File "/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/app/mu/logic.py", line 1107, in show_admin
    packages = installed_packages()
  File "/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/011FB62A-A14D-48E4-8D23-F9176A545338/d/mu-editor.app/Contents/Resources/app/mu/logic.py", line 160, in installed_packages
    lines = f.readlines()
  File "/Users/travis/build/mu-editor/mu_portable_python_macos/python/lib/python3.6/encodings/ascii.py", line 26, in decode
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 2616: ordinal not in range(128)
2019-02-01 20:51:43,773 - root:112(run) INFO: 
</pre>
</details>

<br>
The problem went away after I removed the 'site_packages' folder previously created by Mu running from this branch.

However I then (in the packaged Mu) installed a module, close the settings and had the same crash when trying to open the settings again in the same session.

<details>
<summary>Log</summary>
<pre>
2019-02-01 21:05:37,085 - root:111(run) INFO: 

-----------------

Starting Mu 1.0.2
2019-02-01 21:05:37,085 - root:112(run) INFO: uname_result(system='Darwin', node='Carlos-MBP.local', release='18.2.0', version='Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64', machine='x86_64', processor='i386')
2019-02-01 21:05:37,085 - root:113(run) INFO: Python path: ['/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/MacOS', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/python/lib/python36.zip', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/python/lib/python3.6', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/python/lib/python3.6/lib-dynload', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/python/lib/python3.6/site-packages', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app_packages', '/Users/microbit-carlos/Library/Application Support/mu/site-packages', '/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app_packages/IPython/extensions']
2019-02-01 21:05:37,085 - root:114(run) INFO: Language code: en_GB
2019-02-01 21:05:37,398 - mu.logic:562(__init__) INFO: Setting up editor.
2019-02-01 21:05:37,398 - mu.logic:582(__init__) DEBUG: Creating directory: /Users/microbit-carlos/Library/Application Support/mu/site-packages
2019-02-01 21:05:37,399 - mu.logic:335(get_admin_file_path) DEBUG: Creating admin file: /Users/microbit-carlos/Library/Application Support/mu/settings.json
2019-02-01 21:05:37,399 - mu.logic:584(__init__) INFO: Settings path: /Users/microbit-carlos/Library/Application Support/mu/settings.json
2019-02-01 21:05:37,399 - mu.logic:335(get_admin_file_path) DEBUG: Creating admin file: /Users/microbit-carlos/Library/Application Support/mu/session.json
2019-02-01 21:05:37,399 - mu.logic:585(__init__) INFO: Session path: /Users/microbit-carlos/Library/Application Support/mu/session.json
2019-02-01 21:05:37,399 - mu.logic:586(__init__) INFO: Log directory: /Users/microbit-carlos/Library/Logs/mu
2019-02-01 21:05:37,399 - mu.logic:587(__init__) INFO: Data directory: /Users/microbit-carlos/Library/Application Support/mu
2019-02-01 21:05:37,410 - mu.logic:600(setup) INFO: Available modes: python, adafruit, microbit, debugger, pygamezero
2019-02-01 21:05:37,544 - mu.logic:1195(change_mode) INFO: Workspace directory: /Users/microbit-carlos/mu_code
2019-02-01 21:05:37,548 - mu.logic:649(restore_session) INFO: Restoring session from: /Users/microbit-carlos/Library/Application Support/mu/session.json
2019-02-01 21:05:37,548 - mu.logic:650(restore_session) DEBUG: {}
2019-02-01 21:05:37,549 - mu.logic:1152(select_mode) INFO: Showing available modes: ['python', 'adafruit', 'microbit', 'debugger', 'pygamezero']
2019-02-01 21:05:38,897 - mu.logic:702(restore_session) INFO: Starting with blank file.
2019-02-01 21:05:39,010 - mu.logic:1195(change_mode) INFO: Workspace directory: /Users/microbit-carlos/mu_code
2019-02-01 21:05:41,167 - mu.logic:1099(show_admin) INFO: Showing admin with logs from /Users/microbit-carlos/Library/Logs/mu/mu.log
2019-02-01 21:06:02,062 - mu.logic:1099(show_admin) INFO: Showing admin with logs from /Users/microbit-carlos/Library/Logs/mu/mu.log
2019-02-01 21:06:02,064 - root:94(excepthook) ERROR: Unrecoverable error
  File "/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app/mu/logic.py", line 1107, in show_admin
    packages = installed_packages()
  File "/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app/mu/logic.py", line 160, in installed_packages
    lines = f.readlines()
  File "/Users/travis/build/mu-editor/mu_portable_python_macos/python/lib/python3.6/encodings/ascii.py", line 26, in decode
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 2616: ordinal not in range(128)
</pre>
</details>

```
2019-02-01 21:06:02,064 - root:94(excepthook) ERROR: Unrecoverable error
  File "/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app/mu/logic.py", line 1107, in show_admin
    packages = installed_packages()
  File "/private/var/folders/l1/szsxrhfj7v1fglxfv9b35lz80000gn/T/AppTranslocation/A6CF3417-08AD-48D7-87EC-60E57CF4CCD0/d/mu-editor.app/Contents/Resources/app/mu/logic.py", line 160, in installed_packages
    lines = f.readlines()
  File "/Users/travis/build/mu-editor/mu_portable_python_macos/python/lib/python3.6/encodings/ascii.py", line 26, in decode
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 2616: ordinal not in range(128)
```

I've also noticed the log doesn't include any entries during the installation process, would be useful to have something logged for debugging.

tjguk:
Thanks for building, @carlosperate . I've just tested the Win64 build and it seems to work ok. (Added a package; removed a package).

bennuttall:
I had a quick look over the code earlier and noticed you delete package files based on the package name as installed - you can't assume that package files will match the package name at all unfortunately - for example the package `opencv-python` is `cv2`. How come you're not using pip to uninstall?

tjguk:
@bennuttall I think that's explained in the final bullet point under "How does it work?" above.

ntoll:
@carlosperate awesome commentary there matey... I'll try to address each of your points by the end of the weekend. 

WRT confused users... I'm at a loss for what to do. We **clearly state** in our docs on the website that the plotter doesn't work on Raspberry Pi. We have no way to change this situation since it's a problem to do with Raspbian's ancient PyQt package not having QtCharts. I guess your suggestion of keeping the tab but putting in a helpful "Feature not available on Raspberry Pi" message may save us some bug reports (but I also suspect we'll get bug reports that say, "make this tab work on Raspberry Pi"). ;-) So we're in a sort of no-win situation -- hence my desire to get this in front of real teachers and learners so they have a plenty of time to comment and provide feedback.

@bennuttall the `pip` command doesn't do targetted uninstall (see "How does it work"). :-( The files which are deleted are *only* those referenced from package metadata (i.e. Mu isn't guessing which files to delete). In terms of package name discovery, the *metadata* is always in a directory which matches a particular known pattern. So, for example, the metadata for `opencv-python` is in `opencv_python-4.0.0.21.dist-info` (I just checked to confirm this). The metadata within that directory tells Mu that the files to uninstall are in the `cv2` directory. I can confirm that Mu correctly installs and uninstalls `opencv-python` using this process of depending on metadata. Does this make sense? Put simply, deletion **isn't** done via pattern matching on the name (that would be dangerous!). ;-)

Here's what I'll  try to do next:

* Keep the tab in RPi, but with a "Not available" message instead of the functioning UI.
* Update, address, check @carlosperate's further code related comments.
* Wait for confirmation that the crash is fixed on OSX (it works for me on my Mac - I suspect on ascii characters in your path as the cause).
* Merge this (and some other outstanding PRs) with a view to Mu1.1 alpha 1 sometime next week.

As always, comments, constructive critique and ideas most welcome.

carlosperate:
I think the problem with the macOS packaged version crashing is to do with the content of the METADA file rather than the file path (fails on line 160):
https://github.com/mu-editor/mu/blob/aa346178f3e122e1a22e08822210c897fb1f152c/mu/logic.py#L157-L162

It can be replicated by installing any module, then editing its METADATA file to include a Chinese character (我) in the description area, and then trying to open the settings window.

ntoll:
@carlosperate OK... so what is in your problematic METADATA file (and how did it get there?). Could you attach it to this issue please? :-)

carlosperate:
Problem appeared for one of the ipython dependencies, which turned out to be `pexpect`. Metadata file attached: [METADATA.zip](https://github.com/mu-editor/mu/files/2825579/METADATA.zip)

ntoll:
I have a five hour train journey today (and tomorrow0 so will try to fix the remaining issues while speeding to and from Scotland. ;-)

mu/logic.py
ntoll: The convention from Python packages means the result of `installed_packages` stored in `packages` will always contain lower-case package names.

mu/logic.py
ntoll: You're right! Good spot. :-)

ntoll:
I've finally found the time to complete the work on this branch as I described.

The "Third Party Packages" tab appears in Raspberry Pi, but contains a "does not work on RPi" type message instead of the usual form. @bennuttall can you please check this works OK..?

@carlosperate I was **unable** to recreate your encoding problem, even when I used the faulty METADATA file you attached to this issue. I have no idea what's going on there. However, to mitigate this sort of problem, I've ensured the file is read with the `"rb"` flag (as bytes) and only then extract the package name from the first line. Please can you test this on your system. It *feels* like it could be something odd about your setup since it works *fine* for me on OSX. :-/

Once the CI is green I'll merge this as promised with a view to an imminent Alpha 1

carlosperate:
@ntoll were you using the final app bundle after building locally?
I only encounter this error with the packaged output, not when running with my local Python from the terminal.

ntoll:
@carlosperate aha... I see what you mean. Nope. As far as I can tell, my changes *should* fix it though. I guess there's some encoding being set in your terminal vs encoding used when launching an app. *sigh* I hate computers.

ntoll:
*sigh* the update to Qt5.12 has broken the build. I have a meeting in 2 mins, but will try to fix later. :-/

mu/logic.py
carlosperate: Pygments appears in that list with a capital letter, so if I guess the standard isn't enforced, or something else is going on.

ntoll: So I've just tried installing and removing Pygmants. It works fine. 

I'm not sure what notion of a "false positive" you're getting at here. Looking at the code, the set operations and the way the directories ending in `dist-info` are used, I honestly don't believe this is a problem. 

Please can you give me a **concrete** example of where a failure occurs.

ntoll: By concrete I mean repeatable steps I can follow to observe the false positive. Otherwise I'm investing lots of time in trying (and guessing) to recreate the "problem state", and that's no fun for anyone. :-(

carlosperate: I've created this GIF (big 4Mbs, better viewed in its own tab), hope it makes things more clear, please do let me know if it doesn't and I'll provide more detailed steps.



In essence 'Pygments' is constantly uninstalled and reinstalled every time the settings window is closed because of the capitalisation. So the package can be installed and removed correctly, but every time the list is checked it will go through that cycle.

ntoll: @carlosperate you :star: ... Now I **SEE** what you mean. Thank you. I'll fix this tomorrow (dad-duty beckons)..!

bennuttall:
Sorry for the delay @ntoll - I can confirm that the third party package tab shows the "not supported" message on Raspberry Pi. Thanks again for dealing with the edge case :+1:

bennuttall (approved):

carlosperate:
> @carlosperate aha... I see what you mean. Nope. As far as I can tell, my changes _should_ fix it though. I guess there's some encoding being set in your terminal vs encoding used when launching an app.

Yeah, maybe an environmental variable somehow indicating default encoding, similar to the issue we had with detecting the default system locale.

Anyway, I've just built a mu-editor.app package locally and cannot replicate the MANIFEST decoding issue any more 😃 🎉

ntoll:
@carlosperate good stuff. I'll fix the false-positive stuff in the morning.

mu/logic.py
ntoll: OK... fixed with latest push. Please re-test. Thank you! (Now to try to fix the damn builds)

ntoll:
OK... I finally got the Travis related problems fixed so will merge back to master now that the outstanding issues have been addressed.

Thank you to everyone for all your help with the creation of this feature. I really really appreciate the your time, care and attention taken to improve this PR from my shonky bodge. ;-)

Onwards and upwards...

Daniels118:
On Windows, if adding packages fails with error `CERTIFICATE_VERIFY_FAILED`, then you may append base64-encoded custom certificates to the file
`%LocalAppData%\python\mu\mu_venv-xx-xxxxxxxx-xxxxxx\Lib\site-packages\pip\_vendor\certifi\cacert.pem`.

PR #771, opened 2019-02-12
Adds stripped newline to end

### LINKED ISSUE #769: When saving files on mac, final newline isn't saved
If you are reporting a bug, we would like to know:

* What you were trying to do,
Edit files
* What steps you took to make this happen,
On a mac (haven't checked windows), do the following:
   1. Create a new file
   2. Make sure it is 10 lines long with a couple empty lines at the end
   3. Save the File
   4. Close the tab
   5. Open the file
   6. Notice there is 1 less line than before
* What you expected to happen,
   That there would be the same number of lines
* What actually happened,
   The last newline character was stripped
* Why this difference is problematic (it may not be a bug!),
   When trying to create CircuitPython libraries, then needs to be one empty line and this frequently results in having to commit code an extra time to fix it.
* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 
Mu 1.0.2 and MacOS 10.14.3

Please remember to attach a **copy of the full log files for Mu**. You can get
the logs by clicking on the cog icon in the bottom right of the editor window.
Click on the logs and use CTRL-A to select all, then CTRL-C to copy and CTRL-V
to paste the contents into the issue.

Thank you for contributing to Mu! :-)

makermelissa:
I'm not sure if this is the best way to go about it, but it fixes #769. The other behavior (which I think is desirable) is that if the file does not have a newline at the end, this adds it. I edited the .py file in my Mu Editor App folder and tested that this change works.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -179,7 +179,7 @@ def save_and_encode(text, filepath, newline=os.linesep):
         encoding = ENCODING
 
     with open(filepath, "w", encoding=encoding, newline='') as f:
-        text_to_write = newline.join(l.rstrip(" ") for l in text.splitlines())
+        text_to_write = newline.join(l.rstrip(" ") for l in text.splitlines()) + newline
         write_and_flush(f, text_to_write)
 
 


ntoll:
Hi @makermelissa!

Thank you for your PR. I've pinged @tjguk as a reviewer since I believe you've already been chatting about this issue.

For the record, the build errors you see have been caused by a recent update to some of the libraries we rely on. It's on my radar to fix this very soon.

Thank you once again,

Nicholas.

tjguk:
Thanks @makermelissa; I hope to look at this soon (although at a quick glance it looks fine).

Are you in a position to update our automated tests? (https://mu.readthedocs.io/en/latest/tests.html)

If not, I can do so myself.

makermelissa:
Hi @tjguk, unfortunately I'm not in a position to update those, but thanks for asking.

makermelissa:
FYI, I've been testing this since I submitted it and it's working great on my system.

ntoll:
Hi, so looking at this PR I notice a bunch of tests have broken as a result. @tjguk many of these tests relate to the testing of writing and properly encoding the users code (i.e. they're tests you've written). I've taken a *quick* look and figured out this is not as simple as "just add a newline to various assertions". ;-) Care to take a look (my main reason for hesitating myself is not that I can't fix things, but that I don't have the context for appreciating what the tests are doing and don't want to cock things up!!!). ;-)

tjguk:
Will do. What timescale are you chasing for this release? I'm rather tight for time these next few days but I can almost certainly manage an hour or so to try to nail this

ntoll:
@tjguk thanks for the quick response. No rush on this... perhaps a "yes, this really is just adding newline to the failing asserts" type comment would be enough and I can fix things up. ;-)

ntoll:
OK... I fixed up the tests and merged this! Thank you @makermelissa for your contribution.

Issue #776, opened 2019-02-19
Clean up requirements.txt / setup.py 


ntoll:
There are various issues and inconsistencies in the way we "do" `requirements.txt` and `setup.py`. These should be addressed. The summary is:

* The versioning of modules is inconsistent between `requirements.txt` (which will install the latest version of modules) and `setup.py` which installs "pinned" versions. See the points raised by @tmontes  here: https://github.com/mu-editor/mu/pull/764#issuecomment-464382806
* The `setup.py` file should **NOT** import `mu/__init__py` in order to discern the current version. See the points raised by @tmontes here: https://github.com/mu-editor/mu/pull/764#issuecomment-464915103

My initial thought is that we should do "the right thing" as per Pythonic conventions and Tiago's suggestions in the comments linked to above are a good starting point. Please add further thoughts below.

carlosperate:
## General patterns
The following is a massive generalisation, but most Python projects usually fall into one of these two categories:
* Libraries: They want to be as compatible with as many different dependency versions as possible, they are usually hosted in PyPi.
* Applications: They want to have their dependencies as locked as possible. I'm not sure, but I think most of them are probably not hosted in PyPi (or very few are), instead they get deployed to a server, or somehow packaged for installation on different platforms (wheels sometimes used for this as well).

The general patterns I've noticed is that a lot of projects have unpinned versions in `setup.py` (or as unpinned as possible) and fully pinned versions in `requirements.txt` (i.e. `pip freeze` output) .
This is because is common to use `requirements.txt` for "application-like" packages, were they deploy to a server, and installing exact versions is recommended.

## Our patterns
In our case I would consider Mu an application, but it is also delivered via PyPi, making it have some of the library requirements as well, which puts us in a more complicated position for this.

In our use case we need to have, at least, some minimum versions in `setup.py`, due to bug fixes affecting us, incompatibilities between specific versions, or features we use and only included from a specific version.
We also cannot use the latest versions of some packages because some of these dependencies are fulfilled by debian/raspbian available packages.

We currently use `requirements.txt` as dev dependencies, and `setup.py` for the app dependencies. I guess we can move the dev dependencies to an `extras_require` and install with something like `pip install -e .[dev]`. Sounds like a good alternative to me, although I have not use this before, so don't know if there are any disadvantages.

## Current Python packaging state

[...]

PR #779, opened 2019-02-27
Single source of dependency truth.

tmontes:
Following the discussion in #776, this PR essentially implements the idea of having a single source of dependency truth (it goes a little further though).

`setup.py` is now the single file where Mu dependencies are declared. Separated between:
* Runtime dependencies, as defined in `install_requires`.
* Extra dependencies, for testing, docs, packaging, and more, as defined in `extras_require`.
* These dependencies include environment markers such that, for example, `PyQt5` is not a dependency on the ARM Raspberry PI, or `black` is not a dependency unless Python is >= 3.6.

As such, the files `requirements.txt`, `win_installer32.cfg` and `win_installer64.cfg` are gone.

Setting up a development environment is now `pip install -e .[dev]` on any platform (yes, on the ARM RPI, the Rasbpian packages for PyQt5 and friends still need to be installed). Full details in the docs, also updated in this PR.

Major implications of such change:

* `setup.py` could not import `mu`, otherwise installation on an fresh environment would fail.

  This was addressed with the somewhat naive (but effective!) parsing of `mu/__init__.py` at the top of `setup.py`, in order to collect "about" information which, by the way, was extended with more details (and is also used throughout the Windows packaging process).
  I originally had a simpler and more readable approach at getting the "about" information, but pycodestyle didn't like it and, while this one is a little more brittle, it still is very compact.

* Windows packaging (this is where I had to put more effort and slay a few Dragons). Facts:
  * We're using `pynsist` which needs a configuration file to tell it how to package and which dependencies to bring in.
  * Such dependency specification is not as flexible as `setup.py`'s (or `requirements.txt`, for that matter).
  * It needs to be split between "dependencies which have wheels on PyPI" and "the others".
  * For the former, the dependencies must be specified in a "name==version" format and `pynsist` will fetch them from PyPI at packaging time.
  * For the latter, the dependencies must be specified by "name" only and `pynsist` will copy then from the current Python environment at packaging time.
  * So, if we are to have a single source of dependency truth in `setup.py` (which we need to publish Mu on PyPI and to package for macOS, by the way!) then a new approach was needed.
  * I revamped the `win_installer.py` script to produce the same exact result but with automatic configuration file generation, such that the dependencies declared in `setup.py` are used. Process:
    * Create a temporary directory (don't mess up the source tree).
    * Create a virtual environment dedicated to packaged Windows generation.
    * Install Mu and its runtime dependencies only into it (yes, with `python -m pip install .`)
    * Collect the output of `pip freeze --all`.
    * Use that list and [`yarg`](https://pypi.org/project/yarg/) to determine which dependencies exist as wheels on PyPI (much like `pynsist` itself uses) and which ones don't.
    * With those two lists, write a `pynsist` configuration file based on a builtin template, mostly adapted from the former `win_installer32/64.cfg` files (by the way, the configuration file will include things like version, author, etc. also sourced from `mu/__init__.py` which itself is a single source of other truths like the version number -- no more changing version numbers in multiple places to cut a release!).
    * After that, install `pynsist` into the virtual environment and run it.
    * Finally, copy the resulting executable to the `dist` directory in the repository and clean up the temporary directory.
    * PS: The process of collecting the TkInter related assets is still in place, as before.

* ~~macOS packaging (some friendly fun, here): the current `briefcase` version does not support the environment markers in `setup.py`. I fixed that and issued a [PR](https://github.com/pybee/briefcase/pull/195) that @freakboy3742 promptly merged (how cool is that?).~~ [UPDATE 2019-03-20: `briefcase` 0.2.9 released, macOS packaging depends on that now]

* The Travis and AppVeyor configuration scripts have been adjusted to the new reality: I did run the AppVeyor one successfully on my fork (good!) but failed to convince Travis to run a job against a non-master branch (I'm pretty confident it will work, however).

Things of note that may need changes before a merge:

* To make life simpler and trying to shorten the time of the bazillion installs/builds/rebuilds/etcs. I ran, I excluded the following packages from the Mu runtime dependencies: `requests`, `matplotlib`, and `Pillow`. Given that Mu can now install 3rd party packages, maybe we can leave it at that (or, maybe, given that it can't do that on the RPI -- am I correct? -- we add them as runtime dependencies with an environment marker for the ARM RPI)

* ~~I haven't changed any of the locked dependencies in `setup.py`. They are the same that were there before... I expect everything to work smoothly, regardless of that: installation- and packaging-wise, that is!~~ [UPDATE 2019-03-20: I updated this a bit, chasing a macOS intermittent test failure, moving `PyQt5` from 5.11.3 to 5.12.1, `QScintilla` from 2.10.8 to 2.11.1, `PyQtChart` from 5.11.3 to 5.12, and `qtconsole` from 4.3.1 to 4.4.3]

* I also haven't changed the Python version currently hard-coded to 3.6.3 (as before) in the Windows packaging process. As above, I expect everything to work smoothly, regardless of that, packaging-wise.

* ~~When / if this PR is deemed mergeable, we'd ideally ask @freakboy3742 to cut a new `briefcase` release and update both `setup.py` and `.travis.yml`, both temporarily set to use its master branch. With that out of the way and done "properly", we'd then merge.~~  [UPDATE 2019-03-20: `briefcase` 0.2.9 released, macOS packaging depends on that now]

* AFAICT, the `dmgbuild` dependency that was installed in the previous `.travis.yml` wasn't used. Per the [Release Process documentation](https://mu.readthedocs.io/en/latest/release.html), it is used after the macOS signing step, so it may make sense to include it as an extra `[package]` dependency. However, I have found it to fail running on Python 3... How were the existing `.dmg` files created, with Python 2?

* I did not understand why `simplegeneric` was used in Windows packaging and threw it away in this PR. Everything seems to be working, but I'm probably missing something... Help? :)

What falls outside the scope of this PR but I feel can be useful (so these are all maybes):
* Have the `win_installer.py` script use the current Python version as the Python version to be used in the installer package, instead of the hard-coded value there (unchanged by this PR, at 3.6.3). What sense does it make to run tests with one Python version and package it with another?
* Move the `win_installer.py` script into a `package/windows` directory.
* Move the `debian` directory into `package` (not even sure if possible, but hopefully yes!).
* Reorganize the `package` directory with the `macOS` and `linux` directories and respective files.

Hopefully I've covered it all. If not, discussion will bring it up. Thanks to all for reviewing this. I'll be happy to explain, discuss, review, etc. any topic you may consider under this scope.

PS: I finally confirmed that the AppVeyor log lines are not in the correct order. Once this PR is submitted, it'll be very easy to point out which lines are being displayed after others that, actually, were produced before them (what are they doing there?!).


tmontes:
One last, not irrelevant, thing:
* If anyone wants, I'll be happy to schedule a 15-30m video call to walk you through the changes.
* Hints to read the diff:
  * Start with `mu/__init__.py` then review `setup.py`.
  * Move on to `win_installer.py`: read the top comments and start from `def run()` at line 240.
  * Review the `Makefile` and `make.py` with trivial changes.
  * Review the Travis and AppVeyor changes (note how the latter no longer depends on the version!).
  * Review the docs.

Enjoy. And thanks for taking the time with this.

tmontes:
Interesting test results:
* AppVeyor passed.
* Travis failed with Python 3.5. Motive: `win_installer.py` uses f-strings.
* Travis failed on macOS. Wow, that's unexpected. Especially because that's my main development platform and `make check` passes all tests here. Hmmm, interesting...

Next:
* Think about removing f-strings from `win_installer.py` vs. excluding it from the check on Python < 3.6 given that it's only used on Windows, where >= 3.6 is used.
* Analyse the macOS failure log and figure out what went wrong there. :)

tmontes:
Test results:
* AppVeyor still passing, as expected.
* Travis on Python 3.5 now ok, as expected, after dropping f-strings from `win_installer.py`.
* Travis on macOS now passing, with no changes from the previous commit which failed (?!?!?!).

Thoughts on Travis macOS tests:
* Not sure what to think of the results:
  * Did not expect commit ce591e4 to fail, but it did.
  * Latest commit, c62bed9, unchanged from Travis and macOS perspective, passed.
* Very interesting... Talk about feeling uneasy about the tests or testing infrastructure! :)

Unexpected ce591e4 test failure analysis:
* Failure of the [`test_PythonProcessPane_start_process_windows_path_with_exception`](https://travis-ci.org/mu-editor/mu/jobs/499558509#L776) test in `tests/interface/test_panes.py` with a stack trace showing an exception raised by the `mock` module in the standard library:
  * ...while handling a mocked `builtins.open` [here](https://travis-ci.org/mu-editor/mu/jobs/499558509#L826)...
  * ...called by `pkg_resources.resource_string`...
  * ...within `mu/resources/__init__.py` while running `load_font_data`...
  * ...called by `self.setFont(Font().load())` in `PythonProcessPane.__init__` in the `ppp = mu.interface.panes.PythonProcessPane()` assignment [here](https://travis-ci.org/mu-editor/mu/jobs/499558509#L828).
* Digged further:
  * The mocked `builtins.open` was created with a `side_effect` set to `Exception("Boom")`, [as it is very clear here](https://travis-ci.org/mu-editor/mu/jobs/499558509#L820)...
  * ...to test a particular runtime failure when calling `ppp.start_process(...)`, in the next test code line, that wasn't reached because of the failure instantiating `PythonProcessPane`.
  * So, after all, I guess the failure makes sense and is actually to be expected!
  * If that is the case, then why on earth is this test passing (nearly) everywhere / every time?!
  * Are we facing some kind of subtly hidden away race condition? It definitely does not look like it.
  * Head exploding slowly... :)
* Side notes:
  * The standard library's `unitttest.mock` module has been changed a few times between 3.6.5 (the version running this test) and 3.6.8... Could that be related? Not convinced at all.
  * This PR changes nothing WRT to the failed test, AFAICT. Thus, the question "what is going on with this test?", while valuable, may need to be addressed under a different scope.

(as if fighting the packaging Dragons wasn't enough!) :)

ntoll:
@tmontes this is epic work.

I'm currently trying to fix the Windows installer for the first alpha (long story). But I want to let you know that I'd like to get these changes in a later alpha release.

To shed some more light on things... Python3.6 is hard coded as a version because we need to have the correctly linked tkinter dlls "mixed into" the installer.

Thanks for all the work!

tmontes:
Progress notes:
* Updated macOS packaging in `setup.py` and `.travis.yml` to depend on recently published `briefcase` 0.2.9, with support for environment markers in `setup.py`.
* Fixed merge conflicts with current `master`.
* Fixed `setup.py` reading of external files which are not necessarily ASCII-only; in particular, CHANGES.rst, recently became non-ASCII-only.
* Updated key dependency versions which seemed to be leading to macOS test failures in `test_Editor_connect_margin_1_works` -- which, by the way, would be something I suspect we want. macOS tests are still failing -- intermittently, how fun -- not 100% sure what's going on. Need investigation.
* Updated `.travis.yml` with DRY in mind: macOS packaging to use the `Makefile`, and not to run tests/checks twice.

tmontes:
More progress:
* Created PR #794 to address an intermittently failing test on macOS, preventing full reasoning and validation of this PR, in particular in what relates to the macOS Travis job.
* Had Travis update not only `pip` but also `setuptools` for the macOS (and Linux) jobs, given that it seems to have lead to a failure in [this run](https://travis-ci.org/mu-editor/mu/jobs/509056161) where luckily the macOS tests passed but packaging with `briefcase` failed complaining about the `setuptools` version.

tmontes:
Progress:
* Updated branch to use the custom `pynsist` script created by @ntoll in the context of #791, commit e20976fafdbeaa8731b6c6756dec761b209ab315.
* Generated windows installer gives us a Mu that now successfully installs `arrr`.

Next:
* Will try to simplify the customised `pynsist` script which seems to include a bunch of unneeded/unused definitions, making it harder to read that needed.

ntoll:
This pull request **fixes 10 alerts** when merging ca3fceb6a243400b4ae2695cbe9c6ffdb82a8544 into f297c49d9a58fbed2d5b47d2056797f6910c469a - [view on LGTM.com](https://lgtm.com/projects/g/mu-editor/mu/rev/pr-0094202a8181c40dc6b47ce69576d97b3cbbdeb7)

**fixed alerts:**

* 10 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

tmontes:
Current status:
* PR in merge-able state.
* Packaging methods/tools unchanged from current `master`, but now automated from a single source of truth in `setup.py`.
* Single source of truth for version/author/etc. in `mu/__init__.py`, no need to change multiple files when cutting a release.
* Windows (32/64) and macOS packaged results working, including the issue detected in #791.

Subsequent work:
* Review and pin (or not) dependencies (and their sub-dependencies?).
* Re-organize all packaging related files into the `package` directory?
* Update packaged Python version? (3.6.3 on Windows, not sure which on macOS)
* When packaging, run the tests against the Python version that will be used?

Possible future improvement / simplification:
* Use `briefcase` to produce Windows packaged version, given that it is already being used for macOS?

tmontes:
Quick note: In the context of #809 `simplegeneric` was required for the Windows packaging process. I wonder why my experimentation to date under this branch has shown otherwise. Hmmm... :/

ntoll:
@tmontes I feel that pain. Today has been a series "but how does it work like THAT?" moments. ;-) I guess it comes with the territory.

tmontes:
UPDATE NOTE:
* Updated PR, making it merge-able again, resolving minor conflicts with `master`.
* Passes all tests.
* Builds usable macOS and Windows installers. Hand-tested on both platforms:
  * Runs code in files.
  * Runs the REPL.
  * Installs/removes third party packages, including the ones mentioned in #791 and #809.

ntoll:
This pull request **fixes 10 alerts** when merging 8f5505a01a1133fb0c959b7952eaac3ef5003862 into f67852fa993f2c2749977422e8daea32f93ecf04 - [view on LGTM.com](https://lgtm.com/projects/g/mu-editor/mu/rev/pr-324aaf7dd610329e4bfe663d8ea42bde938070b9)

**fixed alerts:**

* 10 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

setup.py
```diff
@@ -1,43 +1,101 @@
...
-with open('README.rst') as f:
+base_dir = os.path.dirname(__file__)
+
+
+DUNDER_ASSIGN_RE = re.compile(r"""^__\w+__\s*=\s*['"].+['"]$""")
+about = {}
+with open(os.path.join(base_dir, 'mu', '__init__.py'), encoding='utf8') as f:
+    for line in f:
+        if DUNDER_ASSIGN_RE.search(line):
+            exec(line, about)
```
carlosperate: Rather than doing this, would it be possible to move the locale stuff to a different file to keep the __init__.py file free of dependency imports?

tmontes: AFAICT, unfortunately not, due to the nature of `gettext.translation` that, when run, installs the pseudo-builtin `_` functions, use throughout the code. So it needs to be called *very early on*...

PS: To get a feeling of the impact, try commenting out the two locale-setup related lines from `__init.py__` and launch mu. :-)

PPS: Why remove imports from `__init__.py`, by the way? To avoid this code in `setup.py`? IMHO, it is never a good idea to import the module/package to be distributed in there.

tmontes: Even more PS: Thanks for taking a look at this! :)

tjguk:
@tmontes I had an offline conversation with @ntoll and he's asked me to try to review your work here with a view to inclusion in the upcoming release. I'm trying to set aside some time,  but I'm quite sure I'll have questions & queries as I try to work through the codebase. But hopefully we'll be able to get the benefit of all the work you've put into this :)

ntoll:
Hah, @tjguk beat me to it.

@tmontes, my "bandwidth" has been very limited recently and is currently limited due to workload. I DO NOT want all your effort and hard work to remain un-reviewed. In fact, I think this block of work is hugely important for the next release of Mu, ergo, when I saw Tim yesterday I jumped at the chance of his offer to help out with reviews and such like.

Finally, my proposal for a talk about Mu was accepted by EuroPython. If you'll be in Basel, I'll see you there and would love to buy you dinner / a drink.

tmontes:
@tjguk, thanks for taking the time in reviewing this. Much like I [commented earlier](https://github.com/mu-editor/mu/pull/779#issuecomment-468079526), I'll be glad to schedule a call or some such, to introduce you to the PR, if you like. My availability is somewhat limited these days, but I guess I could set apart a time slot this Friday afternoon or Saturday. Let me know if you would like that.

@ntoll, thanks for pushing things forward. :) My proposal for a talk at EP2019 -- reframing the kids workshop I did, as lessons/questions for pro-developers -- wasn't accepted (yet! I've been told I'm on the waiting list). Regardless of that, I expect to be in Basel, and will gladly accept dinner / drink, as long as you accept one dinner / drink back from me! :)

tjguk:
> I'll be glad to schedule a call or some such, to introduce you to the PR, if you like. My availability is somewhat limited these days, but I guess I could set apart a time slot this Friday afternoon or Saturday. Let me know if you would like that.

@tmontes that would be very welcome. I also have a complicated schedule but I'll try to look at my calendar for some time around Fri/Sat and let you know. I'm in the UK; I imagine you're somewhere in Europe?

tmontes:
> @tmontes that would be very welcome. I also have a complicated schedule but I'll try to look at my calendar for some time around Fri/Sat and let you know. I'm in the UK; I imagine you're somewhere in Europe?

@tjguk, I'm in the whereabouts of Lisbon, Portugal. Same time as the UK, too! :)

tjguk:
@tmontes would something like 4.30pm UK time on Saturday work for you? That's just after a Coding Session I run for youngsters so I'll be somewhat in the zone anyway.

Perhaps email me - mail@timgolden.me.uk

tmontes:
@tjguk, that sounds very good. I'll get in touch via email straight away.

tjguk:
I'm about to merge this in now to master. We'll watch the CI builds to see if they fail as a result of this change. 

The PR essentially centralises all dependency and other metadata inside the setup.py. Everything then runs as usual through pip via its develop-mode (`-e`) setting, and making use of the "extras" feature:

`pip install . -e[dev]`

The `[dev]` suffix brings in the dev extras which is a portmanteau of packaging, testing & documentation extras.

Previously the deps were also maintained in the requirements.txt and in the two pynsist config files. The author/version data was also maintained in those files, giving rise to obvious mismatches if the release process wasn't followed rigorously. Now, the pynsist config files are dynamically generated into a fresh venv into which the development version has been pip-installed. This then allows pip freeze to list whatever the current dependencies are which are then inserted into the pynsist config template.

tmontes:
@tjguk, thanks again for having taken the time to review this.

Note that the syntax on your comment does not work. Instead of...

```
$ pip install . -e[dev]
```

...it should be:

```
$ pip install -e .[dev]
```

A short list of possible follow-up tasks:
* Review the dependency specifications in `setup.py`: which to include, which not to include, which to lock/pin, which not to lock/pin.
* Re-organize all packaging related files into the `package` directory? And organize that directory on a per-platform basis?
* Get documentation meta-info needed in `docs/conf.py` from `mu/__init__.py`?


Somewhat related, but clearly independent tasks:
* Update the packaged Python version (currently at 3.6.3 on Windows, not sure on macOS)? Could we also create a single source of truth for that? (we certainly can, not sure it's easy / worth the effort).
* Use `briefcase` to produce Windows packaged version, given that it is already being used for macOS? I suppose this might simplify a lot of the windows packaging code in this PR, but is a bigger change which, IIRC, will produce `.msi` files and not `.exe` files -- there used to be benefits to that too, but I'm not sure where the tech stands, nowadays; it used to be the case that `.msi` file automated mass-deployments were very simple to achieve, something that controlled environments like schools might appreciate.

tjguk:
@tmontes Whoops -- fixed now. Obviously wasn't thinking carefully :o)

I'll let the dust settle a bit, see what if anything gets shaken out as a result of the change for developers. Then, I agree, we could look at the issues you raise above.

Thanks for all.

PR #781, opened 2019-03-04
Modify the Tooltip for sleep() to include Different Behaviour for time.sleep()

amos1969:
Hi

A very minor change.

When I'm helping a kid fix their code, or demonstrating something on the fly, I often need to use `sleep()` but end up using `time.sleep()` by mistake. Almost every time I do this I am caught out by the tooltip telling me that `sleep()` uses milliseconds, forgetting that `time.sleep()` uses seconds instead.

This PR modifies the text of the tooltip, from:

> Put micro:bit to sleep for some milliseconds (1 second = 1000 ms) of time. 
> sleep(2000) gives micro:bit a 2 second nap.

to:

> Put micro:bit to sleep for some milliseconds (1 second = 1000 ms) of time.
> sleep(2000) gives micro:bit a 2 second nap.
> NB: If you use time.sleep(), times are in whole seconds not milliseconds,
> use time.sleep(2) to nap for 2 seconds, in that case.

Thanks
Dave


ntoll:
Yes. Very much this. Thank you for your contribution @amos1969 :-)

PR #783, opened 2019-03-04
enable raspbian packages install

### LINKED ISSUE #782: Enabled Raspbian third party packages install
I have been catching up with @bennuttall, @topshed and @MarcScott and have agreed that it is better to have this functionality available on the Raspberry Pi than not. Without it we disadvantage a subsection of mu's users unnecessarily.

The implementation of the "third party packages" feels like the right first step, its simple and consistent, and as with everything unlikely to be the final one. 

We are all keen to support and explore what could be done in the future to open this up to a wider set of users and support those who would want to install Python packages "system wide" (on any OS) but believe this is the right first step and we should remove the restriction stopping this functionality for Raspberry Pi users.

I have prepped a PR to undo the restriction, which I will raise shortly, hopefully you agree and we can progress. 

:)

martinohanlon:
To resolve #782

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -21,7 +21,6 @@ import sys
 import logging
 import csv
 import shutil
-import platform
 from PyQt5.QtCore import QSize, QProcess, QTimer
 from PyQt5.QtWidgets import (QVBoxLayout, QListWidget, QLabel, QListWidgetItem,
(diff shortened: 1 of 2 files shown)


ntoll:
Thanks Mart,

The style checker is complaining. I'll fix then merge.

N.

Issue #786, opened 2019-03-08
UI/UX: Accidental breakpoint toggling


tmontes:
## Context

Earlier this week I delivered a programming introduction workshop for kids, aged 11-12: with Python and Mu, of course! :) We all had a lot of fun and the kids did indeed learn a lot! :)

## The Issue

One surprising observation I noted was the fact that *all of the four* kids consistently used the mouse/trackpad to position the text editing cursor (instead of the arrow keys on their keyboards) when they wanted to "go somewhere" in their code to change it. In retrospect, this probably makes sense given that they are used to touch oriented devices like modern phones and tablets, on one hand, and that their experience in using computer keyboards is associated to simpler, non-multi-line-editing, text input, on the other -- say, submitting a typed in phrase on YouTube to find videos of their liking. :)

As a result of this, and of non-100% accurate pointing and clicking, they very often ended up inadvertedly triggering a "breakpoint toggle action" when trying to place the text editing cursor right at the start of a given line, to the left of the first column -- or, worse, triggering the "Cannot set breakpoint" notification dialog, if the target line wasn't "breakpointable",

Of course, that led to some frustration in the editing process and questions about what that was. Even though I do believe that having beginners running code through a debugger can be a very valuable learning experience, I'd rather introduce the concepts when I plan to do so (which I didn't, given the short amount of time we had), instead of being somewhat "forced" into a dismissive answer like "that's a different thing we may learn about some other day, try not to worry about it -- oh, by the way, have you tried using the keyboard instead?"... :(

## The Proposal

I figured that a solution to minimize such a usability issue would be to increase the "white space" to the left of the first column, but within the "white editing area" (not on the left margin where the line numbers and breakpoint markers are displayed) -- in HTML/CSS speak, setting a left-padding on the "white editable area" of the editor.

Then I went exploring on how to attain that. After nearly two hours I gave up... Looking at Scintilla's source, that spacing seems to be fixed and hard coded. :(

A solution I found that can be implemented, works the other way around, by adding a "right-padding" to the "gray line number and breakpoints marker left margin". I'll be sharing a PR with that, for evaluation.

## The Question

I wonder if others have observed beginner kids do the same thing: using the mouse/trackpad to position the text editing cursor, instead of the arrow keys on the keyboard. More, if you observed that, did you notice kids accidentally triggering breakpoint toggles? Or are those kids "mouse pointing" experts, even if Python beginners? :)

[...]

PR #787, opened 2019-03-08
Extra margin: no breakpoint toggle if click slightly leftwards of 1st column.

tmontes:
Submitting this PR for evaluation, in the context of #786.

A few notes from my investigation and about this PR:
* What we see as the left gray area in the editor, with line numbers and the breakpoint marks, is what Scintilla calls margins.
* There can be up to 5 margins: each such margin is a vertical column in that gray area.
* Up until now, Mu is using two such margins: the first containing line numbers (margin number 0), the second containing breakpoint marks (margin number 1).
* This PR adds a third margin (but the last available, margin number 4) with no content and relatively narrow (currently hardcoded to 8 pixels wide).
* The "trick" in this PR is having clicks in that margin "do nothing":
  * By default clicks would highlight/select the whole text line in the editor.
  * This PR sets the margin as "sensitive" such that clicks are handled by the "margin click handler" which is common across all margins, and is itself is wrapped in a filtering handler that discards margin 4 clicks. Thus, margin 0 and 1 clicks still work as before, toggling breakpoints.

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -152,6 +152,15 @@ class EditorPane(QsciScintilla):
         self.setMarginSensitivity(0, True)
         self.markerDefine(self.Circle, self.BREAKPOINT_MARKER)
         self.setMarginSensitivity(1, True)
+        # Additional dummy margin to prevent accidental breakpoint toggles when
+        # trying to position the edit cursor to the left of the first column,
+        # using the mouse and not being 100% accurate. This margin needs to be
+        # set with "sensitivity on": otherwise clicking it would select the
+        # whole text line, per QsciScintilla's behaviour. It is up to the
+        # click handler to ignore clicks on this margin: self.connect_margin.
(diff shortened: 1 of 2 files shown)


tmontes:
* Python 3.5 tests failed.
* No `Mock.assert_called_once` in 3.5... *sigh*

tmontes:
Hmmm, what's not to like about fighting the CI on Friday afternoons!... :)

PS: Why on earth has the Travis macOS test failed now when it passed in the previous commit?! No changes to the failed test were made between the two commits. I'll be back! :)

ntoll:
So Travis OSX CI fails at random intervals for no apparent reason. The only thing you can do is to restart the job and wait for it to turn green (as in this case).

So much for determinism.... ;-)

tjguk:
There's a marketing opportunity: "Quantum Integration" (QI for short). It either will or won't pass all the tests, but you can't tell which.

PR #790, opened 2019-03-15
Revised ESP mode

ntoll:
This PR supersedes #676.

It is based upon the work done by @dybber in that PR and revisions which emerged in the resulting discussion. The only reason there's a new PR is because my revisions can't be pushed  onto @dybber's own branch.

@dybber and @murilopolese can you please check this branch out. I've tested it with ESP8266 and ESP32 boards. "It works for me" :-)

If no gotchas are obviously found over the weekend I want to merge into master on Monday.

Many many thanks to @dybber for his amazing work on this PR.


ntoll:
FWIW, I managed to remove a lot of code (and reuse things already in Mu). Please don't hesitate to give constructive critique, ideas and feedback.

My aim *right now* is to get this mode into alpha.1 so we can show it to users. Of course it's marked as "alpha" so folks should expect bugs and instability... (by testing it they can help us squash the bugs). It also means that we can get feedback on the UI and workflow too and have chance to gather feedback from real users and make adjustments before we get to the beta phase of the release cycle.

Thanks..! :-)

dybber:
Great! Here are some initial feedback.

1) When executing a simple example such as a simple "Hello World" program, the output is printed as:
Mu 1 0 2 - untitled -_013
It can be a bit confusing, with all the extra information. If we somehow could simplify the output that would be great. In the ideal case it would work as a normal interactive python (`python -i helloworld.py`), we can probably not achieve that without modifications to MicroPython itself.
At least, a single line break after the "OK" would be nice.

2) If it could be possible to push the run button while the REPL was open, that would be great.

3) File manager still does not support directories. The project file must be named "main.py", and thus in this case mu only supports working on one project at a time, as the main.py must be the one main.py residing in the `mu_code` directory.

ntoll:
Hi @dybber thank you for the quick feedback. Some quick feedback feedback... ;-) Would love to hear your thoughts on my suggestions.

1. Completely agree about the confusing output and having a line-break after `OK`. The current behaviour is just "that's how MicroPython work". However, I believe a simple mitigating act could help here: just prepend `print('\n')` to any set of commands sent to the board via run, so the first thing output is the required newline. ;-)

2. Agree... it could be possible, and perhaps should be possible. My intuition is to allow it and observe how those users who play with the mode in alpha releases react to this.

3. Yeah... it's coming, but I want to get this into master first... remember, an alpha means feature incomplete, and I'll mention this is the release notes too.

If you're OK with the above, I'll make the changes on Monday and then merge. Please also test..! I've already found a "Files" related niggle I want to fix. But, like I said, getting this in front of folks via an alpha release is my main priority at the moment.

Thanks! :-)

dybber:
Sounds good to me! Btw. I'm doing a workshop with 75 danish highschool teachers in Informatics/Programming courses on Wednesday, and would prefer to use an alpha release rather than my own version.

dybber:
We should probably also include some additional allowed VID/PIDs, than just the three I included, to allow more people to test it. 

I might have a NodeMCU and Sparkfun ESP8266 Thing at my office which I could check, but if someone else have more devices they could submit VID/PID here in the thread. On Linux you obtain VID/PID from `lsusb` command.

ntoll:
+1 on the VID/PID combos.

ntoll:
OK... I've addressed points 1 and 2. Point 3 can be fixed for the following alpha.

Please keep giving feedback and/or more VID/PID combos for ESP8266 and ESP32 boards.

dybber:
Some additional comments:

4) It is almost too fast :) It is not visible that you clicked the button, if you run the same program multiple times. This is perhaps an issue that shouldn't be taken here, but an "onclick" animation for the buttons would be nice at some point.

5) I can enable the plotter while the REPL is enabled, but not vice-versa

6) (probably related) Do the following button clicks: Enable REPL, Enable Plotter, Disable Plotter, Disable REPL. Now the run button is still disabled.

mu/modes/esp.py
```diff
@@ -0,0 +1,230 @@
...
+    name = _('ESP MicroPython')
+    description = _("Write MicroPython on ESP8266/ESP32 boards.")
+    icon = 'esp'
+    fs = None
+
+    valid_boards = [
+        (0x1A86, 0x7523),  # Wemos D1 Mini USB VID, PID
+        (0x10C4, 0xEA60),  # Wemos D1 Mini Pro USB VID, PID
+        (0x0403, 0x6015),   # Sparkfun ESP32 Thing VID, PID
+    ]
```
deshipu: Those are the boards that I have immediately available at hand. Note that about half of the actually use ESP8266.
```
| VID | PID | Chip | Board |
| 10c4 | ea60 | CP210x | Adafruit Feather HUZZAH |
| 10c4 | ea60 | CP210x | AMICA NodeMCU |
| 10c4 | ea60 | CP210x | TTGO-W8266 |
| 10c4 | ea60 | CP210x | Heltec HTIT-W8266 |
| 10c4 | ea60 | CP210x | WeMos Lilin32 |
| 10c4 | ea60 | CP210x | TTGO LORA32 |

| 1a86 | 7523 | HL-340 | Lolin32 LITE |
| 1a86 | 7523 | HL-340 | WeMos D1 Mini LITE |
| 1a86 | 7523 | HL-340 | DFRobot WiFi D1 Mini |
| 1a86 | 7523 | HL-340 | TTGO Mini 32 |
| 1a86 | 7523 | HL-340 | Arduino Nano |
```

ntoll: @deshipu many many thanks. I think your list proves that (at least when it comes to ESP8266 boards) we have the two most common VID/PID combos already covered. ;-)

ntoll: I'll annotate the "metadata" you include into the source code for future reference.

ntoll:
@dybber OK... I've addressed 5 and 6. As for 4... fast is good, right? ;-)

@deshipu I've added a comment to that VID/PID block of code, explaining what's going on. AFAICT, we've got the most common combos.

mu/modes/esp.py
cwalther: (0x0403, 0x6015) is an FTDI FT230X, which is also on the https://www.tindie.com/products/onehorse/esp8285-development-board/ . All other ESP boards I have are covered by the CP210x and HL-340.

ntoll:
Right then... I'll merge this into master and cut an Alpha 1 release later this morning.

mu/modes/esp.py
ntoll: @cwalther thanks for the heads up! It feels like we're covered as much as one can reasonably expect at this moment in time. Let's see who shouts during the alpha releases, if at all, in this respect. :+1:

ntoll:
@dybber OK... so releasing the alpha 1 today fell afoul of different bugs in OSX and Windows packaging. The Windows problem breaks another of the new features we want to release (support for third party packages). However, you should be able to use Mu if you install it from source, in a virtualenv using system Python (I'm thinking of your class on Wednesday here). Ugh... bugs.

dybber:
Sorry to hear that, I will use my old version then.

Issue #792, opened 2019-03-19
Error using debug on MacOS Mojave


TommyPujol06:
I just installed the Mu editor on Mojave and I created a test file to see if there are differences between using Mu in Linux and Mac. I pressed on Debug (didn't mean to) and then it said : 
>Overriding built-in editor font due to Issue #552

Then it does nothing.
![Screenshot 2019-03-19 at 12 44 49](https://user-images.githubusercontent.com/43862943/54637499-01ceb180-4a46-11e9-8f1c-5b7dad234405.png)
Mojave version : 10.14.3

ntoll:
Hi @TommyPujol06, thanks for reporting this.

Calling @stestagg. Any thoughts on this..? See our conversation in #552.

stestagg:
Hi @TommyPujol06 

Thanks for this!  It's definitely a weird user experience.

I think what's happening is that functionally, mu is behaving correctly.  When you click on 'Debug', mu pauses on the first executable line, and waits for you to click Continue (or something else).  This is indicated by the yellow background on line 2.

On Windows, you can see this behaviour:
![muwin](https://user-images.githubusercontent.com/704526/54646082-edb2a080-4a95-11e9-8e90-69b561b1afb8.gif)

On OSX, for some reason the 'Overriding...' message is being shown in the console when the debugger starts up (I'm looking into this). and that is drawing the attention away from the debugging buttons above, so it's not clear that Mu is waiting for input to proceed.

I think ideally, we'd output some text to the console in all cases when debug is pressed, to make it clear that mu will pause before the first executable statement.  But we should also stop that font message from being displayed.

carlosperate:
Just for reference: https://github.com/mu-editor/mu/issues/751

PR #793, opened 2019-03-19
#792 - Add help message when starting a debug session

stestagg:
Addresses #792:
* Make the OSX font override warning be a logging.debug message (it's never important) which hides it from the debug console output.
* Add a print to stderr during debugger startup that prints a message to the console
  pointing users to use the Stop/Continue/Step buttons to control execution flow.

Wording of help message is a bit janky, please update as appropriate

Example:


ntoll:
@stestagg thanks for contributing this. I like it and the message isn't at all "janky" (but that's a word I'll borrow). ;-)

PR #794, opened 2019-03-20
Fix intermittently failing macOS test.

tmontes:
PR #787 brought in a test that is failing intermittently on macOS with regards to editor margin click events and possibly associated keyboard modifier keys -- if fails "consistently intermittently" on my dev system, running macOS 10.12. Given that I myself issued that PR, I wonder how I missed it back then... Or maybe something changed? Not sure. :/

This PR changes that test slightly in that the keyboard modifier related information is discarded. Motives:
* Keyboard modifier information is not used by the handler code, in `mu/logic.py`, `Editor.debug_toggle_breakpoint`.
* Allows for a smoother CI ride, helping the evaluation and progress of #779.

Feedback welcome.

### DIFF
--- a/tests/interface/test_editor.py
+++ b/tests/interface/test_editor.py
@@ -129,7 +129,7 @@ def test_Editor_connect_margin_ignores_margin_4():
     ep.connect_margin(mock_fn)
     margin = 4
     line = 0
-    modifiers = Qt.KeyboardModifiers()
+    modifiers = Qt.NoModifier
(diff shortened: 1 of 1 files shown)


ntoll:
OK... I just ran into this. Merged with but one change... `assert_called_once` isn't available in Python 3.5. ;-) I fixed that and it works.

Thanks for the fix!

tmontes:
Thanks for the quick merge Nicholas! :)

Issue #780, opened 2019-03-01
Request mu to allow user to customize which syntax check errors should be ignored


leroyle:
This issue/request is exactly like issue #253. I'm hoping the solution that was implemented for that issue can be enhanced to allow for a better user experience.

There was a change to the syntax checking as a result of issue #253:  "Error checking in mu - request #253  "

That issue/request asked for a way to loosen the syntax checking that mu is running so that young kids do not get bogged down in the nuances of coding style that the PEP8 check is enforcing. As mentioned in that issue, coding style is good but does make it confusing for the younger beginners.

To support issue #253 a change was made to logic.py - check_pycodestyle(code) function. That change overrides "any" ignore rules that may have been specified when pycodestyle's StyleGuide() function was called.

The pycodestyle module has a builtin mechanism to allow the user to configure certain aspects of its behavior. One aspect is to allow the user to define a list of syntax rule numbers which should be ignored during the syntax check. (I don't know when this was added to pycodestyle.)

So, rather than overriding the list of ignores picked up by the pycodestyle module I would suggest merging the mu internal hard coded ignore list with the ignore list picked up by the pycodestyle configuration mechanism. 

 The configuration mechanism is detailed at:
      http://pycodestyle.pycqa.org/en/latest/intro.html#configuration

This would allow the user to add their own syntax rules to the list to be ignored when the check is run. So, for the kids we can relax those rules that do not cause functional errors, but still keep the ability to
enforce better coding standards at a later date.

The downside with my proposed fix incorporated:

If the user does "not" specify a custom configuration file then pycodestyle has a default ignore list that is loaded and merged instead. The current default list contains all of the rules as currently defined in 
the mu hard coded list, plus two extra,
    W504: which looks to be "line break after binary operator"
    E24:  which looks to include "241: multiple spaces after ‘,’" and "242: tab after ‘,’"

mu's current ignore list
        ('E121', 'E123', 'E126', 'E226', 'E302', 'E305', 'E24', 'E704', 'W291', 'W292', 'W293', 'W391', 'W503', )

pycodestyle current default ignore list
        ('E121', 'E123', 'E126', 'E226', 'E24', 'E704', 'W503', 'W504')

The net result is if you do not specify a custom configuration file you end up with a list that contains duplicate rules. Adding the calls to set() and tuple() code as shown in the proposed changes below will eliminate the duplicates within the resulting combined tuple.
             style.options.ignore = tuple(set(ignore))

[...]

PR #800, opened 2019-03-25
Issue #780 allow user defined syntax check override

leroyle:
Changes to allow the user to specify a file to contain a list of error/warning codes which the pycodestyle module should ignore when mu's Check button is pressed. This to shield new kids from having to deal with code style issues that do not impede functionality. We can re-enable to code style checks when they get more comfortable with the language.

The python module pycodestyle.py does currently support the user supplying an override file, unfortunately a previous code change to mu/logic.py resulted in that functionality being lost. This change pulls that support back in.

Per the pycodestyle.py module functionality the override file is located via:
   On Linux/Mac there are a couple of methods used: one has the user setting an environment variable
   "XDG_CONFIG_HOME" which will point to a directory which must contain a file called pycodestyle.
   On Windows the file is expected to be in the users home directory "~\\.pycodestyle".

   This file must contain the following format, the ignore list contains the warnings/errors to be ignored:
   [pycodestyle]
   ignore = E265, EXXX

Further pycodestyle module details can be found at 
[]( http://pycodestyle.pycqa.org/en/latest/intro.html#configuration
)http://pycodestyle.pycqa.org/en/latest/intro.html#configuration

We are currently looking at what is required to add an entry to the "Using Mu" section at codewith.mu to describe how  the user can take advantage of this. Any changes there would need to specify what version of mu contains the override support.

Proposed changes to mu:

      modified:   mu/logic.py check_pycodestyle()
       - update to re-enable allowing user to specify a file which contains warning/error numbers
         that pycodestyle.py module should ignore
       - rather than overwrite the ignore list returned from StyleGuide() we merge that list with the 
         internal hard coded mu list

       - in order to execute a test case:
         the pycodestyle module needs to locate the override file per the above mentioned file location
         scheme. Within test case execution we can manage the Linux/Mac environment variable scheme
         with a bit more code but for Windows we cannot expect to add a file to the test process's home
         directory. Fortunately pycodestyle.StyleGuide() supports a parameter "config_file" that can be
         used to pass a path to the override file. Mu currently sets that param to False.
         - so as a workaround we add a second default parameter to logic.py's  check_pycodestyle()
           changing it 
            From:   check_pycodestyle(code)
            To:        check_pycodestyle(code,  config_files=False)
                            
         Thus when we call check_pycodestyle() from the test case, test_logic.py, we can pass
         StyleGuide() a path within the test directory structure where the file is located. During normal mu
         operation the default value of False is passed thus calling StyleGuide() as is done previously
         allowing pycodestyle's to use its normal config file location scheme.

Proposed changes to the test framework. We add a sample override file, one test case that confirms the override file works. Add a second test case to confirm that we still honor one of the mu internally defined error code overrides
 
      new file:   tests/scripts/pycodestyle
                    the user customized syntax check override configuration file used during our test
 
     modified:   tests/test_logic.py
                    updated to add two test cases
                    1. test_check_pycodestyle_custom_override(): verify the override file is read
                        and error E265 is ignored
                    2.  test_check_pycodestyle_E121(): verify that mu internally defined overrides are
                         still being honored by verifying E123 is ignored as expected

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -431,7 +431,7 @@ def check_flake(filename, code, builtins=None):
     return feedback
 
 
-def check_pycodestyle(code):
+def check_pycodestyle(code, config_file=False):
     """
     Given some code, uses the PyCodeStyle module (was PEP8) to return a list
     of items describing issues of coding style. See:
@@ -446,8 +446,15 @@ def check_pycodestyle(code):
     # Configure which PEP8 rules to ignore.
     ignore = ('E121', 'E123', 'E126', 'E226', 'E203', 'E302', 'E305', 'E24',
               'E704', 'W291', 'W292', 'W293', 'W391', 'W503', )
-    style = StyleGuide(parse_argv=False, config_file=False)
-    style.options.ignore = ignore
+    style = StyleGuide(parse_argv=False, config_file=config_file)
+
+    # StyleGuide() returns pycodestyle module's own ignore list. That list may
+    # be a default list or a custom list provided by the user
+    # merge the above ignore list with StyleGuide() returned list, then
+    # remove duplicates with set(), convert back to tuple()
+    ignore = style.options.ignore + ignore
+    style.options.ignore = tuple(set(ignore))
+
     checker = Checker(code_filename, options=style.options)
     # Re-route stdout to a temporary buffer to be parsed below.
     temp_out = io.StringIO()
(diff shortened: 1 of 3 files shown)


ntoll:
@leroyle thank you for your contribution, it's most appreciated. :-)

This all LGTM. Merging so we get user feedback as part of the next alpha.2 release.

Issue #802, opened 2019-03-30
Asymmetrical Save / Load constraints


tmontes:
## The Issue 

While exploring a fix for #755 I realised that Mu lets users save files with arbitrary names and extensions but that it will only load the ones with a given subset of extensions, depending on the current mode: `*.py` files in all modes, `*.hex` files too, in Microbit mode.

I'm not sure whether the Save constraints should be tightened or the load ones relieved: but one of those things should, for the sanity of users. As of today, with Mu 1.1.0a1, I can save a file as `words.txt` but can't load after the fact: potentially very frustrating. :)

## Steps to reproduce

* Launch Mu.
* Create New file.
* Type a few words: "Hello world".
* Click the Save button.
* Enter `words.txt` the filename and save.
* The editor tab name correctly indicates `words.txt`.
* Close the tab.
* Click the Load button.
* The load file dialog does not show (Windows) / does not allow the selection of (macOS) the `words.txt` file for loading.

## Platforms

Verified on macOS and Windows.

ntoll:
The issue is that Mu is a *Python* editor rather than a general editor, so the syntax highlighting won't work for any other file type (except for CSS and HTML in web mode). Mu assumes we're editing Python rather than anything else.

Happy to hear suggestions for a way forward on this. They may range from, "just don't do that (open non-.py files in Mu)" to "we should just let it happen and be damned...".

Thoughts..?

PR #803, opened 2019-03-30
Fix silent save overwrite

### LINKED ISSUE #755: Should not overwrite files without confirmation.
## Environment and Version

* Windows 10 1803 and macOS 10.12.6.
* Packaged Mu 1.0.2 from https://codewith.mu/en/download.

## Observed behaviour

* Saving a new/untitled editor contents with the name of an existing file will overwrite it without warning if the typed in filename does not include the trailing `.py` extension.
* This becomes particularly "crazy" when two editor tabs are open:
  * First tab, called `a.py`.
  * Second tab (untitled) saved as `a`, ends up as `a.py` on the filesystem with no warning.
  * User has two tabs named `a.py` with different contents.
  * Running the first tab actually runs the code in the second tab.
  * Yikes! :)

## Expected behaviour

* Mu should protect its users from accidentally overwriting files they might have previously created, even when they don't explicitly name the files with the trailing `.py` at save time.

## Steps to reproduce

* Close all editor tabs.
* Create a new file with a single statement `print("a")`, save it as `a` or `a.py`: it will end up as `a.py` in the filesystem.
* Close the tab.
* Create a new file with a single statement `print("new")`, save it as `a` omitting the trailing `.py`.
* The first `a.py` file was just overwritten without warning.

## Wrapping notes

* I haven't investigated further.
* Feels like it should be an easy fix given that Mu already protects the users from overwriting files when they provide a filename explicitly ending in `.py`.
* Could this be related to #722?
* What about #595, could it have been a case like this?

tmontes:
This PR fixes #755, mostly by making full use of Qt's save file dialog capabilities.

Up to this point, it was this dialog that warned/prevented silent overwrites on save: the dialog itself checked the submitted filename against existing disk files. This only worked for fully matching filenames. With regards to auto-added `.py` extensions, it was completely passive: that was handled by code in `logic.py` that worked with the filename obtained from the Qt dialog and, in turn, added the `.py` extension but did not check for name collisions, leading to possible silent overwrites on save.

With this PR, the Qt save file dialog is set to use a `*.py` filter which handles the auto-extension and name collision checking for us. For consistency with the existing behaviour, it also supports a `*.*` filter: not sure that's the best idea but it is definitely something to be considered under a slightly different scope. For that I logged #802.


ntoll:
As always, many thanks @tmontes

tmontes:
> As always, many thanks @tmontes

...nothing's stopping you today: I like that! :)

Thanks.

PR #804, opened 2019-03-30
Fix restore session tab order

### LINKED ISSUE #754: Random editor tab order on re-launch
## Environment and Version

* Windows 10 1803 and macOS 10.12.6.
* Packaged Mu 1.0.2 from https://codewith.mu/en/download.

## Observed behaviour

* On relaunch Mu re-opens the previously open tabs in what seems like a random order.

## Expected behaviour

* Mu should re-open the previously open tabs in the same order.

## Steps to reproduce

* Close all editor tabs.
* Create a new file with a single statement `print("a")`, save it as `a.py`. Keep the tab open.
* Repeat the previous step with two more files, saving them as `b.py` and `c.py`.
* You should have three editor tabs, from left to right: `a.py`, `b.py`, and `c.py`.
* Quit Mu.
* Relaunch Mu.
* Sometimes tabs come out in the right order, most often they do not.

## Wrapping notes

* I haven't investigated further.
* Feels like a synchronisation issue between the driving code and the underlying GUI code in Qt. Is it?

tmontes:
Fixes #754.

Happily an easy fix, no synchronisation issues between Mu and Qt, as I have wondered back then. :)


ntoll:
@tmontes hurrah... this really annoyed me (and I wrote it!), so thank you for fixing it for us all. :-)

Just in time for EuroPython. :cow: :heart: :eu: :snake:

PR #805, opened 2019-03-31
Fixed capability to get files from esp8266 boards

HALtheWise:
In the 1.1.0-alpha release, attempting to drag a file from an esp8266 board into the local filesystem results in an "unable to import microbit" error, and fails. This fixes that bug by altering the microfs library to use `sys.stdout` instead of `microbit.uart` for transmitting data.

This change should not be merged until someone with a BBC micro:bit verifies that it does not break this functionality on that board.

I'm not entirely happy with altering a file in mu/contrib, and would love to hear suggestions about how we could better make this change (perhaps contributing it upstream?)


carlosperate:
Thanks for the PR @HALtheWise.
Unfortunately `sys.stdout` is not available in the micro:bit boards, I've created this PR in the microFS repository with an implementation that should cover most cases: https://github.com/ntoll/microfs/pull/14/
It's be great if you could try that and/or review the PR as well.

ntoll:
@HALtheWise thank you for your contribution. For this to be merged the code will need to meet our style guidelines and have 100% test coverage. You'll be able to see where these niggles are by running "make check". Thank you.

carlosperate:
This PR originally contained a similar change than https://github.com/mu-editor/mu/pull/850.
However, because this is the master branch from the https://github.com/HALtheWise/mu fork, maybe the rest of the commits after 0fcec3c were not intended for this PR?

@HALtheWise is the current state of this PR something you wish to submit for review?

carlosperate:
Thank you for your contribution @HALtheWise.

The original changes for this PR have been covered in #850, and as the PR is pointing to your fork master branch it now contains other commits outside of the original scope. In this case I think we can safely close this PR, but if there are any other bug fixes or features you'd like to contribute to Mu please do send another PR.

Thanks again!

HALtheWise:
Thank you, sorry for neglecting this review.

carlosperate:
No worries @HALtheWise, we really appreciate the effort to submit this PR with your contribution and thanks to it we were able to prepare https://github.com/ntoll/microfs/pull/14 and #850, so thanks again!

Issue #812, opened 2019-04-09
requirements_pi.txt does not exist


robmarkcole:
The dev [setup guide](https://mu.readthedocs.io/en/latest/setup.html) for pi references `requirements_pi.txt` which does not exist

carlosperate:
After PR https://github.com/mu-editor/mu/pull/680 `requirements_pi.txt` is no longer needed, as environment markers are used in the normal `requirements.txt` and `setup.py`.

We should update this line from the docs:
https://github.com/mu-editor/mu/blob/0ce00a74a691290dc381f4ca6d67a9fb28a1b763/docs/setup.rst#L89

carlosperate:
PR https://github.com/mu-editor/mu/pull/813 raised, thanks @robmarkcole!

ntoll:
Fix merged.

PR #813, opened 2019-04-10
Update dev docs to use normal requirements.txt on Pi.

carlosperate:
Tiny docs update from https://github.com/mu-editor/mu/issues/812.


ntoll:
Thanks @carlosperate, you're a :star:

Issue #797, opened 2019-03-24
Serial Console doesn't handle unicode characters properly


olivier-boesch:
If you are reporting a bug, we would like to know:

* What you were trying to do,
display text send by circuitpython script on serial console in adafruit mode.
* What steps you took to make this happen,
open serial console only
* What you expected to happen,
when i write '°' or '\u00b0' in a print statement in my script, i expect '°'
I see this if I connect with putty.
* What actually happened,
I see 'Â°' in the serial console
* Why this difference is problematic (it may not be a bug!),
unicode characters are not handled properly
* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 
installed via installer(64bits version), windows 10, mu version 1.0.2

However that's a good software. Thanks.

carlosperate:
Thanks for the report @olivier-boesch.
I can also replicate this on the micro:bit mode.

Quick notes:
- `°` in UTF-8 is 0xC2 0xB0 and in UTF-16 is 0x00B0
    - https://www.fileformat.info/info/unicode/char/00c2/index.htm
- `Â` in UTF-8 is 0xC3 0x82  and in UTF-16 is 0x00C2
    - http://www.fileformat.info/info/unicode/char/b0/index.htm
- So it looks like each UTF-8 byte sent by MicroPython is being decoded as a UTF-16 character.

carlosperate:
As this was one of the older issues for this problem and there was already a few internal and external github issues linking here I've decided to unify all duplicates we have into this one.

I've also updated the issue title to be more generic, as this affects all serial terminals (CircuitPython, micro:bit and MicroPython).

carlosperate:
We currently have two PRs looking into this:
- https://github.com/mu-editor/mu/pull/1044
- https://github.com/mu-editor/mu/pull/817

Both need to be expanded to deal with incomplete multi-byte unicode characters at the beginning and end of the data array. I've looked into this, but didn't quite finished it up, I'll try to push my current status before the end of this week (can't today as it's my birthday 🥳 ).

I've also run a benchmark on a couple of different implementations to see what would be faster. Mu currently struggles to process serial data coming without interruption at 115200 baud, so I didn't want to make this worse. Good news is a version based on @k0d's implementation is actually faster than the original implementation before any fix, as it decodes the entire data array at the beginning (otherwise it does decoding on smaller chunks when trying to regex VT100 commands).

[...]

PR #817, opened 2019-04-16
Decode incoming REPL data as UTF-8.

carlosperate:
Looking into issue https://github.com/mu-editor/mu/issues/797, this is still a WIP (no tests updated, and I'm not sure this is the best approach).

Basically the main problem is that MicroPython will encode all text in UTF-8, which can use up to 4 bytes per character, and then the REPL pane reads each byte from the serial port and decodes it using `chr()`:

https://github.com/mu-editor/mu/blob/f67852fa993f2c2749977422e8daea32f93ecf04/mu/interface/panes.py#L271-L274

For example, the character mentioned in the issue `°` is formed by 2 UTF-8 bytes [0xC2, 0xB0], and if we run `chr()` on each character in Python3 we get the following, which explains the results observed by the user:

```
>>> chr(0xC2)
'Â'
>>> chr(0xB0)
'°'
```

I could not find any information in the documentation that specifically said what encoding `chr()` and `odr()` are using. I would have assumed UTF-8, but it looks like UTF-16 to me, as `Â'` in UTF-16 is `0x00C2` and `°` is `0x00B0`.
- https://docs.python.org/3.6/library/functions.html
- https://docs.python.org/3/howto/unicode.html
- https://www.fileformat.info/info/unicode/char/00c2/index.htm
- http://www.fileformat.info/info/unicode/char/b0/index.htm

For example, character `뿅` in UTF-16 is `0xBFC5` and a quick test in the Python 3 REPL shows:

```
>>> chr(0xBFC5)
'뿅'
```
- http://www.fileformat.info/info/unicode/char/bfc5/index.htm

So just to give this a general test I've added a loop in that portion of the REPL pane code to try to decode the incoming data as UTF-8 and it seems to work.

Simple test, in Python3 REPL:

```
>>> print('°')
°
>>> print(bytes([0xC2,0xB0]))
b'\xc2\xb0'
>>> print('\xC2\xB0')
Â°
>>> print('\uC2B0')
슰
```

And I get the same if I flash the code in a micro:bit and open the Mu REPL:

```
°
b'\xc2\xb0'
Â°
슰
```

This was really a proof of concept, because I honestly don't know if there is a better approach, having to run all that code on each serial byte received sounds very inefficient. I'm hoping I can get some feedback, as there could be a better way to solve this issue.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -271,7 +271,25 @@ class MicroPythonREPLPane(QTextEdit):
             else:
                 tc.deleteChar()
                 self.setTextCursor(tc)
-                self.insertPlainText(chr(data[i]))
+                # Received data is UTF-8 encoded, with up to 4 bytes per char
+                utf8_extra_bytes = 0
+                while utf8_extra_bytes <= 3:
+                    if len(data) > i + utf8_extra_bytes:
+                        utf8_bytes = bytes(data[i:i + 1 + utf8_extra_bytes])
+                        try:
+                            decoded_data = utf8_bytes.decode('utf-8')
+                        except UnicodeDecodeError:
+                            # Probably don't have all the character bytes yet
+                            utf8_extra_bytes += 1
+                        else:
+                            self.insertPlainText(decoded_data)
+                            i += utf8_extra_bytes
+                            break
+                else:
+                    # As a fallback we can decode this byte using chr(), but
+                    # in the meantime, to easily spot issues, throw exception
+                    # self.insertPlainText(chr(data[i]))
+                    raise Exception("CARLOS CAN'T UTF-8")
             i += 1
         self.ensureCursorVisible()
 


carlosperate:
@tjguk Since you've done a lot of UTF-8 work already I was wondering if you had any advice for good UTF-8 test inputs?

ntoll:
@carlosperate @tjguk care to look into making this not-a-spike..? (i.e. contains tests and appropriate UTF8 example cases). Thank you (as always).

ntoll:
OK... I'll try to "unspike" and merge this (add some tests as best I can).

carlosperate:
Thanks Nicholas! If you prefer to prioritise your time on something else I can write some tests for this PR after PyCon UK. My main question was about the current implementation, as it's a very "manual" way to loop through every incoming character and trying to decode it, which feels a bit inefficient if we have a high baud rate configured .

ntoll:
@carlosperate ack WRT tests and inefficiency. If you have time to look at this during PyCon UK please do. I'll wait until next week to merge anything that comes of it. Have fun..! :-)

carlosperate:
Superseded by https://github.com/mu-editor/mu/pull/1026.

PR #824, opened 2019-05-03
Update setup.py

munocat:
changed installed requirements from == to >= for PyQt5, qscintilla and PyQtChart. this allows new versions to be used. Tested

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,9 +11,9 @@ with open('CHANGES.rst') as f:
     changes = f.read()
 
 install_requires = ['pycodestyle==2.4.0', 'pyflakes==2.0.0',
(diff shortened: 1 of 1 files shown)


carlosperate:
Thanks for the PR @munocat!
One reason to have the versions locked with `==` is that with `>=` you can end up with unmatched versions of PyQt packages, and that would install correctly but throw an exception on runtime.

Is there any specific reason you'd like to use newer versions?

ntoll:
Hi @munocat, thank you for the contribution. Basically, what @carlosperate says. We have, in fact, been bitten by this problem before, hence our use of `==`. However, when we do a new stable release, the pinned requirements will be updated to the latest/greatest version of PyQt that is stable across all platforms Mu supports (Windows, Mac, Linux).

PR #827, opened 2019-05-07
added parameter to suppress auto-save error messages

seantibor:
My students keep encountering a strange loop where the editor attempts to auto-save open tabs to paths without permissions. E.g. Macintosh HD or /. They can't change the path, since the timeout is set to 5 seconds by default and the Save As.. dialog box is closed by the error message.

This change adds a show_error_messages parameter to the save_tab_to_file method. The value defaults to True and will preserve existing behavior. Setting this parameter to False causes the error message to be suppressed and keeps the tab marked as modified.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -905,7 +905,7 @@ class Editor:
                 logger.error('Could not get path for {}: {}'.format(p, ex))
         return result
 
-    def save_tab_to_file(self, tab):
+    def save_tab_to_file(self, tab, show_error_messages=True):
         """
         Given a tab, will attempt to save the script in the tab to the path
         associated with the tab. If there's a problem this will be logged and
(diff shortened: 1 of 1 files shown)


ntoll:
Hey @seantibor... this is totally awesome.

A quick question... did you `make check`..? This will run the test suite... and we have 100% test coverage in our unit tests. Happy to guide you through the process of adding a test for this update if need be. Please don't hesitate to ask! More than happy to help in any way I can. :-)

seantibor:
Yep. I ran it and got 100% pass on the test suite. I originally wanted to add a side effect PermissionError to the mock for the save_tab_to_file but I couldn’t figure out how to do that properly. Then thinking about it further, it seems like the test for save_tab_to_file method should be extended to include cases for both showing and hiding the message. Could you help me do that?

ntoll:
Just about to get on a plane. Depending on how the flight is going, I'll see if I can hack something WRT the unit tests for you. ;-)

ntoll:
@seantibor Hurrah... this'll be in the alpha.2 release. Thank your for your contribution. Teachers coding on tools used for teaching... THAT'S THE BUSINESS. :cow: :heart: :mortar_board:

Issue #822, opened 2019-04-30
Error when installing third-party library - mu 1.1.0 alpha - commit f67852f


abixadamj:
When I want to install my own library: https://pypi.org/project/PyTechBrain/
I got an error:

Collecting pytechbrain
  Downloading https://files.pythonhosted.org/packages/97/a1/4fb5521501dd2d02f5c06f9c0d40e81d53d49a1f8480dad7c7beabbf1bcd/PyTechBrain-0.7.0-py3-none-any.whl
Collecting pymata-aio (from pytechbrain)
  Downloading https://files.pythonhosted.org/packages/c2/38/34346ae97407612ca49091cda062f9fa271c60e80f2cf601c37ed925330d/pymata_aio-2.28-py2.py3-none-any.whl (56kB)
Collecting pyserial (from pymata-aio->pytechbrain)
  Downloading https://files.pythonhosted.org/packages/0d/e4/2a744dd9e3be04a0c0907414e2a01a7c88bb3915cbe3c8cc06e209f59c30/pyserial-3.4-py2.py3-none-any.whl (193kB)
Collecting websockets (from pymata-aio->pytechbrain)
  Downloading https://files.pythonhosted.org/packages/43/71/8bfa882b9c502c36e5c9ef6732969533670d2b039cbf95a82ced8f762b80/websockets-7.0-cp36-cp36m-manylinux1_x86_64.whl (63kB)
Installing collected packages: pyserial, websockets, pymata-aio, pytechbrain
Exception:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/lib/python3/dist-packages/pip/commands/install.py", line 360, in run
    prefix=options.prefix_path,
  File "/usr/lib/python3/dist-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/usr/lib/python3/dist-packages/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/usr/lib/python3/dist-packages/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/usr/lib/python3/dist-packages/pip/wheel.py", line 247, in move_wheel_files
    prefix=prefix,
  File "/usr/lib/python3/dist-packages/pip/locations.py", line 153, in distutils_scheme
    i.finalize_options()
[...]

PR #836, opened 2019-05-11
Adding --system option to pip (Linux Mint problem)

abixadamj:
There is a problem with adding local libraries for user on Linux Mint - I have created issue: https://github.com/mu-editor/mu/issues/822 - this is a solution.


mu/__init__.py
```diff
@@ -8,7 +8,7 @@
 localedir = os.path.abspath(os.path.join(os.path.dirname(__file__), 'locale'))
 language_code = QLocale.system().name()
 # DEBUG/TRANSLATE: override the language code here (e.g. to Chinese).
-# language_code = 'zh'
+# language_code = 'pl'
```
carlosperate: To reduce the diff it would be good to revert this change.

mu/interface/dialogs.py
```diff
@@ -447,7 +447,7 @@ def run_pip(self):
         text area.
         """
         package = self.to_add.pop()
-        args = ['-m', 'pip', 'install', package, '--target',
+        args = ['-m', 'pip', 'install', package, '--system', '--target',
```
carlosperate: Based on https://github.com/pypa/pip/issues/3826 it sounds like this is a debian specific flag patched into pip, is that right?
If that is the case and we run this in any other platform this is likely going to return an error. In macOS:
```
$ pip install requests --system
Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --system
```

mu/locale/pl/LC_MESSAGES/mu.po
```diff
@@ -5,29 +5,26 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: \n"
-"POT-Creation-Date: 2018-09-07 18:34+CEST\n"
-"PO-Revision-Date: 2018-09-11 18:23+0200\n"
+"POT-Creation-Date: 2019-04-30 19:11+CEST\n"
```
carlosperate: This looks like an unrelated language patch, could you submit it in a different PR?

mu/interface/dialogs.py
abixadamj: Look at: 
https://github.com/mu-editor/mu/issues/822

maybe...
'''
f = open("/etc/lsb-release",'r')

f.readlines()


['DISTRIB_ID=LinuxMint\n',
'DISTRIB_RELEASE=19\n',
'DISTRIB_CODENAME=tara\n',
'DISTRIB_DESCRIPTION="FREE_DESKTOP_2019.004c"\n']
'''
and check DISTRIB_ID against 'Debian/Ubuntu/LinuxMint/....' ?
And then add --system flag ?
I can make function like debian_detect() which returns True if it will be Debian or other... and then we could add --system to pip. I know, this will be "a hack', but I cannot see any other way to make it works.

mu/__init__.py
abixadamj: I'am trying, I'm not a git master ;-)

abixadamj: I have made revert - https://github.com/mu-editor/mu/commit/2eaba0f5bed63a0bb7ab0528c593bb337ac81b12
But I do not know what to do next. When I do pull request I see: 
image

carlosperate: That's good, the final diff is only the new flag stuff, so we can always merge and squash to have compress all the final changes into a single commit.

mu/interface/dialogs.py
carlosperate: But do we understand the original problem? Perhaps there is a different workaround or fix.

The issue linked inside #822 is about mixing the `target` and `user' flags in pip, which is not being done in this file.  Is it possible the `user` flag is being added somewhere else?

abixadamj: As I investigated the problem, in Debian's based distros the only way to install libs using pip in users directory (via `--target`) is when you use `--system` option. And the only place I found in source is : @@ -447,7 +447,7 @@ def run_pip(self): - so i decided to place my `--system` there.

carlosperate: But the `--system` flag has been added by debian to pip as a patch to install the package in the "system python directory", which is the opposite of what Mu should be doing, as it needs to install the package in a special location.

Based on your error output in #822 it looks like the problem here _could be_ that somehow the final command executed includes both the flags `user` and `target`:

```
File "/usr/lib/python3.6/distutils/command/install.py", line 274, in finalize_options
raise DistutilsOptionError("can't combine user with prefix, "
distutils.errors.DistutilsOptionError: can't combine user with prefix, exec_prefix/home, or install_(plat)base
```

In this case  the `user` flag will try to install the package in the user home directory (not good), and the `target` flag will try to install it in the location indicated by Mu (good, but contradictory to the other flag). Because of this, the `system` flag "fixes" the problem by "cancelling out" the `user` flag.

However, if this is the case, where is the `user` flag coming from? I think that's the thing we need to figure out.

Saziba: Hi, I manage to find a pip.conf file inside .pip folder in my home directory. It has a parameter "user=true" in it. When I changed this parameter to false, I could finally use the -t option on pip.

carlosperate: Good find @Saziba! 
Were you installing packages with the default Mu installation, or did you edit the source code as well?

carlosperate:
Because there has been wide changes to the user python environment (using virtual environments for running the user code with Mu), I believe this will no longer be a problem, as now the user venv should not be using the system pip.

Thank you very much for working through this issue and all the stellar debugging!

PR #844, opened 2019-05-28
ESP-mode: Show current directory in files pane to support multiple projects.

dybber:
This is a fix that allows users to work on multiple projects in the ESP-mode. Currently, the files pane always show the content of the mu_code directory, however, every project must have main.py files and subdirectories are thus necessary to separate projects.

The simple fix in this patch is to show the directory of the file in the current tab, if the current tab is not associated with a file the workspace_dir (i.e. the mu_code directory by default) will be used.

The patch should be seen as a perhaps temporary fix to address the situation, a longer term fix would be to support directories directly in the file manager.

### DIFF
--- a/mu/modes/esp.py
+++ b/mu/modes/esp.py
@@ -21,6 +21,7 @@ from mu.modes.base import MicroPythonMode, FileManager
 from mu.modes.api import ESP_APIS, SHARED_APIS
 from mu.interface.panes import CHARTS
 from PyQt5.QtCore import QThread
+import os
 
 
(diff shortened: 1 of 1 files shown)


ntoll:
@dybber I'm going to fix the failing test suite, and merge this PR.

Please, please, please remember to `make check` before creating a PR to ensure the test suite passes and we retain 100% test coverage. Thank you!

ntoll:
OK... done and merged. Thank you! :-)

PR #848, opened 2019-05-31
Using CTRL-scroll with a mouse, controls built-in zoom functionality

### LINKED ISSUE #846: Scroll wheel doesn't zoom text
Without the ability to zoom in and out with mouse scroll wheel this otherwise promising editor is of no use for me. I am a teacher and I frequently adjust zoom level when explaining various parts of code to students.

ntoll:
Fixes #846.

This needs testing on:

 * [X] Linux
 * [ ] Windows 10
 * [ ] Windows 7
 * [ ] OSX
 * [ ] Raspberry Pi


tjguk:
Well it's working for me on Windows. (But it was working anyway :) )

carlosperate:
Tested and working in macOS Mojave :)

> Well it's working for me on Windows. (But it was working anyway :) )

It'd be good to check it in Python mode with the REPL opened, that way you can see if the event is trapped and Mu's internal zoom mechanism is used.
You should also see a less granular zoom activity, or at least in macOS the zoom level jumps are bigger with this PR.

carlosperate:
Actually in macOS the Mu zoom works with the `Command` key (as specified by `Qt.ControlModifier`), but the `Control` key will still change the Qscintilla zoom bypassing Mu's internal mechanism (this was the same behaviour as before, basically this PR adds zoom with the `Command` key).

ntoll:
@carlosperate ahh... OK. So what is the expected / usual behaviour for this sort of thing on OSX. We should basically do that. ;-)

carlosperate:
> So what is the expected / usual behaviour for this sort of thing on OSX

To be honest, I don't know, I'm a fairly new mac user (only for about a year or so). In most cases what's normally `Control` in Linux and Windows is usually triggered with `Command` on macOS, and in that case Mu is doing the right thing here (`Command`) and Qscintilla is not (`Control`).

However, I couldn't get the the zoom with the scroll-wheel to work with either key in SublimeText, VSCode or Chome, but in Firefox it worked the same with both keys.

tmontes:
My observations on macOS 10.12 Sierra, which took me a while to grasp:
* CTRL+Mouse Wheel behaves as before: very wide range of text sizes, not based on Mu's zoom code, probably a QScintilla's trick.
* CMD+Mouse Wheel brings in new behaviour: shorter range of text sizes based on Mu's zoom code.

Qt mapping `Qt.ControlModifier` to the CMD key on Macs makes sense, given that CMD is the "standard" modifier key used, for example, for copy/paste with CMD-C/V, etc.

This PR makes it possible for users to zoom in/out in two different ways, getting two similar but slightly different behaviours. I'd say the inconsistency would ideally be avoided.

Possible ways forward:
* Just let it be the way it is.
* The CTRL key on macOS is mapped from `Qt.MetaModifier`, AFAICT: include that one in the mouse wheel handler condition.
* Disable QScintilla's built-in zoom behaviour.

ntoll:
@tmontes I believe disabling QScintilla's zoom behaviour is probably the simplest and most effective way to go.

tmontes:
@ntoll: agreed.

Wondering:
* If disabling it is possible.
* If doing that won't lead to any UI/UX regressions.
* If this PR actually fixes @senkyr's issue. :)

ntoll:
Quite... :-) It's late Friday afternoon, and this can wait until Monday to find out if it's possible. I suspect we just need to override QScintilla's wheelEvent handler to do nothing.

tmontes:
One more thing, after testing with two different mice:
* Each click of the mouse wheel is triggering the `wheelEvent` method twice in my system.
* This results in 0, 2, 4, or 6 as possible `zoom_position` values, instead of the whole 0-6 range which can be set with the menu bar buttons / keyboard shortcut.

Can the others please confirm this is macOS 10.12 Sierra only or if there's something cross-platform going here? :)

carlosperate:
> 10.14 Mojave also exhibits this behaviour, @carlosperate?]

Same in Mojave.

tmontes:
@ntoll, here's something I just tried out which seems to be effectively disable QScintilla's built-in zoom handling. Adding the following method to the `EditorPane` class, in `mu/interface/editor.py`:

```
    def wheelEvent(self, event):                                                          
        event.ignore()
```

Wondering:
* Could there be a simpler way, maybe at initialisation time? (argh, where are QScintilla's docs?)
* If this PR actually fixes @senkyr's issue! :)

carlosperate:
If the `wheelEvent` is ignored in the `EditorPane`, does scrolling still work?

tmontes:
> If the `wheelEvent` is ignored in the `EditorPane`, does scrolling still work?

Good catch! It does not... (shame on me!) :)

tmontes:
...maybe something along these lines, then:

```
    def wheelEvent(self, event):
        if QApplication.keyboardModifiers():                                              
            event.ignore()
        else:
            super().wheelEvent(event)
```

Not sure if the `event.ignore` is important / required. If not, it could be trimmed down to:

```
    def wheelEvent(self, event):
        if not QApplication.keyboardModifiers():
            super().wheelEvent(event)                                                     
```

Still wondering if any of this helps addressing the original issue @senkyr is facing. :)

ntoll:
OK... I'm going to nail this one down and try to merge it once I've addressed all the comments above. ;-)

ntoll:
Hurrah... Appveyor finally works! Merging now... let's see how folks get on with this in the next Alpha.

Issue #778, opened 2019-02-27
Docs: translations.rst is outdated.


tmontes:
While reviewing the documentation for a PR I've been working on to address #776, I found out that `mu/docs/translations.rst` hasn't been updated after merging #764 (my bad!), so it's not 100% accurate.

Leaving this here as a reminder for improvement which I expect to address soon.

tmontes:
This should have been closed a while ago, when #849 was merged!
Closing now.

PR #849, opened 2019-06-01
Translation doc: how to run with a specific locale/language.

tmontes:
This should have been included in #764 and was due as noted in #778. :)


ntoll:
Thank you @tmontes. Both you and @carlosperate have had a busy Saturday afternoon.

Thank you! :-)

Issue #851, opened 2019-06-03
Connecting to ESP32 on 64-bit Windows 10


cgarmstr:
I've been trying to connect to an ESP32-WROVER-KIT from a Windows 10 machine.

I can connect ok to the REPL with TeraTerm on Port COM5, but nothing happens with Mu. Do I need to specify serial port settings somewhere?
![image](https://user-images.githubusercontent.com/7409913/58835444-e8550300-864d-11e9-9782-974ff1a423a1.png)

The log file is below.

Many thanks

Cecil

PS I saw the post about an upcoming fix to ESP32 mode. Is there a way of getting informed when the next alpha is released?
---------------------------------------------------------------
2019-06-03 22:08:36,791 - root:112(run) INFO: 

-----------------

Starting Mu 1.1.0.alpha.1
2019-06-03 22:08:36,793 - root:113(run) INFO: uname_result(system='Windows', node='OPTIPLEX-7040', release='10', version='10.0.17763', machine='AMD64', processor='Intel64 Family 6 Model 94 Stepping 3, GenuineIntel')
2019-06-03 22:08:36,793 - root:114(run) INFO: Python path: ['C:\\Users\\cecil\\AppData\\Local\\Mu\\pkgs', 'C:\\Users\\cecil\\AppData\\Local\\Mu', 'C:\\Users\\cecil\\AppData\\Local\\Mu\\Python\\python36.zip', 'C:\\Users\\cecil\\AppData\\Local\\Mu\\Python\\DLLs', 'C:\\Users\\cecil\\AppData\\Local\\Mu\\Python\\lib', 'C:\\Users\\cecil\\AppData\\Local\\Mu\\Python', 'C:\\Users\\cecil\\AppData\\Local\\python\\mu\\site-packages', 'C:\\Users\\cecil\\AppData\\Local\\Mu\\pkgs\\IPython\\extensions']
2019-06-03 22:08:36,793 - root:115(run) INFO: Language code: en_GB
2019-06-03 22:08:37,081 - mu.logic:574(__init__) INFO: Setting up editor.
2019-06-03 22:08:37,082 - mu.logic:596(__init__) INFO: Settings path: C:\Users\cecil\AppData\Local\python\mu\settings.json
2019-06-03 22:08:37,082 - mu.logic:597(__init__) INFO: Session path: C:\Users\cecil\AppData\Local\python\mu\session.json
[...]

PR #855, opened 2019-06-09
ESP mode: Add VID/PID for other FTDI USB-to-UART chips.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/851.

@cgarmstr could you check if this branch works for you?

### DIFF
--- a/mu/modes/esp.py
+++ b/mu/modes/esp.py
@@ -37,12 +37,13 @@ class ESPMode(MicroPythonMode):
 
     # There are many boards which use ESP microcontrollers but they often use
     # the same USB / serial chips (which actually define the Vendor ID and
-    # Product ID for the connected devices.
+    # Product ID for the connected devices).
     valid_boards = [
         # VID  , PID
         (0x1A86, 0x7523),  # HL-340
         (0x10C4, 0xEA60),  # CP210x
-        (0x0403, 0x6015),   # Sparkfun ESP32 VID, PID
+        (0x0403, 0x6010),  # FT2232C/D/L/HL/Q (ESP-WROVER-KIT, ESP32DemoBoard)
+        (0x0403, 0x6015),  # FT X-Series (Sparkfun ESP32)
     ]
 
     def actions(self):


cgarmstr:
Apologies. I followed the guidelines for Developer Setup https://mu.readthedocs.io/en/latest/setup.html but after doing python run.py got 
ModuleNotFoundError: No module named 'appdirs'
pip install appdirs
fixed that but then got
    from PyQt5.Qsci import QsciScintilla, QsciLexerPython, QsciAPIs
ModuleNotFoundError: No module named 'PyQt5.Qsci'

PS I cheated and put the extra line
        (0x0403, 0x6010),  # FT2232C/D/L/HL/Q (ESP-WROVER-KIT, ESP32DemoBoard)
in esp.py in my existing mu-editor_1.1.0-alpha.1_win64.exe installation.
Mu appears doesn't give a no device connected error, but there is no response to the REPL.

carlosperate:
Okay, let's take this back to issue https://github.com/mu-editor/mu/issues/851 from comment https://github.com/mu-editor/mu/issues/851#issuecomment-500518733.

ntoll:
@carlosperate is this safe to merge..?

carlosperate:
It's not "unsafe", but because the FTDI serial-to-USB bridge creates more than one COM port (one for "debugging" and another for the serial comms), and the first port that Mu uses to attempt a serial connection is not the right one, the user still cannot use the ESP device even with this PR.

The real solution is probably to do as described in #851, and offer a way for the user to select between multiple COM ports.

So we can merge this now and possibly implement that later, or we can leave this open until the other feature is implemented.

ntoll:
@carlosperate let's leave this open until we have a "complete" solution to land in master. Thanks for the clarification.

ntoll:
Any movement on this..?

carlosperate:
Not really, a full solution would be letting the user select a COM port when multiple options are available, and that would require a more significant PR with a new UI window and all.
Unfortunately I probably won't be able to commit the time to do that in the short term.

Feel free to merge, close, or leave open this PR as you see fit.

ntoll:
OK... thank you @carlosperate -- I leave it open for the time being in case others want to work on it. If it's not "fixed" by the move to beta status for 1.1, then I'll close it.

dybber:
Hi all, I have integrated the changes here into #1094, and part of the idea about being able to select devices from a dropdown was just merged (see #1017). I think it's now at a state where we can close this. If people at some point again report issues with this new interface for device selection, we can take a look at it then.

PR #856, opened 2019-06-09
Update logger.warn() deprecated method to warning().

carlosperate:
I've been getting this deprecation warning every time I run the tests locally, so this tiny PR updates the method call.

https://docs.python.org/3/library/logging.html#logging.Logger.warning


ntoll:
@carlosperate aaaand, relax. That niggley annoyance is fixed and merged. ;-)

PR #860, opened 2019-06-17
slovak translation

bletvaska:
initial slovak translation


ntoll:
@bletvaska Ďakujem. Toto je úžasné..! (Thanks Google Translate) :cow: :heart: :slovakia:

Issue #821, opened 2019-04-29
pgzero mode Music button should be relative to code file


tjguk:
(Found with @linguini2004)

Running current master in pygame zero mode. Editing code in a folder which is not the default mu_code folder. 

Clicking the [Music] button opens mu_code/music. Copied a file in there, but when the game is run, pgzero reports that the music file cannot be found.

Without looking, my guess is that pgzero looks for the music folder either relative to the file being run or relative to cwd (which in the dev. version of Mu are the same thing). Mu probably assumes relative to mu_code. Certainly creating a music folder under the code folder and copying the file there solved the problem.

Assuming I'm right I do regard this as a bug in Mu. ie I think Mu should open a Music (and probably Images) folder relative to the path of the currently-edited file. There is some possible source of confusion in that clicking the button against different files will yield different folders. But I think that's outweighed by the fact that *not* doing that will produce a difficult-to-diagnose error message from pygame zero.

tmontes:
@tjguk, given that PR #862 has been merged, I suppose this issue can be closed, right? :)

tjguk:
Yes; thanks. (I'm lousy at back-tracking to close the issues I've PR-ed!)

tmontes:
> Yes; thanks. (I'm lousy at back-tracking to close the issues I've PR-ed!)

...who isn't? That's what computers are good at. :)

PR #862, opened 2019-06-18
Resources relative to code

tjguk:
This change addresses issue #821 which is a necessary follow-on from the fact that you can load code from arbitrary directories outside or below the `mu_code` workspace directory combined with the fact that pygamezero expects its `images/` etc. directories to be relative to the code being run. (Strictly: relative to the current working directory).

Originally mu_code was the only recognised location for code files, so the `images/` `sounds/` etc. directories were expected to be statically `<workspace>/images` etc. Now that code can be loaded from any directory, the images, sounds etc. will be in different places depending on which file is active.

Therefore the [Images] button etc. need to open a folder relative to the code file being viewed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,6 @@ target/
 
 # Movie
 *.mp4
+
+# Virtual Environment
+.venv/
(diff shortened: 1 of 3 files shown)


ntoll:
@tjguk good catch and thank you!

PR #863, opened 2019-06-21
Remove edu packages

martinohanlon:
As per #829 remove the edu packages from setup.py specifically guizero, gpiozero, pigpio


martinohanlon:
This may seem like a little thing but it would be cool to see this in the next alpha as I'm keen to test X platform consistency and using the package installer.

ntoll:
If at all possible, this WILL be in the next Alpha -- next week is a Mu focussed week. The past month (or two) have been full of all sorts of other things which have made Mu work difficult. So, yes, if at all possible, yes. ;-)

Finally, thank you for doing this... I'm pretty certain the only thing stopping this from getting merged is my finding the time to properly review it. My expectation is I'll take a look, grok what's going on and say, "looks good to me" and click merge. That's because you're a great engineer @martinohanlon, and your PRs are always of high quality.

martinohanlon:
Thanks Nic.

tjguk:
Well I just rolled my sleeves up to do an in-depth review, did a fresh clone, git-pulled @martinohanlon fork... and I can confirm that those three lines in setup.py are definitely gone for good.

*However*... the API is still present in https://github.com/mu-editor/mu/blob/master/mu/modes/api/pi.py

*However However*... I can't find the code which actually generated that file. It appears to have been created /ex nihilo/ by https://github.com/mu-editor/mu/commit/7709a26d7045a4559f8a0a1f7a4cdf9da3c3de42

So we could either go ahead and merge this PR anyway, on the basis that the presence of a few extraneous API entries won't be too much of a problem. Or we could just ditch the pi.py API module as most of the things in it are nice-to-haves.

tjguk:
I don't feel strongly about it, but my inclination would be to ditch pi.py from the api folder. If we feel the need, we'll be able to regenerate it in the future. Thoughts, @ntoll and @martinohanlon ?

martinohanlon:
I saw the same, I came to the conclusion that it was still a useful information for the user, regardless of whether the modules are installed by default (although interestingly on Raspbian they will be)...

martinohanlon:
but obvs happy to be led by you?

tjguk:
> but obvs happy to be led by you?

I'll let @ntoll make the call

ntoll:
Folks, all the API docs are generated by running the various scripts found in the `utils` directory in the root of the repository. I should document how this all works, but it's essentially scraping RTD / Sphinx generated docs and turning them into the format Scintilla understands. This is of course an "automated" manual process.

I'm with @martinohanlon for retaining them, since they'll probably be more useful than not.

I'll merge this PR right now. alpha.2 coming today or tomorrow (assuming no mishaps).

PR #864, opened 2019-06-21
First draft of a new web mode using Flask

ntoll:
This PR introduces a brand new mode for Mu, one that allows beginners to very quickly create Flask based web applications. This is very much a *first draft* to be included in the next alpha release.  I final "feature" yet to be implemented is for integration with PythonAnywhere so, assuming a username and secret API key, users can deploy their web application to PythonAnywhere at the touch of a button. The result being a URL on the "live" internet. Currently PythonAnywhere are making updates to the API which will make all this possible from their end, and once complete I'll add the necessary updates via a new PR.

This mode works in several ways:

* Upon start Mu checks to see if the default assets for a simple web application exist in the user's workspace directory. If they don't, they're copied over. This allows users to get up and running with a simple "it works!" web application without typing any code.
* Mu uses the default CSS and HTML lexers from Scintilla for such files.
* The "Run" button starts the local Flask development server.
* The "Browse" button opens the user's default browser at the root endpoint of localhost.
* The other buttons (templates, css, images) open directories to the default locations for these so they work within the Flask app.

Like I said, this is a first draft based on a spike I wrote a month or two ago. All feedback most welcome. I'd like to get this into a new Alpha release *before* EuroPython. :-)


tjguk:
I've just taken it for a quick spin (and reminded myself what a Flask application looks like!). One very initial comment is: I think it would be good to have a boilerplate "Hello World!" in place, albeit as a comment block. Something like this:

```
from flask import Flask
app = Flask(__name__)

@app.route('/')
def hello():
    return 'Hello, World!'
```

I know we don't do this for any other mode, but this seems different to me. In all the other modes we're entering a general-purpose workspace, even if it's narrowed to micro:bit or pygame. Here, you can't do anything without the basic boilerplate I've outlined above.

Now I think about it, it probably makes some sense for the pgzero mode to have a boilerplate as well:

```
def update():
  ...

def draw():
  ...
```

tjguk:
On Windows when I load a template there's an odd mojibake thing going on with the lines of dashes. I'll look into it as it should detect it anyway, but it might be worth using POA (Plain Old ASCII) to avoid the issue in the first place.

tjguk:
The CSS styler has an odd mismatch between the sizes of the  CSS itself and the comment blocks

image

I've seen that before on a Scintilla control: it mostly reflects the fact that Scintilla has far too many "types" of text and you have to style them all consistently!

tjguk:
... and I've just noticed that there _is_ a code template. And when I run it now, the code template appears for a new file. I don't remember what I did the previous time; it's just possible that I had an existing file when I switched to the Web mode and then just filled in the gaps. So I think you can scratch that issue.

tjguk:
When a file is opened which is not a .py file, the logic.py:_load function hands off to the modes (in turn, apparently, regardless of the current mode). The web mode open_file() function just does a unadorned open() without invoking, eg, the sniff_encoding() code from logic.py.

On Windows, at least, this gives the result that the template is loaded via a non-utf8 codec (either ASCII or cp1252 probably)

setup.py
```diff
@@ -39,6 +39,7 @@
     'semver>=2.8.0',
     'nudatus>=0.0.3',
     'black>=18.9b0;python_version > "3.5"',
+    'Flask==1.0.3',
```
carlosperate: Might be easier for debian/raspbian packaging (less patches needed) if the version is 1.0.2:
https://packages.ubuntu.com/disco/python3-flask
https://packages.debian.org/buster/python3-flask

mu/modes/web.py
```diff
@@ -0,0 +1,241 @@
...
+        new files into the opened folder.
+        """
+        img_dir = os.path.join(self.workspace_dir(), 'static', 'img')
+        logger.info(img_dir)
+        self.view.open_directory_from_os(img_dir)
+
+    def browse(self, event):
+        """
+        Open the default browser to http://127.0.0.1:5080/
+        """
```
carlosperate: Should browse warn the user if the server is not running?

carlosperate:
> The CSS styler has an odd mismatch between the sizes of the CSS itself and the comment blocks
>
> image
>
> I've seen that before on a Scintilla control: it mostly reflects the fact that Scintilla has far too many "types" of text and you have to style them all consistently!

As a general question, when did the code background change to the eggshell/yelowish colour? It clashes a bit with the white background colour in this example.

setup.py
ntoll: :+1:

mu/modes/web.py
ntoll: Good catch. :-)

ntoll:
I'm not sure the code background *has* changed. As far as I can tell, it's always beer `#FEFEF7` (set in the Font class in `themes.py`.

carlosperate:
Looks like it was changed in PR https://github.com/mu-editor/mu/pull/644 about 10 months ago, but for a very good reason.
Should the CSS comments background colour be changed as well?

ntoll:
@carlosperate the problem (bug) with the QsciLexerCSS is that it doesn't appear to correctly style comments. I'm (right now) trying to figure out how I might fix this. It's really bugging me.

ntoll:
@carlosperate I've investigated the CSS comment rendering problem this morning... and I'm at a dead end. I've sent a bug report to the QScintilla mailing list (the preferred way to report this sort of thing), so let's see what they come back with.

setup.py
ntoll: :+1:

mu/modes/web.py
ntoll: Done.

ntoll:
Folks, once the tests pass, I'd like to merge this in for the next alpha release in time for EuroPython. We can fix/refine given user feedback.

Issue #835, opened 2019-05-10
Rename Adafruit mode to CircuitPython mode


dhalbert:
Now that CircuitPython is supported on a number of non-Adafruit devices (see [circuitpython.org](https://circuitpython.org)), and non-Adafruit boards have been added to Mu, we're thinking it would be appropriate to rename the mode to "CircuitPython". A CircuitPython device has a CIRCUITPY drive and support CircuitPython-specific modules (`board`, `busio`, etc.).

The mode description
```
Adafruit CircuitPython
     Use CircuitPython on Adafruit's line of boards.
```
would change to something like
```
CircuitPython
     Use CircuitPython on Adafruit, MeowCat, Particle, and other CircuitPython boards.
```

I'm planning to do a PR for this soon, but someone else can too. Mostly it just involves some string changes, and renaming the `adafruit.py` files in the source tree. There are also some images to update. I will also remove the specific Adafruit boards listed in `mu/modes/adafruit.py`, since they are covered by the general `(0x239A, None),    # Any Adafruit Boards` entry.

ntoll:
Should the logo be replaced too? (From the Adafruit corporate logo to Blinka [please suggest an appropriate picture of Blinka since there are several I could choose, and I'd like to reflect your [CircuitPython's] judgement on this]).

bwshockley:
I worked this a bit with PR #868 - I think I got the changes implemented, but I wasn't sure of the VID & PIDs for other boards.

ntoll:
Fixed in #869.

PR #867, opened 2019-06-25
Add Mini SAM board to the mode list.

bwshockley:
I've included files to add a new mode for Mini SAM boards - these are based off Adafruit CitcuitPython.  Additional information can be found here: minisam.cc or on my github repositories.


dhalbert:
How about having all CPy boards be under a general CircuitPython mode: #835, which I have not yet had time to do?

bwshockley:
That would be a better option for me I think - just didn't want to add more work.

PR #870, opened 2019-06-28
Added another ESP32 board.

KKawase0104:
Added USB VID and PID of "Studuino:bit" which is micon board having ESP32.


mu/modes/esp.py
```diff
@@ -43,6 +43,7 @@ class ESPMode(MicroPythonMode):
         (0x1A86, 0x7523),  # HL-340
         (0x10C4, 0xEA60),  # CP210x
         (0x0403, 0x6015),   # Sparkfun ESP32 VID, PID
+        (0x20A0, 0x4269),   # Studuion:bit VID, PID
```
carlosperate: Thanks for the submission @KKawase0104!
Do you have more info about where this VID/PID come from?
Is it a standard UART-to-USB bridge?

KKawase0104: Thank you for your reply.

This VID/PID is the ID for our company's "Studuino:bit" device.
Studuino:bit is a microcomputer board which has ESP32.
We want to use mu's ESP32 mode to control Studuino:bit.

It isn't a standard UART-to-USB bridge.
We're using the microcomputer as the UART-to-USB bridge.

KKawase0104: Since there was a correction in REPL ESC and CSI sequence processing, I closed it once, sorry.

PR #877, opened 2019-07-03
Use rtds logo in README.rst and PyPI description.

carlosperate:
The README file used to display the logo with a `docs/logo.png` relative URL. Then the setup.py would read the file, and replaced that string to a full URL for the PyPI description.

In commit 9997963eda1e4dc3c0a8a3d3be7cd93728aedb19 this was updated to a URL using raw.githubusercontent.com. The setup.py file was not updated, so the PyPI descriptions ends up with a malformed URL:
image

Now that Read The Docs is configured to publish on each commit, the https://mu.readthedocs.io/en/latest/_images/logo.png URL used in setup.py should reflect the logo from master (as raw.githubusercontent.com did), so we can just use that URL in the README and remove that bit from the setup.py file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 Mu - A Simple Python Code Editor 
 ================================
 
-.. image:: https://raw.githubusercontent.com/mu-editor/mu/master/docs/logo.png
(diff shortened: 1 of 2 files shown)


ntoll:
Thanks..! :-)

Issue #857, opened 2019-06-11
Fail to make package on Windows 10: ImportError: Could not find 'mu'


bigeyex:
I tried to package Mu on Windows 10, by make win32 and make win64. 

The same error message appears:

```
INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\qtconsole\4.4.3\qtconsole-4.4.3-py2.py3-none-any.whl
INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\semver\2.8.1\semver-2.8.1-py2.py3-none-any.whl
INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\setuptools\28.8.0\setuptools-28.8.0-py2.py3-none-any.whl
INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\six\1.12.0\six-1.12.0-py2.py3-none-any.whl
INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\toml\0.10.0\toml-0.10.0-py2.py3-none-any.whl
INFO:nsist.wheels:Downloading wheel: https://files.pythonhosted.org/packages/9d/76/8a52c49e0a7e2638442f379d34f55232c773d4df7f69b297e451c980f99d/tornado-6.0.2-cp36-cp36m-win_amd64.whl

INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\traitlets\4.3.2\traitlets-4.3.2-py2.py3-none-any.whl
INFO:nsist.wheels:Using cached wheel: C:\Users\makeblock\AppData\Local\pynsist\pypi\wcwidth\0.1.7\wcwidth-0.1.7-py2.py3-none-any.whl
Traceback (most recent call last):
  File "C:\Users\makeblock\workspace\mu\package\mu_nsist.py", line 192, in <module>
    main()
  File "C:\Users\makeblock\workspace\mu\package\mu_nsist.py", line 182, in main
    ec = MuInstallerBuilder(**args).run(makensis=(not options.no_makensis))
  File "C:\Users\MAKEBL~1\AppData\Local\Temp\mu-pynsist-xjqxmf_j\mu-packaging-venv\lib\site-packages\nsist\__init__.py", line 471, in run
[...]

PR #879, opened 2019-07-04
update pip when making windows binaries.

bigeyex:
Addresses issue #857 : `make win32` fails when pip is outdated in the user's machine.

### DIFF
--- a/win_installer.py
+++ b/win_installer.py
@@ -253,6 +253,10 @@ def run(bitness, repo_root):
         print('Installing mu with', venv_python)
         subprocess.run([venv_python, '-m', 'pip', 'install', repo_root])
 
+        print('Updating pip in the virtual environment', venv_python)
+        subprocess.run(
+            [venv_python, '-m', 'pip', 'install', '--upgrade', 'pip'])
+
         pynsist_cfg = os.path.join(work_dir, 'pynsist.cfg')
         print('Creating pynsist configuration file', pynsist_cfg)
         installer_exe = create_pynsist_cfg(venv_python, repo_root, pynsist_cfg)


win_installer.py
```diff
@@ -253,6 +253,10 @@ def run(bitness, repo_root):
         print('Installing mu with', venv_python)
         subprocess.run([venv_python, '-m', 'pip', 'install', repo_root])
 
+        print('Updating pip in the virtual environment', venv_python)
```
tmontes: I understand this works, but ideally `pip` would be updated right after creating the virtual environment, before the preceeding Mu installation lines.

win_installer.py
```diff
@@ -253,6 +253,10 @@ def run(bitness, repo_root):
...
+        subprocess.run(
+            [venv_python, '-m', 'pip', 'install', '--upgrade', 'pip'])
```
tmontes: For formatting consistency, I'd put the closing `)` on the next line, vertically left-aligned with `subprocess.run(`, as in, say the `shutil.copy` call a few lines down. What do you say? :)

tmontes:
@bigeyex, thanks a lot for your contribution. I left a few notes that I believe could lead to a slight improvement: conceptually and visually, if I may call them so.

Furthermore, would you be willing to include the `pip` upgrade before install in the documentation? Or would you rather leave that to someone else?

Again, thanks! :)

win_installer.py
ntoll: I'll make this minor change and then merge. :-)

win_installer.py
ntoll: +1 I'll change this over... :-)

ntoll:
@bigeyex thank you so much for your contribution. You have a perfect sense of timing since I should be able to squeeze this into the alpha.2 release which I hope to do later this evening or tomorrow morning (assuming my tests of the installers go well). :-)

ntoll:
Thanks... all merged. :+1:

Issue #881, opened 2019-07-05
1.1.0a2 | PGZero | Crash opening assets (images, sounds, etc) with fresh editor tab.


tmontes:
## Observed behaviour

When in Pygame Zero mode, with a blank/fresh/untitled editor tab, clicking any of the Images, Fonts, Sounds, or Music buttons crashes Mu.

Traceback is:
```
(mu-venv) tiago.montes (~/Work/Python/github.com/mu): mu-editor 
Traceback (most recent call last):
  File "/Users/tiago.montes/Work/Python/github.com/mu/mu/modes/pygamezero.py", line 192, in show_music
    os.path.dirname(self.view.current_tab.path), 'music')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/posixpath.py", line 156, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
(mu-venv) tiago.montes (~/Work/Python/github.com/mu): mu-editor 
Traceback (most recent call last):
  File "/Users/tiago.montes/Work/Python/github.com/mu/mu/modes/pygamezero.py", line 159, in show_images
    os.path.dirname(self.view.current_tab.path), 'images')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/posixpath.py", line 156, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

## Expected behaviour

Not really sure, given that the asset paths are relative to the current editor tab's file path. Maybe these buttons could be disabled if the editor tab hasn't been saved yet?

Prompting thought and discussion, here... :)

## Steps to reproduce

* Launch Mu 1.1.0.alpha.2.
* Change mode to "Pygame Zero".
* Create a new editor tab.
* Click the "Images" button.
* Mu crashes.

## Platform notes

* Tested on macOS 10.12.6.
* Running from source, with the `mu-editor` command, within a Python 3.6.8 virtual environment, after a `pip install -e .[dev]`.
* Given the nature of the traceback I expect such failure to occur in all platforms both in packaged and non-packaged variations of Mu.

amos1969:
Just tried on Linux Mint using Python3.7 in a Virtual Environment created using a previous version's requirements.txt with pip. 

I get exactly the same behaviour:

`(myvenv) amos@amos-ThinkPad-T430 ~/mu $ python mu/run.py    
[...]

PR #883, opened 2019-07-05
[INCOMPLETE] 881 crash opening assets

tjguk:
I'm pushing what's there now, which fixes the immediate issue. 

But it raises another: we are now opening an images / sounds etc. folder under the directory which the source code is in, that folder might well not exist. Do we:

(a) Create it automatically and then open it

(b) Default to the workspace dir as we do when there's not file open

(c) Something else (eg raise a specific dialog-box error warning of the situation)

My preference is for (a) because: from a teacher's perspective, it's the easiest thing to do; and it doesn't seem as though it would introduce any problems even if it is slightly magical.

Thoughts?

### DIFF
--- a/mu/modes/pygamezero.py
+++ b/mu/modes/pygamezero.py
@@ -155,7 +155,7 @@ class PyGameZeroMode(BaseMode):
         This should open the host OS's file system explorer so users can drag
         new files into the opened folder.
         """
-        if self.view.current_tab.path:
+        if self.view.current_tab and self.view.current_tab.path:
             base_dir = os.path.dirname(self.view.current_tab.path)
         else:
             base_dir = self.workspace_dir()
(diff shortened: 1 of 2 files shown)


tjguk:
I've added the autocreate logic (refactoring a little as I go...)

tjguk:
(Sorry, should have said: this is addressing #881)

ntoll:
:+1: to this approach. I'll update in light of the web mode also having similar "asset" directories. I'll do a "quiet" release of the updated version later today.

ntoll:
I'll move the `assets_dir` method into the base class and update the web mode.

tjguk:
> I'll move the `assets_dir` method into the base class and update the web mode.

+1

tmontes:
Oh wow! I didn't mean to rush anyone into a fix: all I wanted was to "write down" the failure so that we wouldn't forget about it. But if a fix is coming, well, thanks! :)

ntoll:
OK... so I've refactored and merged into master. I'm going to (quietly) update alpha 2 to include this fix. ;-) Just going to go through the packaging checks and so on.

Issue #892, opened 2019-07-13
EP 2019 | Improve checks


tmontes:
The current Unix `make check` implementation drives a few checks.

Of those, two of them, feel hard to maintain and read. Here's a snippet from the current `Makefile`:

```
pyflakes:
        find . \( -name _build -o -name var -o -path ./docs -o -path ./mu/contrib -o -path
 ./utils -o -path ./venv -o -path ./package \) -type d -prune -o -name '*.py' -print0 | $(
XARGS) pyflakes

pycodestyle:
        find . \( -name _build -o -name var \) -type d -prune -o -name '*.py' -print0 | $(
XARGS) -n 1 pycodestyle --repeat --exclude=package/*,build/*,docs/*,mu/contrib*,mu/modes/a
pi/*,utils/*,venv/*,.vscode/* --ignore=E731,E402,W504
```

Both are based on using `find` to determine the files that should be fed to each tool, respectively. While very powerful and expressive, I believe these can be simplified.

The `pycodestyle` check can be written as a much more direct:

```
pycodestyle --ignore=E731,E402,W504 --exclude=./mu/contrib/,./mu/modes/api mu tests
```
*[EDIT: removed `--repeat` that, according to docs, is now obsolete]*

The `pyflakes` check could be applied the same "approach" with:
```
pyflakes tests mu win_installer.py
```

~*However*, for some possibly very obvious reason that I'm failing to see now, this simpler approach generates many `undefined name '_'`, all over the place, while the existing one seems to understand that `gettext` defines `_`. **What am I missing?** :-/~

*[EDIT: Read a bit more and found the `PYFLAKES_BUILTINS=_` environment variable, set in the `Makefile`!]*

Additionally, the proposed approach to the `pycodestyle` check, makes it run much faster on my already system (~3.5 seconds vs ~10 seconds), which is also nice.

If we change these check on the `Makefile`, then we can probably apply the same approach to their Windows `make.py` counterparts.

~Would love to hear feedback on this, in particular regarding the behavioural inconsistency I'm observing...~ Shall we simplify this? Standing by, thanks!

tmontes:
Explored a bit more.

Here are my `pycodestyle` timing measurements on macOS, Ubuntu 18.04 VM, Windows 10 VM:

|             | current | proposed |
| ----------- | ------- | -------- |
| macOS       | ~10s    | ~3.5s    |   
| Ubuntu      | >150s   | ~3.0s    |
[...]

Issue #891, opened 2019-07-13
EP2019 | Improved repository layout


tmontes:
### Context

This is an "umbrella" issue that follows from a brief discussion I had with @ntoll at EuroPython yesterday, about many Mu related, somewhat high-level, project orientation topics.

One of those is about improving the repository layout such that it becomes, in a word, more readable. That should simplify everybody's lives, especially those who are new to the Mu's codebase.

I'll list the few ideas we discussed as individual "checkboxes" to allow for discussion. From there, we may either relate each such task to a dedicated issue, or have it "closed" by means of one or more PRs.

### Ideas

* [ ] Move to an `src/` based layout in such a way that what is now the `mu/` directory/package at the root, is moved to `src/mu/`. There are multiple benefits to this, in particular related to test and packaging reliability (see, for example, the first part of the article at https://hynek.me/articles/testing-packaging/). There might be a tradeoff in that tracking history changes with git becomes less easy, but that is something that, in time, will no longer be an issue.

* [ ] Re-arranging files in the root:
  * [ ] Move `ROADMAP.rst` to docs?
  * [ ] Remove `force-rechecks.txt`?
  * [ ] Throw away `run.py`, there's no longer a good motive to keep it, AFAICT. As a matter of fact, it leads to confusion in that there are now at least two ways to run Mu, when developing.
  * [ ] Review `todo.txt` and create issues if needed. Then drop the file.
  * [ ] Move `win_installer.py` to the `package/` directory.

* Re-arrange the files in the `package/` directory.
  * [ ] Create a `mac`, `lin`, `win` sub directory (or more, on per linux dist?).
  * [ ] Move the files to the "proper" place.
  * [ ] `package/README.rst` looks outdated: integrate what's relevant in the docs and delete it.
  * [ ] Remove the `__init__.py` from there: it is not a package.
  * [ ] Move the `debian` directory in the root here.

* [ ] Not sure what the `conf/` directory is, but seems packaging related. @ntoll suggested pinging Kushal Das, any pointers? :)

* General code layout:
  * I feel that some modules could be broken down into smaller modules. Maybe that could make reading/understanding the code a bit easier/less frightening.
  * [ ] `mu/logic.py`: Editor class is big, but there are many other bits and pieces.
  * [ ] `mu/modes/base.py`: move MicroPythonMode class to its own module?
  * [ ] `mu/interface/dialogs.py`: break apart with one module per dialog?
  * [ ] `mu/interface/main.py`: break apart with one module per component?
[...]

PR #895, opened 2019-07-14
Simpler, more maintainable, and faster, Unix makefile based checks.

tmontes:
Addresses #892.

Next:
* Would love someone to review the change.
* @tjguk, would you like to create a "parallel" PR for the Windows counterparts? (or maybe review something I could create?)

Standing by. Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,8 @@ before_install:
   - python -c "import sys; print(sys.executable)"
   - python -m pip --version
   - pip --version
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv /tmp/venv --python=python3; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source /tmp/venv/bin/activate; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv venv --python=python3; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source venv/bin/activate; fi
 
 install:
   # Install Mu and its dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,8 @@ before_install:
   - python -c "import sys; print(sys.executable)"
   - python -m pip --version
   - pip --version
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv venv --python=python3; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source venv/bin/activate; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv /tmp/venv --python=python3; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source /tmp/venv/bin/activate; fi
 
 install:
   # Install Mu and its dependencies
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ else
 endif
 
 pyflakes:
-	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ ./utils/ win_installer.py
+	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ win_installer.py
 
 pycodestyle:
 	pycodestyle --ignore=E731,E402,W504 --exclude=./mu/contrib/,./mu/modes/api make.py ./mu/ run.py setup.py ./tests/ win_installer.py
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,7 @@ all:
 	@echo "\nThere is no default Makefile target right now. Try:\n"
 	@echo "make run - run the local development version of Mu."
 	@echo "make clean - reset the project and remove auto-generated assets."
-	@echo "make pyflakes - run the PyFlakes code checker."
-	@echo "make pycodestyle - run the PEP8 style checker."
+	@echo "make flake8 - run the flake8 code checker."
 	@echo "make test - run the test suite."
 	@echo "make coverage - view a report on test coverage."
 	@echo "make check - run all the checkers and tests."
@@ -48,11 +47,8 @@ else
 	python run.py
 endif
 
-pyflakes:
-	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ win_installer.py
-
-pycodestyle:
-	pycodestyle
+flake8:
+	flake8
 
 test: clean
 	pytest --random-order
@@ -60,7 +56,7 @@ test: clean
 coverage: clean
 	pytest --random-order --cov-config .coveragerc --cov-report term-missing --cov=mu tests/
 
-check: clean pycodestyle pyflakes coverage
+check: clean flake8 coverage
 
 dist: check
 	@echo "\nChecks pass, good to package..."
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ pyflakes:
 	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ win_installer.py
 
 pycodestyle:
-	pycodestyle make.py ./mu/ run.py setup.py ./tests/ win_installer.py
+	pycodestyle
 
 test: clean
 	pytest --random-order
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ pyflakes:
 	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ win_installer.py
 
 pycodestyle:
-	pycodestyle --ignore=E731,E402,W504 --exclude=./mu/contrib/,./mu/modes/api make.py ./mu/ run.py setup.py ./tests/ win_installer.py
+	pycodestyle make.py ./mu/ run.py setup.py ./tests/ win_installer.py
 
 test: clean
 	pytest --random-order
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,10 @@ else
 endif
 
 pyflakes:
-	find . \( -name _build -o -name var -o -path ./docs -o -path ./mu/contrib -o -path ./utils -o -path ./venv -o -path ./package \) -type d -prune -o -name '*.py' -print0 | $(XARGS) pyflakes
+	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ ./utils/ win_installer.py
 
 pycodestyle:
-	find . \( -name _build -o -name var \) -type d -prune -o -name '*.py' -print0 | $(XARGS) -n 1 pycodestyle --repeat --exclude=package/*,build/*,docs/*,mu/contrib*,mu/modes/api/*,utils/*,venv/*,.vscode/* --ignore=E731,E402,W504
+	pycodestyle --ignore=E731,E402,W504 --exclude=./mu/contrib/,./mu/modes/api make.py ./mu/ run.py setup.py ./tests/ win_installer.py
 
 test: clean
 	pytest --random-order
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,3 @@
 [pycodestyle]
 ignore = E731, E402, W504
-exclude = ./mu/contrib/, ./mu/modes/api/
+exclude = ./docs/, ./mu/contrib/, ./mu/modes/api/, ./package/, ./utils/
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,3 @@
-[pycodestyle]
+[flake8]
 ignore = E731, E402, W504
 exclude = ./docs/, ./mu/contrib/, ./mu/modes/api/, ./package/, ./utils/
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,3 @@
+[pycodestyle]
+ignore = E731, E402, W504
+exclude = ./mu/contrib/, ./mu/modes/api/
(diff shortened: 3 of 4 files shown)


tmontes:
Wow, look at the CI speed improvements: from 267 seconds [here](https://travis-ci.org/mu-editor/mu/jobs/558504787) down to 21 seconds [here](https://travis-ci.org/mu-editor/mu/jobs/558510310)! :)

Interesting note, the change is checking too many things: `./utils/avatar.py` is 3.6 only and wasn't included in the checks before. Will update the checks, for strict equivalency.

dlech:
It should be possible to put all of the configuration options for both `pyflakes` and `pycodestyle` in `setup.cfg` (or `tox.ini`), then they are maintained in one place for all platforms.

tjguk:
@tmontes I'm neutral on @dlech suggestion of using a config file of some sort, but I'll wait until you respond. Obviously if you do decide to go that route then I can make use of the same file from within the Windows make.py

tmontes:
@dlech, that's actually a very good idea!

I'll have a look at that sometime this week (I hope!) and update the PR.

carlosperate:
One thing we should double check if using config files automagically detected by pycodestyle and pyflakes is to ensure that they are not automatically picked-up by the tools when the user clicks on the Check button for their own code.

tmontes:
> One thing we should double check if using config files automagically detected by pycodestyle and pyflakes is to ensure that they are not automatically picked-up by the tools when the user clicks on the Check button for their own code.

...good catch. I suppose they won't, given that a `setup.cfg` file will not be packaged. But that must be verified to be true, correct. :)

tmontes:
@dlech, @tjguk, @carlosperate,

I've just updated the PR moving the inline `pycodestyle` options from the `Makefile` to a more cross-platform and, importantly, reusable, `setup.cfg`. `pyflakes` isn't using any options thus needed no such "treatment". :)

I checked that it did not affect the in-Mu check behaviour (failed to trigger the ignored `W504`, but easily triggered the other ignored ones: Mu users still get them!)

PS: There is still some inconsistency between the "target files" being checked with `pycodestyle` and `pyflakes`, but I would suggest that to be handled under its own dedicated topic / issue.

Your feedback is appreciated. :)

tmontes:
How nice of macOS CI to fail now... :)

tmontes:
...wondering: other than a "non-substantive/silly/fake" push, is there any way to force a 2nd CI run?

tmontes:
Oh, it's running again? Wondering if it did that by itself or if someone just asked it gently to retry... :)

carlosperate:
Sorry I didn't reply before as I was in a hurry, but restarted the macOS worker.

tmontes:
> Sorry I didn't reply before as I was in a hurry, but restarted the macOS worker.

Thanks @carlosperate ! :)

dlech:
> `pyflakes` isn't using any options thus needed no such "treatment". :)

We are still passing file names, so I would call that options. `setup.cfg` could exclude files as the existing Makefile does instead of explicitly including files and directories.

tmontes:
> We are still passing file names, so I would call that options. `setup.cfg` could exclude files as the existing Makefile does instead of explicitly including files and directories.

...of course! How didn't I realise that? Thanks for highlighting it. :)

I've managed to do that with `pycodestyle` by converting "targets" in the `Makefile` to exclusions in `setup.cfg`. However, `pyflakes` doesn't seem to handle exclusions/targets other than from the command line. I did try a plain `pyflakes .` (the `.` seems to be required) which is just complaining about an unused import in `./docs/conf.py`... So maybe we can go the `pyflakes`-everything *and* fix the unused import.

Thoughts? (or pointers on how to tell `pyflakes` what to check in `setup.cfg`?)

dlech:
Hmm... I guess I usually use `flake8` which in turn runs `pyflakes` and `pycodestyle`

tmontes:
> Hmm... I guess I usually use `flake8` which in turn runs `pyflakes` and `pycodestyle`

That might be a way forward, but I'd like to stick to literal refactoring within this PR, in the sense of "avoiding changing semantics as much as possible". Maybe I'm being too strict... :)

**NEEDS THOUGHT**:
* The previous commit lead to an interesting failure where `pycodestyle` recursed into `venv` which is the virtual environment created for the purposes of installation and test running @ Travis.
* In the current commit I moved the Travis virtual environment away from the repository root, to confirm that the fundamental idea works in CI.
* However, different developers are bound to have their virtual environments somewhere else and/or within the repo with varying names: for example, `.gitignore` includes references to names like `./env`, `./.env/`, and `./.venv/`.
* With that said, I'm not sure that going the *"just run `pycodestyle` and `pyflakes .` on the full repo, grabbing ignores from `setup.cfg`"* will scale nicely across a possibly diverse set of developer preferences!

Thoughts? :)

dlech:
>     * With that said, I'm not sure that going the _"just run `pycodestyle` and `pyflakes .` on the full repo, grabbing ignores from `setup.cfg`"_ will scale nicely across a possibly diverse set of developer preferences!

I think there is a rather short list of common names used for venv folders, so excluding the same ones found in `.gitignore` should be sufficient.

tjguk:
Thanks, @tmontes. Now that we're fairly stable I'll take a look at the PR and -- probably -- branch off it for a windows-based equivalent

.travis.yml
```diff
@@ -45,8 +45,8 @@ before_install:
   - python -c "import sys; print(sys.executable)"
   - python -m pip --version
   - pip --version
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv venv --python=python3; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source venv/bin/activate; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv /tmp/venv --python=python3; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source /tmp/venv/bin/activate; fi
```
carlosperate: Good catch here! I guess we never realised PyFlakes was going crazy analysing this directory.

Makefile
```diff
@@ -49,10 +49,10 @@ else
 endif
 
 pyflakes:
-	find . \( -name _build -o -name var -o -path ./docs -o -path ./mu/contrib -o -path ./utils -o -path ./venv -o -path ./package \) -type d -prune -o -name '*.py' -print0 | $(XARGS) pyflakes
+	pyflakes make.py ./mu/ ./package/ run.py setup.py ./tests/ win_installer.py
```
carlosperate: My main concern in this case is that if we had a new file in the root directory it would not be checked unless this line is edited, right?
In that case somebody could add a new file, run the tests, all pass and there would be no reason to not trust the CI results.

tjguk: Quis custodiet ipsos custodes?

I don't think it's the worst thing in the world if we miss running pyflakes on a file. And whatever solution we come up (eg a metapyflakes command which checks which files are run by pyflakes) would fall foul of the situation where decided to add a new checker... :)

tmontes:
@carlosperate, @tjguk, good observations! Thanks.

I feel that going with `flake8`, mentioned by @dlech, may allow us to get the best of both worlds:

* Still running both `pycodestyle` and `pyflakes`.
* Having both configured from the same `setup.cfg` file: thus creating another "single source of truth" for cross-platform checks.
* Simplifying both the `Makefile` and `make.py` down to "just" running `flake8`.
* Supports exclusions, thus does not need the manual specification of "targets": we just say "check this" it whichever new files/directories etc that are not excluded will be checked.

That, however, will imply revising Mu's check implementation that, AFAUI, is currently running (no idea how) the current tools... I'll try to read a bit more on it tomorrow and see how much effort that implies. But I have a gut feeling that it might be a nice tooling / setup improvement: same checks, more reliable, faster. :)

.travis.yml
tmontes: If I understand the current implementation, `./venv` is being excluded in the `find ` command options.

Makefile
tmontes: @carlosperate, on your note about "new file in the root directory", I'll refer to #891 where I propose a few changes to the general repository layout (main motive: repo readability):

* `run.py` should go away as there is (I suppose) no reason for it to exist, any longer.
* `win_installer.py` should be moved under `./package/windows` or something.
* We're left with `setup.py` and `make.py`.
* I see no reason for more `.py` files in the root of the repository.

Nonetheless, my other comment, concerning the use of `flake8` instead of `pycodestyle` and `pyflakes` (which seems to be nearly the same), has the potential to address such "new files" automatically. :)

tmontes:
@carlosperate, @dlech, @tjguk, I guess this is now looking much better that what I started off with:
* Inspired by @dlech's comments, I brought in `flake8`:
  * It "wraps" `pycodestyle` and `pyflakes` (and can do more, but I left it at that).
  * It supports file-based configuration, in `setup.cfg`, where file/directory exclusions are declared. Better cross-platform portability/reuse and config readability.
  * Running all checks, with the same semantics, is now just a matter of running `flake8`.

Result:
* Simpler, more readable/maintainable, and faster Unix `Makefile` based checks.
* Porting to the Windows `make.py` "makefile" should be easy (hopefully!) and, I suppose, allow us to simplify the code in `make.py` (given that it now needs to do less by itself).

Bonus:
* Given than `flake8` depends on `pycodestyle` and `pyflakes`, no changes to Mu's "code check" tool are needed at all. :)

Side note:
* Either Travis or existing tests seem to be flaky.
* CI failed [here](https://travis-ci.org/mu-editor/mu/jobs/560366714) with no apparent relationship to the committed changes. :/

Have I missed something?

Makefile
carlosperate: Yes, but something like `pyflakes ./*.py` seems to work to look at all the python files in the root directory without looking into subdirectories. Don't know if it's just luck or valid argument syntax, but if it is valid we could just replace this with:

```
pyflakes  ./*.py ./mu/ ./package/ ./tests/
```

tmontes: > Yes, but something like `pyflakes ./*.py` seems to work to look at all the python files in the root directory without looking into subdirectories. Don't know if it's just luck or valid argument syntax, but if it is valid we could just replace this with:

...please see my latest note / commits (down there!) where `flake8` makes everybody's life simpler, while handling all the cases we've been discussing. :)

carlosperate: Sorry, for some reason when I typed my reply my browser didn't show me your reply to this thread, the response on the main thread, nor the new commits (I assume it showed me an old cached version from last night).

I'll have to read all the other issues and PRs some other time (probably during the weekend, there is a lot to digest), but as a quick question, what is the alternative to `run.py`? I still use as my normal entry point to running Mu from source.

tmontes: > [...] but as a quick question, what is the alternative to `run.py`?

```
$ mu-editor
```

or, if you want to go all Pythonic,

```
$ python -m mu
```

carlosperate: `mu-editor` depends on having installed the package in editable mode, which I am always a bit sceptical about, since I've run into bugs in this area within pip, pipenv and poetry (all fixed eventually, but now I am always second guessing if I'm running the version of the code I think I am running).
`python -m mu` depends on running that command from the project root or it will run the installed version instead of the local one, which is fine, but I guess I liked how explicit `python run.py` was :)

tmontes: > `mu-editor` depends on having installed the package in editable mode, which I am always a bit sceptical about, since I've run into bugs in this area within pip, pipenv and poetry (all fixed eventually, but now I am always second guessing if I'm running the version of the code I think I am running).

I think you may be misunderstanding it: `mu-editor` is a `setuptools` [entry point](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) that will be created on all platforms after installation regardless of it being i) from source, in editable mode, ii) from source in "normal mode" with, say, `pip install .`, or iii) from a local or PyPI sourced wheel. Give it a try! :)

> `python -m mu` depends on running that command from the project root or it will run the installed version instead of the local one, which is fine, but I guess I liked how explicit `python run.py` was :)

I guess I understand your perspective but it is very important to run the "installed" version for many reasons (example, you want to ensure that resources such as, say, icons, are properly packaged/installed). Developers tend to run the "installed" versions which "matches" the source version (thus working with editable installs).

For these reasons (which deserve a more complete discussion/explanation) I also propose moving to an `src`-based layout in #891 -- but no hurries there, and happy to discuss. :)

PS: I mentioned `python -m mu` for completeness. We could make it "go away", I guess, and provide a single way of running Mu. :)

carlosperate: I should have clarified that I meant my comment was only from the perspective of running Mu from source as a developer that plans to modify the source code and run that version of Mu (which is why I was mentioning installing in editable mode).

Of course users **should not** be running it this way, and they have different installation instructions for that. And there are other reasons to run the installed version, which all can happily coexists with the `run.py` file. But if it has to go it has to go, I'll just have to be careful when running `python -m mu`.

tjguk:
Sorry, I slightly lost track of this. The changes to make.py are fairly trivial (which is good!) so I'll probably just start again from this PR current position.

ntoll:
Hi Folks, finally playing catch up -- expect much Mu reviewing/merging this afternoon. :-)

@tmontes thank you for making this happen, and thank you to @dlech @carlosperate and @tjguk for the input.

I've read through all the commentary above and can't see any problem with the decisions that have been made or the implementation details.

Question for @tjguk -- is there a Windows `make.py` update somewhere -- I don't see any PR or branch referenced in the discussion above.

I've checked everything works on my laptop, so merging now.

tjguk:
> Question for @tjguk -- is there a Windows make.py update somewhere -- I don't see any PR or branch referenced in the discussion above.

*cough* apparently, although I thought the changes were trivial, I didn't actually do them :) I'll take a look at this when I get a bit of dev time

ntoll:
Thank you! (As always)

tjguk:
Just fixed make.py via #919

PR #898, opened 2019-07-17
Add comment-only requirements.txt to guide beginner developers.

### LINKED ISSUE #776: Clean up requirements.txt / setup.py

tmontes:
Following up on the recently reviewed idea [here](https://github.com/mu-editor/mu/issues/776#issuecomment-511608456), this PR implements a comment-only `requirements.txt` in order to guide users and developers wanting to install from the source repository.

Merging it closes #776. 

Not sure about the "quoting" with backticks: is that intuitive? Can we do better? Feedback welcome.


ukBaz:
Looks good. The quoting looks OK to me but there again I know the answer. As an alternative, is it clearer to people (who don't know the answer) if the command is put on a separate line?
For example

```
# QUICK INSTALLATION
# ------------------
#
# To install from the commandline, run:
# pip install .
# (Make sure you include the trailing dot.)
# This installs Mu and all its runtime dependencies.
#
# To setup a development environment, run:
# pip install -e ".[dev]"
# This installs Mu and its runtime dependencies, as above, along with all the
# development ones.
#
# In either case, the `mu-editor` command should launch Mu.
#
# For more details, including platform specific notes on the Raspberry Pi,
# please refer to https://mu.readthedocs.io/.
```

requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+# This installs Mu and all its runtime dependencies.
+#
+# To setup a development environment, run `pip install -e ".[dev]"`, instead.
+# This installs Mu and its runtime dependencies, as above, along with all the
+# development ones.
+#
+# In either case, the `mu-editor` command should launch Mu.
+#
+# For more details, including platform specific notes on the Raspberry Pi,
+# please refer to https://mu.readthedocs.io/.
```
carlosperate: We could try adding only `.[dev]` here? I tried it on my system and seems to work.

tmontes: > We could try adding only `.[dev]` here? I tried it on my system and seems to work.

We could, but I see two negatives in doing that:

1. We can't cater to both "people installing from source" and "people installing a dev environment". Which criteria would be used to select between either `.[dev]` or `.`? :)

2. We lose the opportunity of "teaching" how to install `setup.py` based projects. :)

PS: I had thought about that, but decided we would be better with either i) no `requirements.txt` at all or, ii) a comments only variation, teaching people used to `requirements.txt` how to use / install a `setup.py` based source repo.

carlosperate: If we were to implement this suggestion (`.[dev]`) it would probably be better to add the editable flag as well `-e .[dev]`.

> We can't cater to both "people installing from source" and "people installing a dev environment".

The requirements.txt is usually a "install everything" type of command. If people are looking for customising their installation they are usually not doing a `pip install -r requirements.txt` blindly. In that case they should look at thee docs.

In most cases requirements.txt is a dump from pip freeze, which usually includes everything the package needs, and package itself if it is installed as part of running the application. Perhaps we should not deviate from that, but having an empty requirements.txt file feels more confusing.

Ultimately if people are going to open the requirements.txt to investigate, then if the file didn't exist they would open the setup.py file to investigate, reaching a similar conclusion to what is described here.

I think in most cases people purposely wanting to do a minimal install (instead of a full one) are more likely to do `pip install .` or `python setup.py` first. I might be wrong, but I always understood the `pip install -r requirements.txt` to be the bells and whistles install.

> Which criteria would be used to select between either .[dev] or .? :)

There shouldn't be one, if it install something it should install everything :)

> We lose the opportunity of "teaching" how to install setup.py based projects. :)

I'd rather teach them to look at the docs.

Specially I would much rather point at the docs rather than duplicate in info.

tmontes: > I'd rather teach them to look at the docs.

Agreed! :) Does that mean that you stand on the "don't add a requirements.txt file at all" side of the fence? :)

carlosperate: Don't feel strongly about including it or not, but if we do include it I would prefer for it to either:
- Have `-e .[dev]` only, which is closer to what I would expect from it. Compared with the previous requirements.txt, this version would be more "pinned" than before.
- If it's just a "teaching opportunity", then the most minimum explanation (think a line or two of text) and a link to the docs. To try to avoid duplications that can get out of sync with what we aim to be the the source of truth (the docs).

tmontes: Oh, I hadn't thought about that! I think I like the "minimalistic" and DRY approach at such file's contents. :)

Maybe it could just say something like:
```
Mu doesn't use requirements.txt:
- To install please refer to https://codewith.mu/.
- To setup a development environment see https://mu.readthedocs.io/.
```

And we'd be down to two URLs, catering to different audiences.

carlosperate: Sounds good.

ukBaz:
I very much like the proposed short `requirements.txt` file and would agree with going with that. I think the benefits greatly outweigh the downsides.
If this assumption proves wrong and it causes issues for users then we can extend at that point.

tmontes:
Flaky Travis or flaky tests?... Won't worry now. Pushed an empty commit to trigger CI rerun. :/

ukBaz:
FYI: To help with debugging Travis issues you might find the following of interest:
https://docs.travis-ci.com/user/running-build-in-debug-mode/

tmontes:
Thanks for the pointer, @ukBaz. I have no access to the settings neither on this project nor on Travis' side, so there's not much I can do myself... But that is useful information, indeed.

Back on to the PR: it seems that the macOS Travis' Gods aren't around today, given the nature of the failures in the last two runs. Maybe they'll be back tomorrow! :)

PS: Back of mind wondering if life would be better with Azure Pipelines... *sigh!*

tjguk:
I certainly don't feel strongly about this one. I'm basically +1 with @ukBaz: if it causes problems we can go back and expand it easily.

ntoll:
LGTM :shipit:

Thanks, once again @tmontes and everyone for pushing this forward.

PR #899, opened 2019-07-18
Dirty tab styling

ZanderBrown:
Not sure if this is what we want to go with but the code has been written so here it is

image


tmontes:
First impression: I guess I like it. :)

On macOS, however, the "x" is displayed towards the left-hand side of the tabs, and we get something like this, where the "saved, thus, blank" may feel a bit off:

Screen Shot 2019-07-18 at 11 07 54

Maybe adding another "blank" icon to the right of the "title" could give as a more balanced and readable tab... What's your feeling?

ZanderBrown:
> adding another "blank" icon to the right of the "title"

Unfortunately unlike [GtkNotebook](https://developer.gnome.org/gtk3/stable/GtkNotebook.html) [QTabWidget](https://doc.qt.io/qt-5/qtabwidget.html) is quite limited

Originally close was right aligned on macOS as well, unfortunatly Qt had a bug handling that correctly

/me dives down into the depth of Qt docs

ZanderBrown:
Right so having spent some time in the depth of Qt docs followed by a trip to the Qt source code we have

image

How is this done? Well Qt doesn't let us set custom content in tab labels *but* is does let us set custom buttons

However it doesn't actually care if the custom button is a button!

Thus we can set the button to a container with an actual button (to replace the close button) and a label to show an icon

Who knows why Qt uses a label to show icons

tmontes:
Oh, that's some very nice Qt wizardry! :)

I wonder, maybe for aesthetics motives, if the tab title/label could be slightly padded on the left side to somehow balance the "blank-ness" on the right side when it is saved... (thinking out loud, not asking for it... more: I wouldn't say I'm an authority on aesthetics / usability!!!) :)

I played around with it for a little bit and managed to generate the following stack trace (tried to reproduce but couldn't)... Does it ring any immediate bell?
```
  File "/Users/tiago.montes/Work/Python/github.com/mu/mu/interface/main.py", line 210, in close
    self.removeTab(tab_id)
  File "/Users/tiago.montes/Work/Python/github.com/mu/mu/interface/main.py", line 159, in removeTab
    modified = self.widget(tab_id).isModified()
AttributeError: 'NoneType' object has no attribute 'isModified'
```

ZanderBrown:
And here we thought 13517 lines of tests (ignoring comments & blanks) was supposed to catch all these things, fixed :-) 

We could increase padding on the left but I'm not sure it adds much other than reducing the space available for other tabs

The spacing between the title and close buttons is perhaps a little odd but web browsers generally seem to have a gap there as well ¯\_(ツ)_/¯

ntoll:
@ZanderBrown having tests is no guarantee there are no bugs. :-)

FWIW tests are simply a safety net for our future selves, IYSWIM.

ntoll:
OK... I've read through the code, checked it works as advertised and run the test suite. LGTM.

Great work @ZanderBrown and thanks for the input @tmontes :-)

Issue #896, opened 2019-07-15
Misbehaving "make clean": breaks development environment


tmontes:
Once the development setup became based on `pip install -e <path-to-mu-repo>`, the existing `make clean` needs fixing.

The reason is that the `-e` installation creates a `mu_editor.egg-info` file in the repository root that `setuptools` / `pip` use to track the editable installation. Among other things, the file needs to be there for the `mu-editor` command to work. This does not affect "normal", non-editable installations, though.

The current `make clean` implementation removes the `mu_editor.egg-info` file, thus breaking the editable install, thus breaking the development environment.

The near term solution is trivial: fix the `Makefile` such that the `clean` rule does not remove the file. A complementary solution / evolution to this scenario, that "pollutes" the root of the repository with a somewhat alien `*.egg-info` solution, is to convert the repository to an `src`-based layout which, by itself, brings a few other benefits (see one of the proposed entries in the umbrella issue #891). :)

tmontes:
#901 merged. Closing this. :)

PR #901, opened 2019-07-19
Don't remove `mu_editor.egg-info` on make clean: breaks editable installs.

tmontes:
Addresses #896.

Interestingly, the Windows `make.py` counterpart does not remove it. Go figure! :)


ntoll:
Aha... so *that's* why it happens. Makes sense. An easy one to merge. Thank you @tmontes :-)

Issue #762, opened 2019-02-04
UI/UX | Check button vs. Positive feedback


tmontes:
## Context

I've been using Mu recently, more and more, while preparing an introductory programming workshop for kids. One of the things I've been trying really hard is to "step into their experience level" as much as possible. From that, I have found a few UI/UX issues I think can be improved.

## The issue

Much like the "Mode button vs. Current mode indicator" issue #761 I just logged, the "all ok" feedback and the Check button that triggers it are on completely opposite "locations" within Mu's window:

![mu-ui-study-check-current](https://user-images.githubusercontent.com/1912206/52223518-b00b0700-289d-11e9-9e1b-52e54bb1068a.png)

While the "not ok" feedback is displayed inline with the code, which I find good, the "all ok" friendly feedback can easily be missed (possible beginner's experience: hmmm, let me click this button here to "check".... nothing happened?... maybe I'm not understanding... moves on with a feeling of "I didn't get that one!").

## Proposal (for discussion, of course!)

Let's have the "all ok" positive and friendly feedback be displayed more "in-your-face" as a reaction to clicking the Check button. Here's a possible idea where a tooltip like message would be displayed right under the button:

![mu-ui-study-check-allok](https://user-images.githubusercontent.com/1912206/52223885-8e5e4f80-289e-11e9-9026-96ce412b852f.png)

Notes:
* I went for a somewhat larger font size, given that "all ok" messages tend to be short sentences, helping the differentiation from tooltips which are displayed in a smaller font and a grayish background.
* A yellow background (same yellow as the button borders') helps the feedback stand out in the white/grayish background.
* Not to be implemented as a tooltip: the message would only go away once any mouse click and/or keyboard typing happens anywhere within Mu. Not timeout based.
* A variation on the idea: overlay a semi-transparent grey layer on top of the editor and display the "all ok" message in a vertically and horizontally centered yellow background box, pretty much like the one above. This would be a more in-your-face feedback and behave exactly the same, going away after a click.
* In any case, the status bar feedback would be completely eliminated.

## Related thoughts

Given that the check button is somehow stateful, acting as a "highlight/hide problems" toggle when there are "problems" in the code, maybe the overall behaviour could be improved in a way that its own icon could change depending on the current state:

* After "all ok" feedback, it would become the current "thumbs up" icon.
* After an "errors found" feedback, it could become a "thumbs down" icon variation.
* Any editing or front tab selection change would reset the button an "idle, i don't know" icon.
[...]

PR #904, opened 2019-07-21
Checker feedback

ZanderBrown:
Possibly resolves https://github.com/mu-editor/mu/issues/762

Peek 2019-07-21 18-44

Nothing particularly fancy just turning the hypothetical into real so we can better assess the idea


ntoll:
@ZanderBrown you're on :fire:

Just about to check/test this.

PR #907, opened 2019-07-26
DO NOT MERGE - Feedback, comments and critique please..!

ntoll:
I was in Boston at Tufts university earlier this week and was able to get feedback about how university level beginner developers find using Mu. One of the most interesting suggestions made was the ability to copy-and-paste code from the main editor window into the live REPL. This isn't the first time this feature has been requested (I seem to remember it being mentioned in a long distant micro:bit-only days), but there was a particularly compelling use case which made me go "aha!":

Chris Rogers (a professor in the engineering school) explained how being able to highlight a section of interesting code and have only that code evaluated, one line at a time, in the REPL was a really helpful way of debugging and watching things unfold, especially if the code is run on a MicroPython based board (where the graphical debugger won't work).

repl_demo

So, I've quickly bodged together a very simple proof of concept (see the diff) which gives you a sense of what I think Chris means. It is buggy, has no tests and is incomplete (see TODO in the code). However, I believe it shows promise and would love feedback, comments and constructive critique with a view to finding a consensus for what things should be named, how things should work and the scope of this potential feature. As always, I'm not precious about my code or suggestions, so please don't hesitate to knock holes in what I'm suggesting.

Open questions in my mind's eye:

* What should be written on the button in the REPL pane so it is both short and makes sense? ;-)
* What should happen if no code is selected in the main editor? Run it all, or simply run nothing?
* Un-indenting (or dedenting (or whatever the opposite of indenting is)) should happen automatically to code which may be copied from within an existing block of code... discuss...
* This may appear to be a sort of duplication of the Run button in the ESP mode... I wonder if / how we can consolidate such a potential overlap? (Would love to hear your thoughts on this @dybber)
* Since this is a feature of the REPL pane, it's featured across all modes that have a REPL. I wonder, since we're discussing this, what other universal REPL-ish interactions may be helpful / useful (the opposite sort of action -- copy all REPL based commands to editor pane -- immediately springs to mind).

In any case, this is a proof of concept with a view to encouraging debate to arrive at a consensus. Would love to hear your thoughts..!

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -82,6 +82,7 @@ class EditorPane(QsciScintilla):
 
     # Signal fired when a script or hex is droped on this editor
     open_file = pyqtSignal(str)
+    code_selected = pyqtSignal(int, int)
 
     def __init__(self, path, text, newline=NEWLINE):
         super().__init__()
@@ -515,6 +516,7 @@ class EditorPane(QsciScintilla):
             # Highlight matches
             self.reset_search_indicators()
             self.highlight_selected_matches()
+            self.code_selected.emit(line_from, line_to)
 
     def toggle_line(self, raw_line):
         """
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -24,7 +24,7 @@ from PyQt5.QtCore import QSize, Qt, pyqtSignal, QTimer, QIODevice
 from PyQt5.QtWidgets import (QToolBar, QAction, QDesktopWidget, QWidget,
                              QVBoxLayout, QTabWidget, QFileDialog, QMessageBox,
                              QLabel, QMainWindow, QStatusBar, QDockWidget,
-                             QShortcut, QApplication)
+                             QShortcut, QApplication, QHBoxLayout, QPushButton)
 from PyQt5.QtGui import QKeySequence, QStandardItemModel
 from PyQt5.QtSerialPort import QSerialPort
 from mu import __version__
@@ -307,6 +307,7 @@ class Window(QMainWindow):
         Adds a tab with the referenced path and text to the editor.
         """
         new_tab = EditorPane(path, text, newline)
+        new_tab.code_selected.connect(self.repl_run_code)
         new_tab.connect_margin(self.breakpoint_toggle)
         new_tab_index = self.tabs.addTab(new_tab, new_tab.label)
         new_tab.set_api(api)
@@ -501,7 +502,21 @@ class Window(QMainWindow):
         Adds the referenced REPL pane to the application.
         """
         self.repl_pane = repl_pane
-        self.repl = QDockWidget(_('{} REPL').format(name))
+        title_widget = QWidget()
+        self.repl = QDockWidget()
+        repl_title = QHBoxLayout()
+        label = QLabel()
+        label.setText(_("{} REPL").format(name))
+        repl_title.addWidget(label)
+        self.repl.run_select = QPushButton("RunMe")
+        self.repl.run_select.setEnabled(bool(self.current_tab.getSelection()))
+        self.repl.run_select.clicked.connect(self.repl_run_click)
+        repl_title.addWidget(self.repl.run_select)
+        repl_title.addStretch()
+        repl_title.setSpacing(0)
+        repl_title.setContentsMargins(4, 4, 4, 4)
+        title_widget.setLayout(repl_title)
+        self.repl.setTitleBarWidget(title_widget)
         self.repl.setWidget(repl_pane)
         self.repl.setFeatures(QDockWidget.DockWidgetMovable)
         self.repl.setAllowedAreas(Qt.BottomDockWidgetArea |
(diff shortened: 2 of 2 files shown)


chrisbuerginrogers:
Where I have seen this most powerful is when running a code snippet and then being able to interrogate and mess with the variables (or class methods etc) because they are all still in memory.  It has allowed my students to play with and discover more easily.  Thanks Nicholas.

carlosperate:
Sounds like good feature for learners, I really like it!

Below are my opinions to the questions. I'm writing this quickly, so apologies if the tone sounds a bit more imperative, I do mean all of these as suggestions:

I think adding this functionality to the right click menu on the selection would be useful as well.

>    What should be written on the button in the REPL pane so it is both short and makes sense? ;-)

`Run selection` or `Execute selection`?

>    What should happen if no code is selected in the main editor? Run it all, or simply run nothing?

The button should be disabled until code is selected.

>    Un-indenting (or dedenting (or whatever the opposite of indenting is)) should happen automatically to code which may be copied from within an existing block of code... discuss...

I would agree with this statement.
And I think we should maintain lower indentations relative to the first line. So something like this:

```
...
        while True:
            pass
...
```
Is copied as:
```
while True:
    pass
```

I would look at the first line, find the first non-white-space character, and take that column as the initial indentation. Then subtract that indentation from all lines.
If we encounter a line that has a lower indentation than the first line, I would show an error message with a helpful message explaining the problem.

>    This may appear to be a sort of duplication of the Run button in the ESP mode... I wonder if / how we can consolidate such a potential overlap? (Would love to hear your thoughts on this @dybber)

I think limiting the button functionality to text selection is a good differentiator.

>    Since this is a feature of the REPL pane, it's featured across all modes that have a REPL. I wonder, since we're discussing this, what other universal REPL-ish interactions may be helpful / useful (the opposite sort of action -- copy all REPL based commands to editor pane -- immediately springs to mind).

`Paste to REPL` could be a good option.
Pasting is one of the things I've seen a lot of learners struggle with in the REPL.
Some are used to press "Ctrl + V" and when that doesn't work they generally don't know what to do. In some of the workshops I've been very few actually tried to right click to find the option.
So maybe having a button for that could be useful as well.

tjguk:
[Caveat: I haven't tried running the PR yet; I'm going off the screenshots above]

Amusingly, soon after I got involved with Mu, I made a fork called -- wittily, I thought -- "Nu". And the key point was exactly this kind of functionality. I may have discussed it at the time with @ntoll; I don't remember. I got something working but then I left it and now it's just gathering dust.

But I honestly see this idea -- running code from within the editor itself -- as a great way to experiment with code while building it up. Professionally I use this *all the time* in the SQL context, where you're building up what might well end up being a fully-fledged SQL script from some core statements which you can run and keep running in combination to refine your query.

In the case of SQL, it's slightly simpler because most of the output is tabular and consistent. But I don't see it being very different for Python. I'll take the PR for a spin to see how it works for me, but assuming it's going the way I think it's going, I very much applaud the notion.

tmontes:
Much like @tjguk, I haven't tried it, but wanted to share feedback as soon as possible. :)

This looks very nice and useful.

> * What should be written on the button in the REPL pane so it is both short and makes sense? ;-)

"Run" / "Execute" may be slightly misleading given that in the example, clicking it does not run the selected code. Maybe a simple "Paste" is appropriate? "Copy Selection"?

Alternatively, for a more streamlined UI (and minimizing translation efforts?) maybe it could be an icon? (which, in turn, might need a tool tip so, the "minimizing translation efforts" idea might not really apply here...)

I would also try to figure out where/how to layout the button: the "REPL" pane title feels cluttered and difficult to read the way it is (understood, this is a proof-of-concept).

> * What should happen if no code is selected in the main editor? Run it all, or simply run nothing?

Button should be disabled and its tool tip dynamically change to hint at the need to select some text to copy it into the REPL?

> * Un-indenting (or dedenting (or whatever the opposite of indenting is)) should happen automatically to code which may be copied from within an existing block of code... discuss...

Agreed, it feels like [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) will be helpful here.

> * Since this is a feature of the REPL pane, it's featured across all modes that have a REPL. I wonder, since we're discussing this, what other universal REPL-ish interactions may be helpful / useful (the opposite sort of action -- copy all REPL based commands to editor pane -- immediately springs to mind).

Supporting the REPL to editor feels sensible, yes. Again, without much thought, we may be talking about two buttons in the "REPL" pane title bar: "Copy" and "Paste". Or, instead, "Copy from Editor to REPL" and "Copy from REPL to Editor". I think, the key difference in wording (and behaviour!) is essentially dependent on whether or not such "text transfer" operations use the clipboard or not. Should they?

tjguk:
Trying it out. Don't know if it's a Win-specific issue, but the final line of selected code isn't copied to the REPL unless I select the following line feed. (And if there isn't a final line feed then I'm stuck). I'll dig a bit to see if a PR is appropriate.

I've also had hits and misses with the first [RunMe] after opening the REPL. I'll see if I can reproduce consistently.

Possibly influenced by my usual experience of this feature, I was expecting the REPL to clear on every click of [RunMe]. Is the intention here to allow, essentially, arbitrary code snippets to be run in succession, with each execution adding something to the REPL?

dybber:
Hi all, here is my input.

I think this could be a useful feature, but perhaps it will be to cluttered with another button in the panel for "Run selection", at least the button should only be shown when something is selected. An alternate location could be in the right-click menu, when you are selecting something, you are using the mouse anyway, but for Mac-users this might not be ideal I guess?

When thinking about I also think that you might want to edit something in the code you want to interact with before running, so it could also be a "Copy to REPL", then the user will have the opportunity to adjust the code before pushing enter. A "Copy to REPL" menu item in the right-click menu right next to the ordinary "Copy" menu item, would make sense to me, but other Mac users might disagree (I always use an external two-button mouse with my Mac)

Wrt. ESP-mode and the Run button, I guess the differentiator should be that this is new feature is only for when text is selected, where as the Run button runs the entire button. I think it wouldn't make sense (not in the other modes either) to have two buttons that Run the entire buffer (as would be the case if the "Run selection" button would run everything when nothing is selected)

ntoll:
Hi Folks, Thank you for all the feedback. I'll find some time to integrate all the suggestions and turn this into something that's not a spike and includes tests / docs etc... :-)

carlosperate:
Superseded by the merged PR https://github.com/mu-editor/mu/pull/1563 🎉

PR #908, opened 2019-07-31
Add snek mode

keith-packard:
Snek is a small Python subset which runs on a variety of embedded
controllers. It currently supports the ATmega328p, ATmega32U4 and
SAMD21G processors with ports to a range of boards that include those
processors.

https://keithp.com/snek


devdanzin:
Is there something I could do to help this move along? Splitting into self contained commits, resolving conflicts, reviewing the changes...?

keith-packard:
> Is there something I could do to help this move along? Splitting into self contained commits, resolving conflicts, reviewing the changes...?

Yes, I'll get these changes rebased, at which point I'd love someone to review them.

devdanzin:
Keith, thank you for creating a new mode for a great project!

It applies cleanly, neat code, sweet talking to my ol Mega 2560-ish ([cheap clone drivers needed](https://www.instructables.com/How-to-fix-bad-Chinese-Arduino-clones/)).

So far, the only issue I found is that slow_boards is optimistic: my board was sending/receiving gibberish because it isn't in the slow list. So maybe we should have a way to signal (or test that) the board is slow, even if unknown. Or just add ids for most common cheap clones.

When the device is detected, it identifies as HL-340 and Mu offers to change into MicroPython mode (which won't work with an ATmega2560), so we should take a look at how to avoid confusion there.

I really like what I see here.

mu/modes/snek.py
```diff
@@ -0,0 +1,394 @@
...
+            message = _("Cannot run anything without any active editor tabs.")
+            information = _(
+                "Running transfers the content of the current tab"
+                " onto the device. It seems like you don't have "
+                " any tabs open."
+            )
+            self.view.show_message(message, information)
+            return
+        python_script = tab.text()
+        if python_script[-1] != "\n":
```
devdanzin, line 276: Raises IndexError if the document is empty, should be something like `if not python_script or python_script[-1] != "\n"`.

keith-packard:
> So far, the only issue I found is that slow_boards is optimistic: my board was sending/receiving gibberish because it isn't in the slow list. So maybe we should have a way to signal (or test that) the board is slow, even if unknown. Or just add ids for most common cheap clones.

There are a couple of device-dependent modes which are (currently) selected only by USB id. The trouble with devices using the CH341 chip is that we have no way of knowing what kind of board they're pretending to be. Could be almost anything, from an ATmega328 to a 2560. The snek binaries currently use different baud rates depending on the USB interface in use. That's the result of a fairly ugly hack -- Snek really wants XON/XOFF flow control, but the Arduino boot loaders need a raw binary interface. FTDI usb devices, like those used for real Duemilanove boards, can be configured by the host to enable XON/XOFF flow control. Built-in USB controllers, like those in the ATmega32U4, cannot be configured by the host (due to limitations in the USB standard). So, Snek has replacement firmware for the USB interface chips in boards like the Mega. This firmware enables XON/XOFF flow control for lower baud rates (which Snek selects) while disabling it for higher baud rates (which the boot loader uses).

XON/XOFF is most important when loading programs from the host to EEPROM; you don't want to drop parts of the program.  I'm making changes to the firmware to make EEPROM loading synchronous, and that seems to solve the problem nicely, and removes the really critical need for XON/XOFF support in the serial path. With that change, we can run every Snek board at 115200 baud and use synchronous loading on serial ports which need it -- that *can* be reliably detected by USB id as all CH341 chips will need to use this, along with boards using the ATmega16u2 like the Mega.

> When the device is detected, it identifies as HL-340 and Mu offers to change into MicroPython mode (which won't work with an ATmega2560), so we should take a look at how to avoid confusion there.

I think it would be safe to add all of the CH341 USB ids to the snek list; anything capable of running MicroPython is likely to have built-in USB, and hence custom USB ids.
 
> I really like what I see here.

Thanks much for your kind words. I'll work on getting the synchronous program loading working in 'mu' and change the serial rate on all boards to the same value. That should make your boards work reliably.

mu/modes/snek.py
keith-packard: thanks. I added a 'not python_script or' to the beginning of the condition.

keith-packard:
> So far, the only issue I found is that slow_boards is optimistic: my board was sending/receiving gibberish because it isn't in the slow list. So maybe we should have a way to signal (or test that) the board is slow, even if unknown. Or just add ids for most common cheap clones.

I've reworked the internals of Snek to provide a more reliable mechanism for flow control. Instead of requiring XON/XOFF support, snek now provides an explicit synchronization handshake method. When the host sends an ENQ character (ASCII 0x05), snek responds with an ACK character (ASCII 0x06). With this, we can run all snek devices at the same rate and reliably send programs to the device.

I'll be releasing a new snek version with this feature soon.

keith-packard:
Ok, new snek release uses the ENQ/ACK based flow control and sets all boards to run at 115200 baud, avoiding the 'slow_boards' issues while still providing reliable storing of programs on boards. I'm hoping this version is ready for merging as I've got a few people downloading temporary builds for use with Snek-based devices.

keith-packard:
I've rebased this patch on v1.1.1, there are a few changes needed to support the Arduino LilyPad devices as that cannot run
at 115200 baud, so mu now needs to auto-detect the baud rate.

ntoll:
:+1: I'd like to merge this now 1.1.1 is out of our hair. What, in particular, in the main branch needs to change (baud rate detection?), or could this be a change included in this PR?

keith-packard:
The PR has all of the necessary changes already implemented. The autobaud code isn't super fancy and is only implemented for snek devices.  It knows that snek devices using a uart usually run at 115200 baud, although some (8MHz ATMega 328 boards) run at 57600 baud. It also knows that snek devices will echo b"\x14" even if they are busy. It just checks each baud rate (there are only two) to see at which rate the device echos b"\x14".

I tried to make the changes to existing files as easy to review for potential effect on other targets. There's a new function in mu/interface/main.py and a new class in mu/interface/panes.py, other than that it's just the code to hook up snek mode and device detection in mu/app.py and include snek functions in mu/modes/api/__init__.py.

Let me know if you have any concerns with this merge; I'm running 1.1.1 with this patch here and it's super awesome!

Oh, I added another patch on this branch which seemed to make CI happier with older python versions; I still haven't seen a completely clean CI run yet, but the union of successful runs appears to be complete now.

ntoll:
Hi @keith-packard - I've just taken this branch for a spin. I'm happy to merge it if you are (i.e. you've not some late breaking changes before it's "in in" the Mu main branch).

Huzzah! :tada:

keith-packard:
Thanks much! It's passing all of my tests, so I'm ready to see it integrated. Mu-ve on up!

ntoll:
Tada :tada:

PR #916, opened 2019-09-11
Add USB VID/PID for PewPew boards to the CircuitPython mode

deshipu:
I won't be using Adafruit's VID/PID indefinitely.

### DIFF
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -48,6 +48,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x1B4F, 0x8D23),  # SparkFun SAMD21 Dev Breakout
         (0x1209, 0x2017),  # Mini SAM M4
         (0x1209, 0x7102),  # Mini SAM M0
+        (0x1D50, 0x60E8),  # PewPew Boards
     ]
     # Modules built into CircuitPython which mustn't be used as file names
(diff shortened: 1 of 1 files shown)


mu/modes/circuitpython.py
```diff
@@ -48,6 +48,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x1B4F, 0x8D23),  # SparkFun SAMD21 Dev Breakout
         (0x1209, 0x2017),  # Mini SAM M4
         (0x1209, 0x7102),  # Mini SAM M0
+        (0x1D50, 0x801D),  # PewPew Boards
```
carlosperate: I've only found the `0x60e8` VID in https://github.com/openmoko/openmoko-usb-oui/blob/master/usb_product_ids.psv, is there a different source for this list or has the VID been reserved some other way?

deshipu: Och, you are right I have no idea how I made that mistake, thanks for catching it!

deshipu: I see what I did, I copied the Adafruit VID instead of mine. Thanks!

dybber:
Thanks for the contribution! I've made the necessary changes to be able to merge this and made a new pull request at #1094, and I'm thus closing this PR.

Issue #496, opened 2018-06-20
Detect blocking / dead / infinite loop in debugger child process.


ntoll:
See #494 for context.

I think we could attempt to detect that the script is blocking (for example, by adding a time-out for responses from the child process). If the response comes back, the timeout is cancelled. If the timeout is hit, all debug related buttons except for stop (which will kill the child process and clean things up) are disabled.

tmontes:
PR #924 just raised to address this.

tmontes:
Just merged #924, addressing this. Closing.

PR #924, opened 2019-09-29
Disable debug control buttons during debug execution.

tmontes:
Addressing #496 in full, IIUC. I went with a hard-coded 100ms timeout which subjectively felt reasonable in various scenarios I tried:
* Endless loops and non-"immediate" I/O (network transfers, disk reading/writing) disable the debug control buttons (except stop, which is useful).
* Faster code and I/O does not -- with too short a timeout (or none at all), the debug control buttons would tend to flicker as the user steps through the code.

Will appreciate review and feedback.

Possible follow-up work:
* Refactor the sequences of `disable_buttons_later` + `reset_debugger_highlight` into a common method (maybe `enter_debug_execution`? which could be paired with a `leave_debug_execution` that as of now just re-enables the buttons?)
* Refactor the debugger tests by creating a fixture to provide the mocked `DebugMode` object (note the repetition of `editor = Mock` + `view = Mock` + `dm = DebugMode(editor, view)`).
* Maybe we could introduce another kind of indicator: a different kind of visual hint telling the user which was the last line about to be executed when they last clicked "continue" / "step" / ...


tmontes:
Interesting, WRT to Travis testing 2854d3b21c7a3be2cf1fd6476331fdfe05935285:

* [Failed macOS tests here](https://travis-ci.org/github/mu-editor/mu/jobs/731939211#L599) with a `TypeError` in `mu/interface/themes.py`:97.
* Could it be intermittent? Tests are passing on my dev system.

tmontes:
Merging after:
* Quick validation with @tjguk on Gitter so as not to complicate work on the venv branch.
* Being more stubborn than Travis! :)

PR #929, opened 2019-10-07
Blackify

ntoll:
This PR makes two changes:

1. Adds the Black code formatter into the toolchain defined in the Makefile (i.e. code is tidied as part of `make check`).
2. The Python source files within this repository have all been reformatted by Black.

I had to make a few modifications to the flake8 command in the Makefile so Black-style-Python passes the very strict PEP8 checker.

I completely understand that folks may view this like Marmite (you'll either love or hate it, but nothing in between). I'm more than happy to NOT merge this. However, I believe we should try to dogfood, and since we have a "tidy" button, connected to Black, it could legitimately be asked why we don't use it ourselves. Personally, I quite like the fact that I can lazily write non-PEP8 compliant Python and Black will look after it.


ntoll:
Ahh... I need to add the `tidy` command to Windows. I'll do that in the morning.

Makefile
```diff
@@ -47,15 +48,25 @@ else
 endif
 
 flake8:
-	flake8
+	flake8 --ignore=E203,E731,E402,W503,W504
```
tmontes: @ntoll, if you add these to the already existing ones in `setup.cfg`:
* They will automatically be applied in both Windows and UNIX-like OSes.
* The new ignores will be right along with the already existing ones, and support minor commonts.

Makefile
```diff
@@ -47,15 +48,25 @@ else
...
 
 test: clean
 	pytest --random-order
 
 coverage: clean
 	pytest --random-order --cov-config .coveragerc --cov-report term-missing --cov=mu tests/
 
-check: clean flake8 coverage
+tidy: clean
```
tmontes: This Makefile rule ends up needing to be "duplicated" in Windows' `make.py` equivalent... I'm wondering if there could be a way of avoiding such repetition, for increased maintainability. :)

Maybe a [`pyproject.toml`](https://github.com/psf/black#pyprojecttoml) file could be created with a `[tool.black]` section specifying the correct `line-length` and `exclude` values such that "just running" `black` does what we want. That might make it DRY and work the same way on any platform, Windows or UNIX-line. :)

make.py
```diff
@@ -141,19 +143,37 @@ def flake8(*flake8_args):
...
     print("\nflake8")
     os.environ["PYFLAKES_BUILTINS"] = "_"
-    return subprocess.run([FLAKE8]).returncode
+    return subprocess.run(
+        [FLAKE8, "--ignore=E203,E731,E402,W503,W504"]
+    ).returncode
+
+
+@export
+def tidy():
```
tmontes: Refer to my note, above, on the `Makefile` about possibly DRYing this up. :)

make.py
```diff
@@ -141,19 +143,37 @@ def flake8(*flake8_args):
     """
     print("\nflake8")
     os.environ["PYFLAKES_BUILTINS"] = "_"
-    return subprocess.run([FLAKE8]).returncode
+    return subprocess.run(
+        [FLAKE8, "--ignore=E203,E731,E402,W503,W504"]
```
tmontes: See my note above, in the context of `Makefile` WRT to this. :)

tmontes:
@ntoll, I think this is indeed a very good idea. A few notes:

* `Makefile` and `make.py` can be DRYed up a bit, per my notes. Let me know if you want me to be more specific.
* Wondering whether running `make check` should run `black` or not. Will contributors expect a "check" to potentially change their working trees? Or would we be better off by running with `black --check` there, forcing contributors to `black` themselves thus being more aware of such `black`-driven file edits? This would have the added benefit that current CI would fail with non `black`-ified code, while not actually changing it in place! :) Let me know of your thoughts.

ntoll:
@tmontes thank you for the feedback. All taken on board. I'll adjust the `setup.cfg` as per your suggestion (I didn't realise `flake8` had this capability). Thanks for pointing this out.

I'm not sure how we might DRY up the `make` command for both unix and Windows. I think, once I've fixed up flake8 options, I'll merge this and we can refactor make in a separate ticket. I believe @tjguk will have a valuable perspective from a Windows PoV.

tmontes:
> I'm not sure how we might DRY up the `make` command for both unix and Windows. I think, once I've fixed up flake8 options, I'll merge this and we can refactor make in a separate ticket. I believe @tjguk will have a valuable perspective from a Windows PoV.

Possibility:
* Create a `pyproject.toml` in the repo root with contents much like:

```
[tool.black]
line-length = 79
# set include / exclude such that they match the "targets" in current Makefile / make.py
```

(details can be found [here](https://github.com/psf/black#what-on-earth-is-a-pyprojecttoml-file))

* Change `Makefile` and `make.py` to "just run `black`" with no command like arguments.
* This way, any future `black`-related adjustments will automatically be picked up on all platforms.

More:
* What do you feel about my suggestion: not actually running `black` in "edit mode" when running `make check`, but instead run it in "check mode" by passing in the `--check`?

ntoll:
@tjguk aha... I see what you mean. OK... let's make this a separate issue (#930 930)

carlosperate:
>Wondering whether running make check should run black or not. Will contributors expect a "check" to potentially change their working trees? Or would we be better off by running with black --check there, forcing contributors to black themselves thus being more aware of such black-driven file edits? This would have the added benefit that current CI would fail with non black-ified code, while not actually changing it in place! :) Let me know of your thoughts.

I agree with this, a `check` command, specially for CI, should return an error exit code if the code not formatted correctly, but I wouldn't expect it to automatically edit the code.

Right now black will just format the code right? So if somebody submits a PR with the wrong formatting, will black format the code in the CI server, return 0 and the tests be marked as passed?
Or will it format the code and return non-zero, so that the test fail?

Makefile
carlosperate: It would have been nice if flake8 supported `pyproject.toml` as well (https://gitlab.com/pycqa/flake8/issues/428), that way we could have replaced the current `setup.cfg`.
I guess we'll need to add yet another file :/ (can't wait until `pyproject.toml` takes over the world!).

carlosperate:
Btw, from git v2.23 (August 2019) it can now ignore commits on blame:
https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt

```
git blame --ignore-rev 84e91ec2c22d0abaee1295db28fdc9df112d7b18 ...
```

We could also add a file with a list of commits to ignore, but not sure that's really worth the trouble right now: https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt

Makefile
tmontes: @carlosperate, I'm with you. At least WRT `flake8` getting its configuration from `pyproject.toml`... But that issue and associated MR (PR in GitLab's words) seem to be somewhat on hold. The discussion there is interesting though. :)

tmontes:
> Right now black will just format the code right? So if somebody submits a PR with the wrong formatting, will black format the code in the CI server, return 0 and the tests be marked as passed?

AFAICT, CI will format the code and pass, but the underlying code in the repo is not necessarily `black`-ified. The "just check" approach would ensure that.

carlosperate:
> AFAICT, CI will format the code and pass, but the underlying code in the repo is not necessarily black-ified. The "just check" approach would ensure that.

I haven't had a chance to check yet, but if this is how it's currently configured I'd say it's a good case for running the black check on `make check` and run the black formatter on `make tidy`.

ntoll:
@carlosperate yes, I agree with this approach (black check on `make check`, black format on `make tidy`).

ntoll:
@carlosperate fixed in #937

PR #933, opened 2019-10-08
Travis: Add Python 3.5 tests without checker

carlosperate:
Assuming Mu still supports Python 3.5, it'll be good to continue running the tests in Python 3.5, in case Python 3.6 features accidentally sneak into the codebase.

Since Black is 3.6+, the 3.5 environment runs `make coverage` only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ matrix:
       dist: xenial 
       sudo: required
       language: python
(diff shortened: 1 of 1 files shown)


ntoll:
@carlosperate yes. I should have thought of this. But in my haste, I didn't. Thanks to you, we're back on track. :smile:

Issue #893, opened 2019-07-13
Review / Update dependency versions


tmontes:
Somehow superseding #488, including the context as a followup of #799, the currently single-sourced dependencies need to be reviewed, inline with @carlosperate and @ntoll's ideas, at least.

Thoughts:
* We have both runtime dependencies and development dependencies.
* Which should be locked (`==`), which should be latched (`>=`/`<`), which should be left alone?

More:
* Locking everything is good because it guarantees a known working configuration.
* On the other hand it is bad because it prevents security updates and, much like reported by @bennuttall in #488, may prevent packaging across different platforms.
* Not locking may result in different packaged versions of dependencies for, say, the Windows and macOS packaged distributions (imagine Windows is packaged first and, before macOS is packaged a particular dependency is updated... do we want to prevent that? do we not?).

Not really sure where I stand, but one thing is true: locked dependencies, like they're currently set for runtime, *need* to be explicitly updated. Maybe, we can go for latched runtime dependencies and locked development dependencies (ensuring, thus, that every contributor uses the same exact test/check/doc tools?).

Where should this stand?

Issue #746, opened 2019-01-23
macOS | pygame zero | Fails running with non-ASCII comment


tmontes:
## Preabmle

* This is not *strictly* a Mu bug.
* I narrowed it down to a Pygame Zero behaviour that could/should be improved:
  * I'll log and link an issue in the respective issue tracker.
* Nevertheless, once fixed in Pygame Zero, Mu distributions will need to bundle an updated version.
* I also decided to log the issue here to share my analysis and workaround with other Mu users.

## Platform Notes

* Does fail on macOS.
* Does not fail on Windows.
* Not sure if it fails on Linux, but I guess it might not.

## Steps to Reproduce on macOS

* Launch Mu, installed from the [distributed DMG](
https://github.com/mu-editor/mu/releases/download/1.0.2/mu-editor_1.0.2_osx.dmg), via the Finder or LaunchPad.
  (launching from the terminal does not lead to the failure, explanation below)
* Set Mu mode to **Pygame Zero**.
* Create new file with:

```python
WIDTH = 300
HEIGHT = 200
```

* Save it as "pgz_failure.py".
* Click **Play**.
* Pygame Zero runs the file successfully.
* Click **Stop**.
* Edit the "pgz_failure.py" file by adding a comment containing non-ASCII characters like:

```python
# Résumé
WIDTH = 300
HEIGHT = 200
```

* Save and click **Play**.
* Running fails with the following output on the lower *Running* pane:

```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "/Users/travis/build/mu-editor/mu_portable_python_macos/python/lib/python3.6/runpy.py", line 193, in _run_module_as_main
  File "/Users/travis/build/mu-editor/mu_portable_python_macos/python/lib/python3.6/runpy.py", line 85, in _run_code
  File "/Applications/Local/mu-editor.app/Contents/Resources/app_packages/pgzero/__main__.py", line 3, in <module>
    main()
  File "/Applications/Local/mu-editor.app/Contents/Resources/app_packages/pgzero/runner.py", line 77, in main
    src = f.read()
  File "/Users/travis/build/mu-editor/mu_portable_python_macos/python/lib/python3.6/encodings/ascii.py", line 26, in decode
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 3: ordinal not in range(128)


---------- FINISHED ----------
exit code: 1 status: 0
```

[...]

PR #937, opened 2019-10-09
Update deps and fix Black

ntoll:
This branch introduces two changes:

1) The package dependencies in `setup.py` have all been updated to the latest versions as of time of writing. Addresses #893.

2) As per suggestion by @carlosperate, `make check` now runs with `black --check` so will error if the code hasn't been tidied. Use `make tidy` to automatically tidy the code. See discussion in #929 for context.

### DIFF
--- a/make.py
+++ b/make.py
@@ -167,11 +167,33 @@ def tidy():
 
 
 @export
+def black():
+    """Check code with the 'black' formatter.
+    """
+    print("\nTidy")
+    for target in [
+        "setup.py",
+        "win_installer.py",
+        "make.py",
+        "mu",
+        "package",
+        "tests",
+        "utils",
+    ]:
+        return_code = subprocess.run(
+            [TIDY, "--check", "-l", "79", target]
+        ).returncode
+        if return_code != 0:
+            return return_code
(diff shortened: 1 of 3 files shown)


tmontes:
Somewhat on a tangent, `pgzero` has a fix for #746 in [this issue](https://github.com/lordmauve/pgzero/issues/173) and [this commit](https://github.com/lordmauve/pgzero/commit/97c989dda2c7d82260551a6cc5762ed0248d91a3)... It would be wonderful if we could help @lordmauve publish an updated release and integrate it into the next Mu version. :)

make.py
```diff
@@ -166,12 +166,34 @@ def tidy():
     return 0
 
 
+@export
+def black():
+    """Check code with the 'black' formatter.
+    """
+    print("\nTidy")
```
carlosperate:
```suggestion
    print("\nBlack")
```

carlosperate:
The black changes have been added in isolation in https://github.com/mu-editor/mu/pull/1113, and the dependency updates should be done after we have some of the other PRs merged first. So we can close this one.

PR #943, opened 2019-10-16
Adjust 'black' code to support version 18 as well

keith-packard:
Version 18 of 'black' has a slightly different API; check which
version we've managed to load and adapt on the fly

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1615,14 +1615,27 @@ class Editor:
         # Only works on Python, so abort.
         if tab.path and not tab.path.endswith(".py"):
             return
-        from black import format_str, FileMode, PY36_VERSIONS
+        from black import format_str, FileMode
 
         try:
             source_code = tab.text()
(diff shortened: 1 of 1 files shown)


ntoll:
Hi @keith-packard..! Thank you for the contribution. Just so we get some context here, why does Mu need to support version 18 when version 19 is already released..?

keith-packard:
> Hi @keith-packard..! Thank you for the contribution. Just so we get some context here, why does Mu need to support version 18 when version 19 is already released..?

Debian stable ships version 18, and I need to use that release. With this change, mu runs great.

Note that setup.py requires version 18, not 19, so this change is required to actually work when that is the version available.

carlosperate:
Sorry Keith! I missed this PR when creating https://github.com/mu-editor/mu/pull/1519. I'll close this one the other has already been merged.

Thanks again for the contribution though!

Issue #909, opened 2019-07-31
Travis OS X builds broken due to docutils version conflict


keith-packard:
I submitted pull request #908 yesterday and the Travis OS X build failed due to a conflicting
dependency on docutils:

mu-editor → sphinx → docutils >= 0.12
        gets docutils 0.15.2

mu-editor → briefcase == 0.2.9 → boto3 >=1.4.4 → botocore <1.13.0,>=1.12.99 (1.12.199) → docutils <0.15,>=0.10
        conflicts with 0.15.2

Here's the full log:
        https://travis-ci.org/mu-editor/mu/jobs/565847853

I kludged around this in my pull request by adding an explicit dependency on docutils<0.15, but I suspect someone with more knowledge about python packaging would have a better fix.

tmontes:
@keith-packard,

I found this intriguing and investigated a bit: dependency conflicts are things we all want to avoid! :) Here are my preliminary findings and thoughts:

* The Travis test script is [installing "everything"](https://github.com/mu-editor/mu/blob/acf9cb00893d776b5fe52d1c8c4fb9ab04c48503/.travis.yml#L53)  before running tests on macOS which, clearly, is not needed. If it only installed Mu and the test dependencies (via a `pip install .[tests]`, for example), neither the "sphinx -> docutils >= 0.12" dependency would be brought in (used for doc building), nor the "briefcase == 0.2.9 -> ... -> docutils <0.15,>=0.10" would (used for macOS packaging).
* On the other hand, I found it curious that botocore was explicitly depending on "docutils < 0.15". I found [this recent discussion](https://github.com/boto/botocore/pull/1788) around the topic.

IIUC, if we did nothing, someday in the (hopefully) near future, the "briefcase" dependency path will unlock more recent docutils version and things will eventually be fine. However, I guess we could improve (and optimize) the Travis test script not to install unneeded things in its preparation to run tests.

Thoughts? :)

keith-packard:
It's a bit disturbing to find a ceiling for any dependency as that could pin packages behind security or stability fixes. Not sure how to deal with this in the context of mu-editor though.

I think it's probably good that Travis caught this now, rather than discovering this when trying to release, so I wouldn't jump to trying to work around this by reducing the work done during the Travis build.

tmontes:
> It's a bit disturbing to find a ceiling for any dependency as that could pin packages behind security or stability fixes. Not sure how to deal with this in the context of mu-editor though.

[...]

PR #952, opened 2019-11-06
Remove repl and plotter panes when device disappears

keith-packard:
The first patch avoids having the view and mode state out-of-sync when the view closes the repl and plotter panes in response to the serial device disappearing. The alternative would be to have the mode in control of closing the repl/plotter panes when the device disappears, which would be more complicated to implement, but might be preferred.

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -307,6 +307,7 @@ class Window(QMainWindow):
     timer = None
     usb_checker = None
     serial = None
+    serial_name = None
     repl = None
     plotter = None
     zooms = ("xs", "s", "m", "l", "xl", "xxl", "xxxl")  # levels of zoom.
@@ -519,6 +520,7 @@ class Window(QMainWindow):
         """
         self.input_buffer = []
         self.serial = QSerialPort()
+        self.serial_name = port
         self.serial.setPortName(port)
         if self.serial.open(QIODevice.ReadWrite):
             self.serial.setDataTerminalReady(True)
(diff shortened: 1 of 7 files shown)


keith-packard:
travis is failing -- I think this is due to #909

dybber:
Hi @keith-packard, sorry for not getting to this earlier. Things have been moving slowly with Mu lately, but we are now back and actively working towards a new release.

Thanks for the contribution, but I'm not entirely sure, that we would want to do what you suggest. What if the user needs something in the REPL, e.g. an error message or some data?

keith-packard:
> Hi @keith-packard, sorry for not getting to this earlier. Things have been moving slowly with Mu lately, but we are now back and actively working towards a new release.
> 
> Thanks for the contribution, but I'm not entirely sure, that we would want to do what you suggest. What if the user needs something in the REPL, e.g. an error message or some data?

Glad your finding some time to work on Mu again! 

My issue was that students were confused by the open REPL window that didn't do anything when the device was detached. I thought that having the REPL disappear when the device disconnected would provide good visual feedback. I didn't think about having data lost in the REPL window as that isn't an important part of our curriculum at this point.

It's been a while since I used this (classes having been canceled since march), but as I recall, you would disconnect and then re-connect the device and the REPL window would not start working again and that *really* confused the students. If so, then just making the REPL window automatically recover when the device is re-connected would probably be sufficient, and would avoid losing any information remaining from the last REPL session.

I'll find some time and see if I can't make it do that instead of just closing.

dybber:
Making Mu better at reestablishing the connection seems like a good idea, my students can also get confused as the situation is now, and I also have to tell them individually to try and close and reopen the REPL to try to reconnect.

You might find it easier to start from the newest version from master, as I've made a complete rewrite of the device-related code.

keith-packard:
This was fixed by 4ec70b4c3d1647351a1242c936ddf66428019671, closing this request

Issue #921, opened 2019-09-23
ui is quite small


ak-is-rolling:
i just installed MU on my surface pro 6
i was about use it as it use it at school but in school the ui seems to be alot bigger than usal.
it not too bad but its a bit annoying and it hurts after long time sqwinting at it[
![Annotation 2019-09-23 201638](https://user-images.githubusercontent.com/55714635/65455418-19488600-de3f-11e9-96d5-7b38de744bad.png)
](url)

many thanks 
Alex

ZanderBrown:
In school being:

A) A different device
B) The same surface, just in a different building

I'm also slightly confused, you say you have to squint (and in your screenshot it is quite small) but say the UI is "alot bigger"?

I seem to remember the Surface has a HiDPI display so it's possible we aren't scalling correctly

ak-is-rolling:
A) different device
is there a way to change it 
it is just that it looked like this
![mu](https://user-images.githubusercontent.com/55714635/65457580-cf15d380-de43-11e9-8784-2d7ecaff9b62.gif)

ZanderBrown:
So the Surface is a misdirect :-)

I know that GIF is from the website, I'm assuming your screenshot is from the school device.

What can you tell us about the school device?

ak-is-rolling:
the screen shot is the surface. the gif is what i usually expect at school. we have basic hp nuc's, i expect nomore than a pentium. i think the moniters are 1920 x1080 p.

ZanderBrown:
Okay no I'm really confused, it the problem at school or on the Surface

It _sounds_ like your having a problem at school but based on that screenshot there's an issue on the Surface

ak-is-rolling:
the surface

ZanderBrown:
School was the mislead this whole time?? That's a plot twist :-)

So the problem is that the buttons are _smaller_ than they should be on the Surface?

If so I return to my earlier hypothesis that we aren't handling HiDPI properly

ak-is-rolling:
im sorry about that. im amsuming that HiDPi means somehting to do wiht the high pixel denisty of the screen. would this be something looked into for next release or not?

ZanderBrown:
Well given how increasingly common HiDPI is it would be something to prioritise really

Yes, HiDPI or "high DPI" is where a display has an especially high pixel density, as happens when you have a high resolution display in a small size (such as on the Surface)

ntoll:
Hi folks... I love the epic thread you two are having... :-)

[...]

PR #955, opened 2019-11-21
Scale ButtonBar fonts and icons based on zoom level, not window size

keith-packard:
This makes the whole UI get bigger or smaller depending on the zoom
level. Icons are scaled by the screen resolution as well so that they
will always appear in the same size relative to all fonts.

This provides user control over appearance, which should make the
interface more accessible for users with low vision and across a
wide range of displays.


dybber:
I just tried your suggestion locally. With the suggested changes, the entire row of buttons switches to the left, when you Zoom in. I would typically like to increase the font size 3 steps (e.g. when teaching connected to a projector), and would then do 3 taps at the Zoom-in button. With the changes here I will have to do a sequence of: 
 - 1 tap
 - move the mouse to the right to find the button
 - 1 tap
 - move the mouse to the right to find the button
 - 1 tap.

A middle way could be to only increase the font-size, not the icon size of the buttons. The buttons are in general very large already, but I can see that the button labels doesn't increase following the zoom-level, which I guess might be the real problem. It will still move a bit to the right if we only increase font sizes on the buttons, but not as much as when also the icons are increased in size.

keith-packard:
> With the changes here I will have to do a sequence of:
> 
>     * 1 tap
> 
>     * move the mouse to the right to find the button
> 
>     * 1 tap
> 
>     * move the mouse to the right to find the button
> 
>     * 1 tap.

Yeah, the user experience with this change really suffers because of this. I made this change not to *increase* the icon size on higher resolution devices, but to *decrease* the icon size on the tiny laptops I'm using for instruction, which have a 1024x768 screens where the icons cover quite a bit of the screen.

If the zoom buttons were stacked on the far right side of the toolbar, then they wouldn't move when sizes were changed; does that seem like a plausible work-around?

dybber:
Oh, I see! 

Perhaps then, the real problem is not even with the Zoom-buttons, but that all the default sizes are described in pixels (absolute sizes), where they should be specified in a unit that's independent of the DPI of the devices (relative sizes). E.g. percentage of the viewport width or relative to the system font-size.

That might be a much bigger change to make, but we also have issues in the other end with HiDPI screens: https://github.com/mu-editor/mu/issues/921 Perhaps @ntoll or @ZanderBrown can chime in.

ZanderBrown:
> independent of the DPI of the devices

[CSS pixels](https://drafts.csswg.org/css-values-3/#visual-angle-unit) != real (device) pixels, if Qt/Win32 is doing HiDPI support right it scales itself by magic:tm:

Of course if it isn't (as seen in some other issues) then it won't, but font relative sizes won't save you either

> percentage of the viewport width

Not supported by QSS (only px, pt, em, ex)

dybber:
Okay, thanks. Haven't done much with CSS since 2006, so I'm a bit rusty!

But apparantly it seems like it doesn't work in the other direction, scaling the icons down on small displays. Perhaps @keith-packard, you could provide us with a screenshot on a 1024x768 resolution?

keith-packard:
Here are two screen shots of a 1024x768 screen. This is running under Linux, where Qt scales the screen based on the DPI setting for fonts (QDesktopWidget().logicalDpiX()). I've set that to 96 for these examples.

 * Current fixed-size icons.
 mu-icon-fixed
 * Icons scaling based on zoom level. 
mu-icon-scaling

As you can see, with the fixed size icons, the button bar consumes a lot of screen space (107 lines out of 768 available). The scalable bar uses 84 lines. Also, when I add a couple more buttons for a target-specific mode, the buttons no longer fit in the bar, which hides them in an overflow area.

I just had an idea though -- instead of scaling these sizes based on the screen size, we can scale them based on this logical DPI value. That way the icons stay the same size when zooming in/out (avoiding the moving icon issue), but are smaller on low DPI devices and large on hi DPI devices. And, the user can select how big they are by changing that DPI value (which is configurable under Linux and Windows, I don't know about Mac OS X). I've attached that patch to this message for comments. When you change that DPI value, the set of available zoom levels changes to match.

[0001-Set-button-sizes-based-on-logical-DPI-of-monitor.txt](https://github.com/mu-editor/mu/files/5368965/0001-Set-button-sizes-based-on-logical-DPI-of-monitor.txt)

PR #957, opened 2019-11-24
Update .travis.yml for current Python versions

cclauss:
Tracks with https://devguide.python.org/#branchstatus

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,9 @@ jobs:
     - python: 3.8
     - os: osx
   include:
-    - python: 3.8
+    # TODO: When a production verion of pygame is compatible with Python 3.8,
+    # allow_failures above and the pip install pygame below should be removed.
+    - python: 3.8  # When a production version of 
       before_install: pip install pygame==2.0.0.dev6
     - os: osx
       osx_image: xcode11.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ python:
 jobs:
   allow_failures:
     - python: 3.8
-    - os: osx
   include:
     # TODO: When a production verion of pygame is compatible with Python 3.8,
     # allow_failures above and the pip install pygame below should be removed.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,70 +1,57 @@
 # Travis can building for Linux and macOS
-matrix:
-  include:
-    - os: linux
-      dist: xenial 
-      sudo: required
-      language: python
-      python: 3.5
-    - os: linux
-      dist: xenial 
-      sudo: required
-      language: python
-      python: 3.6
-    - os: linux
-      dist: xenial
-      sudo: required
-      language: python
-      python: 3.7
-    # To maximise compatibility pick earliest image, OS X 10.12
-    - os: osx
-      osx_image: xcode8.3
-      sudo: required
-      language: generic
-      python: 3.6
+language: python
+addons:
+  apt:
+    update: true
+    packages: libxkbcommon-x11-0  # Linux package needed for Qt to work
+  homebrew:
+    update: true
+    packages: freetype
 
-before_install:
-  # OS and default Python info
-  - uname -a
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then sw_vers; fi
-  # Linux packages needed for Qt to work.
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install libxkbcommon-x11-0; fi
-  # Python 3 installation required
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then bash package/install_osx.sh; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then pyenv install 3.6.5; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then pyenv versions; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then pyenv global 3.6.5; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then eval "$(pyenv init -)"; fi
-  - pip install --upgrade pip setuptools
-  # Check everything was correctly installed
-  - echo $PATH
-  - python --version
-  - python -c "import struct; print(struct.calcsize('P') * 8)"
-  - python -c "import sys; print(sys.executable)"
-  - python -m pip --version
-  - pip --version
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then virtualenv venv --python=python3; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source venv/bin/activate; fi
+services: xvfb
+python:
+  - 3.5
+  - 3.6
+  - 3.7
+jobs:
+  allow_failures:
+    - python: 3.8
+    - os: osx
+  include:
+    - python: 3.8
+      before_install: pip install pygame==2.0.0.dev6
+    - os: osx
+      osx_image: xcode11.2
+      language: generic
+      python: 3.7
+      before_install: ln -s /usr/local/bin/pip3 /usr/local/bin/pip
 
 install:
-  # Install Mu and its dependencies
-  - pip install .[dev]
+  # OS and default Python info
+  - uname -a ; python3 --version ; pip3 --version ; sw_vers || true
+  - pip3 install --upgrade pip setuptools
   # Check everything was correctly installed
-  - pip freeze
+  - echo $PATH
+  - python3 -c "import struct; print(struct.calcsize('P') * 8)"
+  - python3 -c "import sys; print(sys.executable)"
+  # Install Mu and its dependencies
+  - pip3 install .[dev]
+  # Check everything was correctly installed
+  - pip3 freeze
 
 script:
   # For Linux run only the tests on 3.5 and the full checker elsewhere, all need the "X Virtual Framebuffer"
-  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$TRAVIS_PYTHON_VERSION" = "3.5" ]; then xvfb-run make coverage; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$TRAVIS_PYTHON_VERSION" != "3.5" ]; then xvfb-run make check; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$TRAVIS_PYTHON_VERSION" = "3.5" ]; then make coverage; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$TRAVIS_PYTHON_VERSION" != "3.5" ]; then make check; fi
   - make clean
-
   # Run the tests on macOS and package it: "make macos" runs checks+tests first.
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then make macos; fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -54,17 +45,15 @@ install:
   - pip freeze
 
 script:
-  # For Linux run only the tests on 3.5 and the full checker elsewhere, all need the "X Virtual Framebuffer"
-  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$TRAVIS_PYTHON_VERSION" = "3.5" ]; then xvfb-run make coverage; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$TRAVIS_PYTHON_VERSION" != "3.5" ]; then xvfb-run make check; fi
```
carlosperate: The tests are un as part of `make macos` below, but removing these two lines basically removes running the tests on Linux, so I think they need to stay.

.travis.yml
```diff
@@ -1,51 +1,42 @@
...
+python:
+  - 3.5
+  - 3.6
+  - 3.7
+addons:
+  apt:
+    update: true
+    packages:  # Linux packages needed for Qt to work.
+    - libxkbcommon-x11-0
+services: xvfb
```
carlosperate: Nice! I didn't realised this was option, this will help keep the rest of the script more generic between Linux and macOS.

.travis.yml
```diff
@@ -1,51 +1,42 @@
...
+language: python
+python:
+  - 3.5
+  - 3.6
+  - 3.7
+addons:
+  apt:
+    update: true
+    packages:  # Linux packages needed for Qt to work.
+    - libxkbcommon-x11-0
```
carlosperate: Tiny nitpick, but should this be nested one more level?

cclauss:
Travis CI macOS test fails on https://travis-ci.org/mu-editor/mu/jobs/616403598#L610
https://github.com/mu-editor/mu/blob/master/tests/interface/test_panes.py#L1232

cclauss:
Travis CI macOS test fails https://travis-ci.org/mu-editor/mu/jobs/616403598#L610 on
https://github.com/mu-editor/mu/blob/master/tests/interface/test_panes.py#L1232

cclauss:
OK...  Ready for review.  Python 3.5, 3.6, and 3.7 are working on Linux.  Windows is working. 
macOS has a flakey test that passes sometimes and fails sometimes.  Python 3.8 is not working yet.

.travis.yml
```diff
@@ -1,70 +1,57 @@
...
-      language: python
-      python: 3.6
-    - os: linux
-      dist: xenial
-      sudo: required
-      language: python
-      python: 3.7
-    # To maximise compatibility pick earliest image, OS X 10.12
+    - python: 3.8
+      before_install: pip install pygame==2.0.0.dev6
```
tmontes: Wondering if we could declare this in `setup.py` instead, where all dependencies are set. Otherwise, we risk going back to scattered around `pip installs` for dependencies which, in the long run, make life more difficult. ;)

.travis.yml
```diff
@@ -1,70 +1,57 @@
...
     - os: osx
-      osx_image: xcode8.3
-      sudo: required
+      osx_image: xcode11.2
       language: generic
-      python: 3.6
+      python: 3.7
+      before_install: ln -s /usr/local/bin/pip3 /usr/local/bin/pip
```
tmontes: Could you please clarity the intent here? Why would we "write" to the test operating environment?

package/install_osx.sh
```diff
@@ -4,6 +4,6 @@ brew update >/dev/null 2>&1  # This produces a lot of output that's not very int
 
 # Install Python 3.6
 #brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f2a764ef944b1080be64bd88dca9a1d80130c558/Formula/python.rb
-brew upgrade pyenv 
+brew upgrade pyenv || true
```
tmontes: Can you please clarify this? Is `pyenv` misbehaving somehow, by returning non-successful exit statuses when it should return success? Something else? I tend to feel that `|| true` in shell scripts are a lot like `try/catch: pass` code -- something you don't really want, that is bound to hide away future unexpected failures... Thoughts? :)

.travis.yml
cclauss: Python 3.8 is purely experimental here (jobs: allow_failures: - python: 3.8) and I felt is right to keep all the experimentation in one place.  We do not support Py38 in setup.py's trove classifiers and putting a ___dev___ version of pygame into setup.py did not seem right.

.travis.yml
cclauss: This file was using pip in some places and pip3 in others.  https://snarky.ca/why-you-should-use-python-m-pip/

I am just trying to ensure in across all OSes that pip3 is the pip that is bound to the Python we are testing with.

package/install_osx.sh
cclauss: brew will throw an error if pyenv is already at the latest version.

.travis.yml
tmontes: Understood. I would then add a VERY VISIBLE comment saying something like that and advising moving such dependencies into `setup.py` when the time comes. :)

.travis.yml
tmontes: Thanks for the link to Brett's article: I'm very familiar with the matters.

If we were to be strict, in the sense of being certain that we're using the correct Python version, then no `pip` / `pip3` would be used.
Given that `python3 --version` is run early, letting us know which exact version is CI running then, strictly, and per Brett's article, which I do agree with, we would go `python3 -m pip` from then on... :)

What I'm wondering is why no such `ln` seems to be run under `linux`. But I'm far from a Travis expert. Feel free to instruct me, please! :)

package/install_osx.sh
tmontes: (preliminary note: I stand corrected, the eventual misbehaviour I was referring to was `brew`'s, of course!)

Are you sure? This is not what I am observing on my system (admittedly running a somewhat oldish macOS version, 10.12.6):
```
$ brew list | grep pyenv
pyenv
$ brew update
Already up-to-date.
$ brew upgrade pyenv
(big warning about older macOS version)

Warning: pyenv 1.2.15 already installed
$ echo $?
0
$
```

I somehow insist on this because I'd prefer CI to fail as early as possible, and going with `|| true` means that any transitory but real `brew` failure at this step will be ignored and, most likely, lead to job execution failure at a later point. :)

.travis.yml
```diff
@@ -1,70 +1,57 @@
...
+
+services: xvfb
+python:
+  - 3.5
+  - 3.6
+  - 3.7
+jobs:
+  allow_failures:
+    - python: 3.8
+    - os: osx
```
tmontes: I suppose I understand the motivation to go `allow_failures` on macOS. But I'd rather have CI fail from time to time and remind us of something that needs fixing. Alternatively, we risk shipping something that does not work for a subset of the users because it is not "properly" tested. Feelings?

PS: And I've been bit by such flaky-ness a few times in the past. ;)

carlosperate: > I suppose I understand the motivation to go allow_failures on macOS

Not sure I understand it though, is it the tests flakiness?

.travis.yml
```diff
@@ -1,70 +1,57 @@
...
-      osx_image: xcode8.3
-      sudo: required
+      osx_image: xcode11.2
       language: generic
-      python: 3.6
+      python: 3.7
+      before_install: ln -s /usr/local/bin/pip3 /usr/local/bin/pip
 
-before_install:
+install:
```
carlosperate: I think the original intent here was to do all "info commands" in the `before_install`, and only the actually installation in the `install` entries.
Is there is a reason not to use `before_install`?
I don't mind either way, just wondering if there is something I wasn't aware of.

.travis.yml
```diff
@@ -1,70 +1,57 @@
...
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then zip -r -X dist/mu-editor.zip macOS/mu-editor.app; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then du -sk dist/; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then mv dist/mu-editor.zip dist/mu-editor_$(date '+%Y-%m-%d_%H_%M')_${TRAVIS_BRANCH}_${TRAVIS_COMMIT:0:7}.zip; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then
+      make macos;
+      mkdir dist;
+      zip -r -X dist/mu-editor.zip macOS/mu-editor.app;
+      du -sk dist/;
+      mv dist/mu-editor.zip dist/mu-editor_$(date '+%Y-%m-%d_%H_%M')_${TRAVIS_BRANCH}_${TRAVIS_COMMIT:0:7}.zip;
+    fi
```
carlosperate: One advantage of having these as individual entries, is that Travis lets you expand/collapse them, and even when it doesn't it's easy to see the individual steps as they are separated by green text with the commands.

As an example, if I wanted to see the output of `du -sk dist` I have to read the thousands of lines outputted by this single entry, as even if I search for the text `du -sk dist` I wont' find the specific output entry.

Makefile
```diff
@@ -106,7 +106,7 @@ win64: check
 
 macos: check
 	@echo "\nPackaging Mu into a macOS native application."
-	python setup.py macos --support-pkg=https://github.com/mu-editor/mu_portable_python_macos/releases/download/0.0.6/python3-reduced.tar.gz
+	python3 setup.py macos --support-pkg=https://github.com/mu-editor/mu_portable_python_macos/releases/download/0.0.6/python3-reduced.tar.gz
```
carlosperate: To be completely honest I'd rather modify the travis environment to ensure `python` points to the right version than to modify a single entry from the Makefile, as everywhere else in the file uses `python`.

package/install_osx.sh
carlosperate: @tmontes does your note at the top mean brew doesn't return 0 when the package is installed?

.travis.yml
```diff
@@ -1,70 +1,57 @@
...
-      sudo: required
-      language: python
-      python: 3.7
-    # To maximise compatibility pick earliest image, OS X 10.12
+    - python: 3.8
+      before_install: pip install pygame==2.0.0.dev6
     - os: osx
-      osx_image: xcode8.3
-      sudo: required
+      osx_image: xcode11.2
```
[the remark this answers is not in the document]
cclauss: Plenty of options in https://docs.travis-ci.com/user/reference/osx/#macos-version but if homebrew complains that brew and Apple no longer support a version of macOS then consideration should be given to why the mu-editor should support it.

.travis.yml
cclauss: When different OSes need different install steps, I tend to use __before_install__ for OS-dependent steps and __install__ for steps that are the same on all OSes.

.travis.yml
cclauss: The motivation for __allow_failures__ is explained at:
https://github.com/mu-editor/mu/pull/957#issuecomment-557967312 and https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail

.travis.yml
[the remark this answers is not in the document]
carlosperate: Oh yeah, absolutely we should change it from the old version, which I think was going to be turned off soon, just preferred not to move it all the way to the latest.

.travis.yml
carlosperate: But is that a test that only fails sometimes? As right now the macOS build is passing, right?

.travis.yml
carlosperate: Oh, that's a nice convention! Sounds good.

Issue #958, opened 2019-11-24
PyQtChart wheel change causes error in win_installer.py


carlosperate:
I've been looking at the current AppVeyor failures on recent PRs like this one, that haven't touched the CI config:

https://ci.appveyor.com/project/carlosperate/mu/builds/29084193

I'm writing here my current findings as I might not be able to raise a PR today, but maybe somebody else could continue looking into it. 

The error:

```
Checking for wheel availability at PyPI.
...
- PyQtChart==5.12.0 Traceback (most recent call last):
  File "win_installer.py", line 300, in <module>
    run(bitness, repo_root)
  File "win_installer.py", line 264, in run
    installer_exe = create_pynsist_cfg(venv_python, repo_root, pynsist_cfg)
  File "win_installer.py", line 196, in create_pynsist_cfg
    wheels = pypi_wheels_in(requirements)
  File "win_installer.py", line 155, in pypi_wheels_in
    if any(r.package_type == "wheel" for r in releases):
TypeError: 'NoneType' object is not iterable
```

The code:
https://github.com/mu-editor/mu/blob/ee600ff16753194f33e39975662aba438a8f5608/win_installer.py#L152-L155

Neither the `win_installer.py` code or the dependency version have changed for quite a long time, so something external had to change to produce this error now.

I tried to see the output from that snippet of code locally and indeed we are getting a `None` value returned:

```python
>>> import yarg
>>> package = yarg.get("PyQtChart")
>>> package.release("5.12.0")
>>>
```

Tracked the last successful build and I notice that the `pip freeze` output for PyQtChart has changed!
From `5.12` (as indicated in setup.py) to `5.12.0`.
https://ci.appveyor.com/project/carlosperate/mu/builds/28868244

And that does work:

```python
>>> package.release("5.12")
[<Release 5.12>, <Release 5.12>, <Release 5.12>, <Release 5.12>]
```

I checked the version of pip in case something has changed in the `freeze` command, but we are using the same version in both builds.

So then I check PyPI and found something fishy:

<img width="1140" alt="Screenshot 2019-11-24 at 20 55 57" src="https://user-images.githubusercontent.com/4189262/69501787-30b7f400-0f00-11ea-9d7a-c20bcb49102d.png">

Version 5.12 was released in February, yet the wheels were uploaded 2 days ago? 🤔

### Conclusion

[...]

PR #959, opened 2019-11-25
setup.py: Upgrade dependencies to fix AppVeyor and Travis CI macOS

cclauss:
A subset of #957 that gets the CI green again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,15 +21,15 @@ with open(os.path.join(base_dir, "CHANGES.rst"), encoding="utf8") as f:
 
 
 install_requires = [
-    'PyQt5==5.12.1;"arm" not in platform_machine',
-    'QScintilla==2.11.1;"arm" not in platform_machine',
-    'PyQtChart==5.12;"arm" not in platform_machine',
+    'PyQt5==5.13.2;"arm" not in platform_machine',
+    'QScintilla==2.11.3;"arm" not in platform_machine',
+    'PyQtChart==5.13.1;"arm" not in platform_machine',
     # `flake8` is actually a testing/packaging dependency that, among other
     # packages, brings in `pycodestyle` and `pyflakes` which are runtime
     # dependencies. For the sake of "locality", it is being declared here,
     # though. Regarding these packages' versions, please refer to:
     # http://flake8.pycqa.org/en/latest/faq.html#why-does-flake8-use-ranges-for-its-dependencies
-    "flake8 >= 3.7.8",
+    "flake8 >= 3.7.9",
     "pycodestyle >= 2.5.0, < 2.6.0",
     "pyflakes >= 2.1.0, < 2.2.0",
     "pyserial==3.4",
@@ -38,7 +38,7 @@ install_requires = [
     "appdirs>=1.4.3",
     "semver>=2.8.0",
     "nudatus>=0.0.3",
-    'black>=18.9b0;python_version > "3.5"',
+    'black>=19.10b0;python_version > "3.5"',
     "Flask==1.0.2",
     "python-dateutil==2.8.0",
 ]
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,7 @@ install_requires = [
     "nudatus>=0.0.3",
     'black>=18.9b0;python_version > "3.5"',
     "Flask==1.0.2",
+    "python-dateutil==2.8.0",
 ]
 
 


tmontes:
@cclauss,

Thanks for the PR, this relates to #893 so leaving this note here to link things around. :)
There's an outstanding decision around "locking" vs "latching" vs "nothing" with regards to dependency versions there which I feel needs consideration and decision.

Nevertheless, I feel your PR should be merged. There are a few things I don't understand though:

* Why are you changing `flake8`'s and `black`'s latching?
* More importantly, why is `python-datetutil` being explicitly added and locked to that specific version?

Standing by.

cclauss:
The first commit was called __setup.py: python-dateutil==2.8.0 to fix appVeyor__.

On flake8 and black, I saw that we were passing tests with the latest and greatest so I "latched them up".  In general, I tend to pin (lock) runtime dependencies but let testing tools (pytest, flake8, black) use the latest and greatest so that as they add more tests, we test our code more.  Definitely a matter of personal taste.

tmontes:
> The first commit was called **setup.py: python-dateutil==2.8.0 to fix appVeyor**. It will fail without pinning.

You mislead me here: locking `python-dateutil` addresses the macOS Travis job failure, in the packaging stage, not AppVeyor (that failure is explained in #958 and fixed by #961).

The unfortunate motive for that is that, on one hand, Mu depends on `qtconsole==4.4.3`:
* ...that depends on `jupyter-client>=4.1`...
* ...that depends on `python-dateutil>=2.1`.
* This leads `pip` to select `python-dateutil==2.8.1` for installation.

But then, on the other hand, Mu also depends on `briefcase==0.2.9` for macOS packaging:
* ...that depends on `boto3>=1.4.4`...
* ...that depends on `botocore<1.14.0,>=1.13.28`...
* ...that very recently, [in this commit](https://github.com/boto/botocore/commit/e87e7a745fd972815b235a9ee685232745aa94f9), updated its `python-dateutil` dependency to `python-dateutil<2.8.1,>=2.1`.
* At this stage `pip` emits a warning as it was the case [in this build line](https://travis-ci.org/mu-editor/mu/jobs/618370793#L665), that I quote here:
```
ERROR: botocore 1.13.28 has requirement python-dateutil<2.8.1,>=2.1; python_version >= "2.7",
but you'll have python-dateutil 2.8.1 which is incompatible.
```

PS: Yes, `pip` should be more clever about it. But it isn't. Yet. The PSF is promoting such improvement, which is a good thing. :)

With this in mind, and without diving into the larger discussion of #893 (what to lock, what to latch, etc), I would suggest:

* Bring the `python-dateutil==2.8.0` line further up, maybe right before the `qtconsole` one.
* Precede it with a comment explaining the motive for such apparently out of context lock.

What do you say? Would you improve it in some other way? :)

> On flake8 and black, I saw that we were passing tests with the latest and greatest so I "latched them up". In general, I tend to pin (lock) runtime dependencies but let testing tools (pytest, flake8, black) use the latest and greatest so that as they add more tests, we test our code more. Definitely a matter of personal taste.

That sounds perfectly reasonable. Ultimately I do not think it comes down to a matter of personal taste though. :) I'll be glad to discuss the topic, and coming up with a strategy and policy in this regard that better serves Mu, its contributors and, in the end, all Mu users. #893 is waiting for us there! :)

ntoll:
Hi folks, I intend to deep dive into all things packaging related this week, including ensuring CI things are green -- thank you so much for the work done so far. I'll make sure I reference this ticket as I make progress (or not). ;-)

dybber:
The version conflict issues surrounding `python-dateutil` reappeared today, when looking at this again in #1065,  but with another package `docutils`, that also is brought in by briefcase 0.2.9 through its `boto3` dependency. The longterm solution will be to upgrade to briefcase 0.3.0, which doesn't have the dependency on boto3. I've made an issue #1066 on doing such an upgrade.

tjguk:
Thanks for the work here, @cclauss . I've just merged #1065 which *looks* as if it supersedes your patch here. Could you just confirm and either close (with our many thanks for taking the trouble) or let me know if there's still something which needs merging? Thanks

dybber:
We're getting warnings on pycodestyle and pyflakes, and flake8 brings them in, so it could be that we should also apply that change to master (i.e. remove the explicit dependencies on pycodestyle and pyflakes), however, the explicit python-dateutil pinning doesn't seem to be necessary anymore.

tjguk:
Ok; I've given it a quick test-run locally and I don't see any immediate issues. Let's merge it now and watch the builds

dybber:
I still don’t see the need for the python-dateutil dependency.

tjguk:
If it's not causing a problem building, I'll pull it out of my current dev branch and check that we're still ok...

PR #960, opened 2019-11-28
Docs: remove mbed driver reference from setup.rst

microbit-mark:
Windows 7 reaches EOL in Jan 2020 and Micro:bit Educational Foundation no longer want to recommend installing the mBED driver as it conflicts with later versions of Windows


carlosperate (approved):
Thanks Mark!

As the driver installer causes issues with WebUSB in Windows8.1+ it is also discouraged to install it unless you are on Windows 7, which as you mentioned has been EOL for a year and a half 👍

Issue #940, opened 2019-10-14
macOS Catalina | New packaging requirements?


tmontes:
Just saw this article passing through my regular RSS feeds:

* [Mac Python Distribution Post Updated for Catalina and Notarization](https://glyph.twistedmatrix.com/2019/10/catalina-notarize-python.html)...

Which essentially links to a Catalina updated version of a previous packaging article:

* [Tips And Tricks for Shipping a PyGame App on the Mac](https://glyph.twistedmatrix.com/2018/01/shipping-pygame-mac-app.html).

Both very well written by @glyph, clearly pointing to a "dragons ahead" warning WRT supporting macOS Catalina Mu users. The gist:

* New code signing requirement.
* Needs to be submitted to Apple, somehow.
* A new entitlements file seems to be needed.
* All executable files must now be signed.
* ...and more?

The article uses a different tool for the first packaging step (`py2app` where we use `briefcase`) but from a cursory read the signing / getting the packaged `.app` bundle ready for distribution seems tool agnostic.

PS: macOS Catalina has just been released, earlier this month. As time progresses, we can expect more and more Mu mac users to use Catalina. Actually, #936 is a report from a Catalina user -- wondering how Mu worked for the reporter, there.

glyph:
Always nice to hear that my writing is proving useful to people!  Just a few minor notes:

> * Needs to be submitted to Apple, somehow.

The "somehow" are the `altool` and `stapler` invocations in the library - there's no out-of-band communication other than the email notifying you that you *can* run `stapler`.

> * ...and more?

This should really be it :)

> The article uses a different tool for the first packaging step (`py2app` where we use `briefcase`) but from a cursory read the signing / getting the packaged `.app` bundle ready for distribution seems tool agnostic.

Your reading here is correct.  The `zip` / `codesign` / `altool` / `xcrun` invocations are fairly agnostic to how the bundle was actually built; these ought to work with non-Python stuff, for that matter, as long as they happen to require that one extra entitlement that Python needs.

> PS: macOS Catalina has just been released, earlier this month. As time progresses, we can expect more and more Mu mac users to use Catalina. Actually, #936 is a report from a Catalina user -- wondering how Mu worked for the reporter, there.

Upgrade curves on new macOS releases are shockingly fast; you're likely to see a *majority* of users on Catalina in the not too distant future.

tmontes:
Thanks for chiming in @glyph, and thanks for the article. We will soon fight this fight. :)

[...]

Issue #925, opened 2019-10-01
macOS packaged Mu has hardcoded "1.0" version information


tmontes:
## The issue

As we move towards a non "1.0" release, I realised that the macOS packaged Mu contains hardcoded metadata in what concerns its version.

In macOS, Mu is distributed as an *application bundle*, something that looks like a file but is actually a directory structure.

Using the **Finder** > **File** > **Get Info** (or hitting CMD-I on the keyboard) on such *application bundles* brings up a small informational window on whichever "file" is selected. When such "files" are *application bundles*, the application version is shown, as illustrated in this image:

<img width="400" alt="Screen Shot 2019-10-01 at 13 00 33" src="https://user-images.githubusercontent.com/1912206/65960071-836ab780-e44b-11e9-8952-3d08a129da9f.png">

Note how "Version: 1.0" is displayed. It should, instead, display whichever version is set in Mu's single source of truth for that -- the `mu/__init__.py` file -- much like the window title displays when Mu is running. In this case, "1.1.0.alpha.2".

## What I did so far

A cursory glance at the packaging process, using [`briefcase`](https://pypi.org/project/briefcase/) from the [BeeWare](https://beeware.org) project, shows no indication of something that we might be "doing wrong". Moreover, the `briefcase` docs don't seem to include any information related to setting the *application bundle* version.

Another quick glance at `briefcase`'s execution log and source points to something I wasn't aware of yet: the [`Python-macOS-template`](https://github.com/beeware/Python-macOS-template) project which seems to be a [`cookiecutter`](https://pypi.org/project/cookiecutter/) template. That seems to be the culprit, given that its `Info.plist` file includes a hard-coded `1.0` version [right here](https://github.com/beeware/Python-macOS-template/blob/8a0b98067e586e10aa9ece4d98b8a67904df4916/%7B%7B%20cookiecutter.dir_name%20%7D%7D/%7B%7B%20cookiecutter.formal_name%20%7D%7D.app/Contents/Info.plist#L20).

PS: The `Info.plist` file is a mandatory file in *application bundles* including, among other things, meta-information about the application such as the version string.

PPS: Another thing I checked was whether the Windows packaged Mu might have the same limitation. Apparently it does not: even though the file properties themselves do not seem to include a version (shows up as blank), the correct version is displayed in the installed software list/dialog.

## What next

Feels like both the `briefcase` and `Python-macOS-template` projects might need improvements to support fixing this on Mu's side.

[...]

PR #961, opened 2019-11-28
Windows packaging: exit early on PyPI lookup failures + sub-process failures.

tmontes:
Attempting to address the windows packaging failure in #958, originally identified in #959. 

Even though the [docs state otherwise](https://yarg.readthedocs.io/api-search.html#yarg.package.Package.release), [`yarg`](https://pypi.org/project/yarg)'s `Package.release(...)` method does not always return return a `list`. When not found, it returns `None`, which is not iterable.

### DIFF
--- a/win_installer.py
+++ b/win_installer.py
@@ -166,17 +166,17 @@ def pypi_wheels_in(requirements):
         package = yarg.get(name)
         releases = package.release(version)
         if not releases:
-            feedback = (
-                "WARNING: Just installed but not found at PyPI."
-                " (bad meta-data?)"
+            raise RuntimeError(
+                "ABORTING: Did not find {!r} at PyPI. (bad meta-data?)".format(
+                    requirement
+                )
             )
-        elif any(r.package_type == "wheel" for r in releases):
+        if any(r.package_type == "wheel" for r in releases):
             wheels.append(requirement)
             feedback = "ok"
         else:
             feedback = "missing"
         print(feedback)
-
     return wheels
 
 
--- a/win_installer.py
+++ b/win_installer.py
@@ -106,13 +106,26 @@ TKINTER_ASSETS_URLS = {
 }
 
 
+def subprocess_run(args):
+    """
+    Wrapper around subprocess.run: when the sub-process exits with a non-zero
+    return code, prints out a message and exits with the same code.
+    """
+    cp = subprocess.run(args)
+    try:
+        cp.check_returncode()
+    except subprocess.CalledProcessError as exc:
+        print(exc)
+        sys.exit(cp.returncode)
+
+
 def create_packaging_venv(target_directory, name="mu-packaging-venv"):
     """
     Creates a Python virtual environment in the target_directry, returning
     the path to the newly created environment's Python executable.
     """
     fullpath = os.path.join(target_directory, name)
-    subprocess.run([sys.executable, "-m", "venv", fullpath])
+    subprocess_run([sys.executable, "-m", "venv", fullpath])
     return os.path.join(fullpath, "Scripts", "python.exe")
 
 
@@ -265,12 +278,12 @@ def run(bitness, repo_root):
         venv_python = create_packaging_venv(work_dir)
 
         print("Updating pip in the virtual environment", venv_python)
-        subprocess.run(
+        subprocess_run(
             [venv_python, "-m", "pip", "install", "--upgrade", "pip"]
         )
 
         print("Installing mu with", venv_python)
-        subprocess.run([venv_python, "-m", "pip", "install", repo_root])
+        subprocess_run([venv_python, "-m", "pip", "install", repo_root])
 
         pynsist_cfg = os.path.join(work_dir, "pynsist.cfg")
         print("Creating pynsist configuration file", pynsist_cfg)
@@ -284,11 +297,11 @@ def run(bitness, repo_root):
         unzip_file(filename, work_dir)
 
         print("Installing pynsist.")
-        subprocess.run([venv_python, "-m", "pip", "install", PYNSIST_REQ])
+        subprocess_run([venv_python, "-m", "pip", "install", PYNSIST_REQ])
 
         mu_pynsist_script = os.path.join(repo_root, "package", "mu_nsist.py")
         print("Running custom pynsist script at", mu_pynsist_script)
-        subprocess.run([venv_python, mu_pynsist_script, pynsist_cfg])
+        subprocess_run([venv_python, mu_pynsist_script, pynsist_cfg])
 
         destination_dir = os.path.join(repo_root, "dist")
         print("Copying installer file to", destination_dir)
--- a/win_installer.py
+++ b/win_installer.py
@@ -139,29 +139,6 @@ def about_dict(repo_root):
     return about
 
 
-def pypi_wheel_release(name, version):
-    """
-    Returns a `name==version` string if there's such a wheel at PyPI. When
-    version ends with ".0", a shorter version may be returned, matching what's
-    found at PyPI (name=`PyQtChart` version=`5.12.0` returns `PyQtChart==5.12`,
-    as of this writing). When no releases are found, a ValueError is raised.
-    """
-    package = yarg.get(name)
-    while True:
-        releases = package.release(version)
-        if releases or not version.endswith(".0"):
-            break
-        version = version[:-2]
-
-    if not releases:
-        raise ValueError("No {n!r} package found at PyPI.".format(n=name))
-
-    if any(r.package_type == "wheel" for r in releases):
-        return "{n}=={v}".format(n=name, v=version)
-
-    return None
-
-
 def pypi_wheels_in(requirements):
     """
     Returns a list of the entries in requirements which are distributed as
@@ -173,19 +150,18 @@ def pypi_wheels_in(requirements):
     for requirement in requirements:
         name, _, version = requirement.partition("==")
         print("-", requirement, end=" ")
-
-        try:
-            found_wheel = pypi_wheel_release(name, version)
-        except ValueError as exc:
-            feedback = "ERROR: {}".format(exc)
+        package = yarg.get(name)
+        releases = package.release(version)
+        if not releases:
+            feedback = (
+                "WARNING: Just installed but not found at PyPI."
+                " (bad meta-data?)"
+            )
+        elif any(r.package_type == "wheel" for r in releases):
+            wheels.append(requirement)
+            feedback = "ok"
         else:
-            if found_wheel:
-                wheels.append(found_wheel)
-                feedback = "ok"
-                if found_wheel != requirement:
-                    feedback += " ({})".format(found_wheel)
-            else:
-                feedback = "missing"
+            feedback = "missing"
         print(feedback)
 
     return wheels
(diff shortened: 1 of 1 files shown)


tmontes:
Notes for reviewers:
* This PR addresses the Windows packaging failure identified in #957 / #959, and further detailed in #958. 
* It is not expected to pass on macOS. That is for an orthogonal motive and what lead the author or PR #959 to propose locking `python-dateutil==2.8.0` (see my comments there).

Due to the way `pynsist` works (see top comments on `win_installer.py`), I initially thought that the first commit would be enough: in other words, if `yarg` could not find a matching PyPI release, then don't add it as a wheel hoping that `pynsist` could just handle it by copying the associated package from the running environment -- that was not the case (maybe it could work for other, simpler packages? no idea!).

With that, I went for what I thought would be a reasonable generic stragety:
* Try to find a release on PyPI.
* If none was found and the version string ends with `.0`, try finding a release without the trailing `.0`.
* Repeat while no releases are found and the shrinking version string still ends with `.0`.

While not 100% intuitive, this change is generic, handles the current `PyQtChart` observed case, and leaves us with a better tool for such future mis-releases (if I may call them so) in PyPI.

win_installer.py
```diff
@@ -139,6 +139,29 @@ def about_dict(repo_root):
...
+    while True:
+        releases = package.release(version)
+        if releases or not version.endswith(".0"):
+            break
+        version = version[:-2]
+
+    if not releases:
+        raise ValueError("No {n!r} package found at PyPI.".format(n=name))
+
+    if any(r.package_type == "wheel" for r in releases):
```
carlosperate: I guess the original code already had this issue, but we are only detecting if there any *any* wheels here right?
So if we are building for Windows 32 bits, and only 64 bit wheels are provided (or no windows wheels), we would still think everything is okay.

I guess pynsist will probably check for the right wheels to be available when downloading them, and break at that point, and that to be fair that is fine. If that's the case there is no need to make this script more complicated.

win_installer.py
```diff
@@ -139,6 +139,29 @@ def about_dict(repo_root):
...
+        return "{n}=={v}".format(n=name, v=version)
+
+    return None
```
carlosperate: If we raise a valuerror here as well we can also simplify the code in `pypi_wheels_in`.

tmontes:
@carlosperate, thanks a lot for your review. After going through it, I do agree with most of it and will simplify things back. I'd still like improve the script in order to produce more helpful "this looks that it will fail" messages. :)

I'll comment further along with your code comments.

win_installer.py
```diff
@@ -139,6 +139,29 @@ def about_dict(repo_root):
...
+    """
+    Returns a `name==version` string if there's such a wheel at PyPI. When
+    version ends with ".0", a shorter version may be returned, matching what's
+    found at PyPI (name=`PyQtChart` version=`5.12.0` returns `PyQtChart==5.12`,
+    as of this writing). When no releases are found, a ValueError is raised.
+    """
+    package = yarg.get(name)
+    while True:
+        releases = package.release(version)
+        if releases or not version.endswith(".0"):
```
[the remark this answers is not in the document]
tmontes: In retrospect I do agree. I did go back a re-read [PEP 440](https://www.python.org/dev/peps/pep-0440/) and [PEP 345](https://www.python.org/dev/peps/pep-0345/) looking for something stating that "5.12" is the same as "5.12.0", but that does not seem to be the case.

I'll undo this out-of-spec behaviour.

win_installer.py
tmontes: Correct. That's due to the way `pynsist` works -- which does feel odd, but it is what it is. We must populate the config file with two sets of dependencies: those available as wheels at PyPI, and the remaining ones.

win_installer.py
tmontes: Returning `None` is different from raising a `ValueError` here:
* `None` the requested package version exists, but there are no wheels at PyPI.
  This will have the dependency be added as a non-wheel dependency in `pynsist`'s config.
  AFAICT, `pynsist` will copy those dependencies directly from the environment into the
  packaged artifact -- this fails badly in the case of `PyQtChart`.
* `ValueError` no such package version exists at PyPI (ie: `PyQtChart==5.12.0`).
  This is the `PyQtChart` case we're hitting now: given the meta-data inconsistency,
  even though our `setup.py` requires `==5.12`, the packaging venv version comes out
  as `==5.12.0` (`pip freeze --all` output) which `yarg` fails to locate and, thus,
  fails to add to the `pynsist` config wheel list. Now here we have two options, I guess:
  * Print out a "BIG THIS LOOKS BAD" message and move on (having the dependency
     added as a non-wheel in `pynsist` config, and bound to fail later)
  * Print out a big "INCONSISTENT METADATA" message and abort right there.

What do you say?

tmontes:
@carlosperate,

I have updated the PR such that:
* It no longer "fixes" the issue derived from the inconsistent `PyQtChart` metadata -- so builds are bound to fail.
* It prints a somewhat "BIG" message complaining about such inconsistencies -- see https://ci.appveyor.com/project/carlosperate/mu/builds/29276600#L679
* Additionally, if exists early when any sub-process call is non-successful -- this helps pinpointing failure cases, avoiding cascading failures and their messages in the logs.

Of course, none of the CI builds are expected to succeed:
* Windows fails but `win_installer.py` now better points at the suspect source.
* macOS fails due to the `python-dateutil` dependency that `pip` doesn't handle well enough.

PS: The outcome is a better, if marginally, `win_installer.py`. Let me know of your thoughts. Thanks again!

PPS: Why on earth are AppVeyor's logs out of order?!... There's some buffering going on, for sure. :(

ntoll:
@tmontes I've just spent my morning catching up on a bunch of outstanding email and have to write a presentation for tomorrow. Late afternoon / this evening I want to get to this work, my ticket #966 on Linux based Windows packaging and anything else I can think of vaguely to do with Windows packaging (e.g. updating to latest Python 3.7).

My aim is to find a way forward amongst all these tickets and consolidate this into a working master. ;-)

@carlosperate once Windows has a working master, my aim will be to update OSX to Python 3.7 and fix any outstanding issues on that front -- especially the SSL related problems we've been encountering.

Both, as always, huge thanks and appreciation for all the work and effort you put into Mu. :cow: :heart: :+1:

Sound like a plan..?

tmontes:
@ntoll, thanks for sharing your plans. :)

Here's how I'd go about it, if I were to work on packaging related things:
* I'd merge #959, ensuring a very clear comment on why `python-dateutil` is i) explicitly listed, and, ii) locked at that specific version.
* Then I'd merge this PR #961 that slightly improves `win_installer.py`.
* Lastly I'd go about #966, which I'll go ahead and review now.
* On a tangent, I'd update #937 to DRY-up the `black` configuration as noted [here](https://github.com/mu-editor/mu/pull/929#issuecomment-539467092), simplifying both the `Makefile` and `make.py`, and merge that one too.
* I'd confirm #925 is no longer an issue after the improvements made to the associated tooling.
* Then, there is #940 which sooner or later will need to be addressed, I suppose.

PS: Isn't packaging fun? :)

ntoll:
@tmontes OK.... that sounds like a good plan for me to follow. It turns out my morning emails have caused a flurry of further emails that require my attention. Tomorrow and Thursday will definitely be Mu days and I've earmarked the rest of December for working on Mu too. I feel there's a logjam that needs considerable attention and then we can "flow" to 1.1 beta.

Regarding packages... here's a photo I took of the "open spaces" board at PyCon 2019... ;-)

packaging

tmontes:
> @tmontes OK.... that sounds like a good plan for me to follow. It turns out my morning emails have caused a flurry of further emails that require my attention. Tomorrow and Thursday will definitely be Mu days and I've earmarked the rest of December for working on Mu too. I feel there's a logjam that needs considerable attention and then we can "flow" to 1.1 beta.

@ntoll, I expect to have above average availability during December so its a matter of coordinating efforts towards coming up with an improved release. :)

win_installer.py
carlosperate: Sounds good.

win_installer.py
carlosperate: Yeah, I agree with your reasoning of using None in this case.
I generally try to avoid returning None unless it's very clear that is an option simply by reading the code, but that's what function docs are for as well, and in this case it's not that unexpected.

carlosperate: > -   Print out a big "INCONSISTENT METADATA" message and abort right there.


This is my preferred option.

win_installer.py
```diff
@@ -152,23 +165,36 @@ def pypi_wheels_in(requirements):
         print("-", requirement, end=" ")
         package = yarg.get(name)
         releases = package.release(version)
-        if any(r.package_type == "wheel" for r in releases):
+        if not releases:
+            feedback = (
+                "WARNING: Just installed but not found at PyPI."
+                " (bad meta-data?)"
+            )
```
carlosperate: So in this case we are only writing a warning to the console.

The reason I would prefer to fail the build is because I honestly don't think we will ever catch a warning like this unless there is a problem with the build.

Now, if PyNsist fails somewhere else then there is no problem.

However, if we ended up downloading the package source code and compiling in the build server, the build could still pass, right? Or does PyNsist refuse to build anything from source?
If the build passes in this case then we could end up with a version of Mu that works in most of our tests system, but fails in some user environments and we might not catch that.

Generally if we compile something we should aim to do that in the lowest common denominator (Win 7?), as we had issues related to that in the past, specifically with some dynamically linked win libraries that are common, but not everywhere.

tmontes: > So in this case we are only writing a warning to the console.
> 
> The reason I would prefer to fail the build is because I honestly don't think we will ever catch a warning like this unless there is a problem with the build.
> 
> Now, if PyNsist fails somewhere else then there is no problem.

That is the case as of today, with regards to `PyQtChart`, yes.

> However, if we ended up downloading the package source code and compiling in the build server, the build could still pass, right? Or does PyNsist refuse to build anything from source?

AFAICT, `pynsist` does not build anything: it collects wheel dependencies from PyPI, on one hand, and copies any other dependencies from the Python environment it is running on -- and fails miserably at handling `PyQtChart`, for example, but does its thing nicely with the `tkinter` or `turtle` modules.

> If the build passes in this case then we could end up with a version of Mu that works in most of our tests system, but fails in some user environments and we might not catch that.
> 
> Generally if we compile something we should aim to do that in the lowest common denominator (Win 7?), as we had issues related to that in the past, specifically with some dynamically linked win libraries that are common, but not everywhere.

Agreed, even though `pynsist` itself does not build anything, in the more strict Python packaging build sense. Additionally, and into the realms of #893, if we're looking for consistency and reproducibility, we should probably lock all direct and indirect dependencies in `setup.py`. Such a policy would have avoided the macOS failing builds we're having now due to the `python-dateutil` version conflict that pip doesn't know how to solve. :)

win_installer.py
tmontes: I'll update the PR for that simpler, stricter behaviour. :)

tmontes:
@carlosperate, now exiting loudly and early when failing to locate a particular requirement at PyPI. See https://ci.appveyor.com/project/carlosperate/mu/builds/29286719#L585.

tjguk:
Damn. I've just applied a mini version of this fix without spotting this well-worked PR. Let me go back and see if I can merge stuff in here.

carlosperate:
I was just about to tag you in case you wanted to have a look at this PR. 🙂
I haven't check your changes to the script, but I suspected that updating PyQt would basically "resolve" the issue the CI was encountering, however we probably still need to address the underlying issue.

PR #966, opened 2019-12-02
Windows packaging to work on Linux

ntoll:
**DO NOT MERGE**

This is a first attempt to fix things up so pynsist works again on Linux and OSX based computers.

I've got it working on Linux and have managed to build the installer using `make`.

I've had to update the PyQt based packages (see the changes in `setup.py`) since the check for available wheels wasn't working for the older versions. There are a couple of hacks to deal with `pip` related strangeness and the fact that some packages are not installed on Linux, which would be installed on Windows.

In any case, the installer works *AND* Mu runs.

However, and here's where I could do with some help, when you try to run the REPL Mu crashes. Looking at the log I see an exception raised somewhere in Jupyter (i.e. not *our* code) because it can't `import win32api`. Check out the installer I made (http://ntoll.org/static/files/mu-editor_64bit.exe), and please don't hesitate to give me some pointers. :-)

I'm assuming this PR will result in an installer created by Appveyor (i.e. on Windows). It'd be good to see what the difference, if any, is.

My guess is we need to include something (pywin32) in the `pkgs` directory, but I've not yet figured out that aspect properly yet.

### DIFF
--- a/package/mu_nsist.py
+++ b/package/mu_nsist.py
@@ -22,7 +22,6 @@ import ntpath
 import os
 import shutil
 import sys
-import winreg
 import zipfile
 
 from nsist import InstallerBuilder
@@ -32,20 +31,6 @@ from nsist.util import download, get_cache_dir
 pjoin = os.path.join
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
-
-
-def find_makensis_win():
-    """Locate makensis.exe on Windows by querying the registry"""
-    try:
-        nsis_install_dir = winreg.QueryValue(
-            winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\\NSIS"
-        )
-    except OSError:
-        nsis_install_dir = winreg.QueryValue(
-            winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\\Wow6432Node\\NSIS"
-        )
-
-    return pjoin(nsis_install_dir, "makensis.exe")
 
 
 class InputError(ValueError):
--- a/setup.py
+++ b/setup.py
@@ -21,9 +21,9 @@ with open(os.path.join(base_dir, "CHANGES.rst"), encoding="utf8") as f:
 
 
 install_requires = [
-    'PyQt5==5.12.1;"arm" not in platform_machine',
-    'QScintilla==2.11.1;"arm" not in platform_machine',
-    'PyQtChart==5.12;"arm" not in platform_machine',
+    'PyQt5==5.13.2;"arm" not in platform_machine',
+    'QScintilla==2.11.3;"arm" not in platform_machine',
+    'PyQtChart==5.13.1;"arm" not in platform_machine',
     # `flake8` is actually a testing/packaging dependency that, among other
     # packages, brings in `pycodestyle` and `pyflakes` which are runtime
     # dependencies. For the sake of "locality", it is being declared here,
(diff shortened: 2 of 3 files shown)


ntoll:
FWIW, once I've got it building a working version of Mu (i.e. the REPL stuff is fixed) I'll make sure this is updated to use the assets required for Python 3.7.

carlosperate:
@ntoll there is also this PR about updating dependencies, has more info about why it was failing, and it is currently failing in macOS.

ntoll:
@carlosperate yup... ack on that... I'm currently spinning plates, juggling ferrets or whatever other colourful analogy you want to use for several concurrent tricky tasks... when it comes to packaging Mu. They're (hopefully) all resolve over the coming days as I find time to concentrate on all this stuff.

tjguk:
It's almost certainly to do with the fact that the pywin32 modules do some trickery to get themselves in the right place on the path. And that tends not to play well with redistribution mechanisms like pynsist. I'll try to dig around to get a handle on the best way forward. I'm quite sure we can find something even if we have to resort to a bit of hackery

carlosperate:
> it's almost certainly to do with the fact that the pywin32 modules do some trickery to get themselves in the right place on the path. 

I've heard there was issues related to this area in pip 19, might be worth building with pip 18 to see if that works, it works is not an actual solution, but might help narrow down the problem.

carlosperate:
No luck: https://ci.appveyor.com/api/buildjobs/69wc90b69sy3ubdc/artifacts/mu_2019-12-03_01_12_win-pip_27ddb8f_64bit.exe

image

tjguk:
Not much time now, but just spotted this from pynsist:

https://github.com/takluyver/pynsist/blob/master/examples/pywebview/installer.cfg

I'll come back later to see if we can integrate the approach

setup.py
```diff
@@ -33,13 +33,14 @@
...
     "pyserial==3.4",
-    "qtconsole==4.4.3",
+    "qtconsole==4.6.0",
     "pgzero==1.2",
     "appdirs>=1.4.3",
     "semver>=2.8.0",
     "nudatus>=0.0.3",
     'black>=18.9b0;python_version > "3.5"',
     "Flask==1.0.2",
+    'pywin32>=227;platform_system == "Windows"',
```
tmontes: AFAICT, this is the only non-direct dependency that is listed here: is it needed, given that it is being explicitly brought in during the packaging process in `win_installer.py`?

If so, could we include a comment on why that is the case (much along the lines of the comment on `flake8` that is not a direct dependency, but `pycodestyle` and `pyflakes`, in which it depends, are).

win_installer.py
```diff
@@ -187,12 +190,21 @@ def create_pynsist_cfg(python, repo_root, filename, encoding="latin1"):
     icon_file = os.path.join(repo_root, "package", "icons", "win_icon.ico")
     license_file = os.path.join(repo_root, "LICENSE")
 
+    # On Linux Debian systems "pkg-resources" is erroneously reported.
+    # This is a bug in Debian's patched version of pip.
+    excluded_packages = {mu_package_name, "pkg-resources", "noise"}
```
tmontes: Understand the note on `pkg-resources`. What's the case with `noise`? Do we need to extend the comment a bit or should `noise` not be in the `excluded_packages` set?

package/mu_nsist.py
```diff
@@ -34,20 +33,6 @@
...
-            winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\\NSIS"
-        )
-    except OSError:
-        nsis_install_dir = winreg.QueryValue(
-            winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\\Wow6432Node\\NSIS"
-        )
-
-    return pjoin(nsis_install_dir, "makensis.exe")
-
-
```
tmontes: Being ignorant of `pynsist`'s inner workings, I wonder why this is no longer needed. :)

tjguk:
I'm trying a few casts with the Windows packager now, but I'm hitting several surprising roadbumps including a missing "requsets" and the lack of makensis referred to earlier by @tmontes. I'll persevere and see what I can land.

carlosperate:
This is outdated now that we've moved to Pup.

PR #980, opened 2020-01-01
Add new zh_CN translation

Tairraos:
I use xgettext rescanned source file, updated old po file. and translated all shared.py and part of microbit string. And will keep translating.

Thanks for this great IDE, it is very helpful for new pythoner.


leafdown:
@asb @illume @inglesp @ntoll, could you help to check why mu/modes/microbit.py  can't pass in  AppVeyor build? I verified local, it did 100% pass

Issue #834, opened 2019-05-10
Incomplete translation of the Mu user interface


wayoda:
### Incomplete translation of the Mu user interface
Hi, I dug a little deeper into the Mu sources to find out why some parts of the
user interface are not translated when Mu is run in a non-english environment.
This is a kind of summary what parts of the app are are affected and what are
the options to fully translate Mu. 

For the rest of this comment I will illustrate the problem by showing what are
the effects on the user when he runs Mu in either the `de_DE.utf8`-locale or in
the `pt_BR.utf8`-locale. The Mu interface is translated for both languages but
there are some slight differences between these two when Mu is run. 

#### Related issues on github
For me this all started out with #832 (now closed) which I opened because the `Ok` and
`Cancel` Buttons on MessageBoxes where *not* translated into german but showed
the english text. In the meantime I have closed this issue because it was just
an edgecase to the broader problem. 
And there is #756 by @tmontes which refers to another edgecase: *Context Menus
are not localized* which got me started to dig a little bit deeper...

#### What parts of the interface are affected?
1. QMessageBoxes/QDialogs that ask the user for confirmation of some action. (Like
   quitting Mu while there are unsaved files) The buttons should read `OK` and
`Abbrechen` in a german locale, but the english labels are shown.
2. QFileDialog for opening and saving files. The buttons should read `Öffnen` or
   `Speichern` again the english labels are shown.
3. The (right-click) context-menu in the `editor` pane
4. The (right-click) context-menu in the `REPL` pane

#### Where do translated strings come from?
There are 3 different sources for translations that must work together to
result in a fully localized Mu interface.
1. The `po`-files created by Mu developers. These are complete for german
   and portuguese (and a few other languages).
2. Default translations provided by the Qt-Lib itself that localize the
   QMessageBoxes and the QFileDialog. They are available for german but **not**
for portuguese :-(
3. Translations for the context-menu in the editor are provided by the
   `qscintilla` project, but **only** german and spanish versions are shipped
with the library. (There are resources for 3 more languages in `qscintilla` but they are only
partly translated and miss the translations for the context-menu)

#### Localizing `QMessageBoxes` and `QDialogs` 
There are two options for this:
1. Load the standard translations from the `Qt-Lib` and let Qt do the work. You
   would have to install a `QTranslator` in the app and load the resources
shipped with `Qt`. This works for german but **not** for portuguese because the
translation for this language is incomplete!
[...]

PR #983, opened 2020-01-03
Prepare standard buttons on dialogs for translation

wayoda:
The Mu interface uses dialogs in a few places (Find/Replace; Mode-Changes). The `QDialogButtonBox` provides some standard buttons for `Ok`, `Cancel`. But the text on these buttons is not translated by Qt itself.
I run mu in a `de_DE.UTF8` locale but other languages are affected too (`fr_FR.UTF8` in this case): 
mu-dialog

The changes in this pull request prepare the standard buttons on these dialogs for translation into all supported languages.

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -101,6 +101,8 @@ class ModeSelector(QDialog):
         button_box = QDialogButtonBox(
             QDialogButtonBox.Ok | QDialogButtonBox.Cancel
         )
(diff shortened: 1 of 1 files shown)


ZanderBrown:
In theory Qt does provide translations, not that it seems to work

wayoda:
I did a little research in #834 with resulted in a few options for solving the problems

Issue #985, opened 2020-01-17
File listing Raw REPL error due to startup delays in boot.py


WayneKeenan:
If you are reporting a bug, we would like to know:

* What you were trying to do,

List files on an ESP32 board on macOS.

* What steps you took to make this happen,

Click Files

* What you expected to happen,

Show files 
* What actually happened,

`OSError: Could not enter raw REPL.`

* Why this difference is problematic (it may not be a bug!),

Can't see files.

* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 

This is occurring for me because there is code in boot.py that is taking a few seconds (5ish) to complete before presenting the REPL prompt. When I deleted boot.py all is good.

Is there a way for a user to configure the timeout before Mu gives up?
Or can the default be increased (or use an exponential backoff heuristic) please?

related:  https://github.com/mu-editor/mu/issues/858

WayneKeenan:
Simple `boot.py` test cases:

Works:
```
from utime import sleep_ms
sleep_ms(500)
```

Fails:
```
from utime import sleep_ms
sleep_ms(1000)
```

dybber:
I had the problem with M5Stack devices.

15 seconds is perhaps a bit long to wait for an error message to appear - in the case where the problem is not boot.py

Does your boot.py really take that long to run? I think we shouldn't increase it more than necessary. If the error appears again for someone else we can increase it some more.

Perhaps we should consider indicating in the UI that Mu is waiting with some kind of Busy indicator

WayneKeenan:
15 secs was a bit finger in the air.  I think the `boot.py` script took about ~6 seconds (when connecting to my main wifi AP), but the connection setup time varied between different access points and there's environmental factors that introduces variable delays to connection setup time, sometimes.

Perhaps defaulting it, say to 5, and adding it as a configurable setting in the settings file or as a UI config option, rather than having to (re)build mu to change it.

dybber:
Hi @WayneKeenan 

Okay, actually I think you shouldn't put wifi-connectivity stuff in the boot.py, but in main.py. Let me try to explain why.

Both boot.py and main.py run on startup but serves different purposes:
 - boot.py is for setting up various configuration options, that should be the same across any program you might wish to write
 - main.py is for executing your startup program: that is running your actual application

[...]

PR #986, opened 2020-01-17
Increase serial timeout for RawREPL when listing files

WayneKeenan:
This is an untested and noddy suggestion for #985

### DIFF
--- a/mu/modes/base.py
+++ b/mu/modes/base.py
@@ -440,7 +440,7 @@ class FileManager(QObject):
         """
         # Create a new serial connection.
         try:
-            self.serial = Serial(self.port, 115200, timeout=1, parity="N")
+            self.serial = Serial(self.port, 115200, timeout=15, parity="N")
             self.ls()
         except Exception as ex:
             logger.exception(ex)


WayneKeenan:
I've tested this against a current master dev build and it "WorksForMe(tm)"

dybber:
Hi @WayneKeenan, could we meet in between, e.g. at 5-6 seconds when listing files?

WayneKeenan:
> Hi @WayneKeenan, could we meet in between, e.g. at 5-6 seconds when listing files?

Fixed timeouts are never really a fix, exponential back off delays with an upper retry count bound are.

ntoll:
Just waiting for an extended period of time without any indication to the user that something's going, on is a UX "bad smell". I also wonder if the serial connection hasn't been made after 1 second then it'll never be made. Thoughts..?

WayneKeenan:
> Just waiting for an extended period of time without any indication to the user that something's going, on is a UX "bad smell". I also wonder if the serial connection hasn't been made after 1 second then it'll never be made. Thoughts..?

There’s more detailed info as to why it’s successfully taking longer than 1 sec in #986,  tl;dr performing Wifi setup in boot.py on ESP32 blocks serial for a number of seconds.

dybber:
Just added a comment over at #985 about how I'm dealing with wifi connectivity while teaching.

WayneKeenan:
I'm going to close this PR as it has a wider ranging impact on the more typical use case, it's just too noddy.

It would be better for me to create a new PR that reads an optional timeout set in Mu's config, or otherwise default it to 1.

This would support cases where a users knows that their particular `boot.py` takes a while to start, for what ever reason.

PR #1010, opened 2020-02-21
Allow device recognition based on manufacturer-string.

dybber:
This allows detection of devices using generic VID/PID (such as the
currently popular M5Stack devices).

This change also makes device names independent of the mode
name. Making device name independent of the mode name, which make it
possible to show not only that a device of a specific mode has been
connected, but also the name of the device (e.g. Adafruit
Feather).

Screenshot 2020-02-21 at 13 01 36


dybber:
I will close this PR without merging, as the code here is also integrated in PR #1017.

Issue #988, opened 2020-01-21
Feature Request: Manually Select COM port


the-stanely:
This request is for Windows 10.

Please provide a means for the user to select desired COM port.  It would be useful for users with a number of active COM ports. 

Also, multiple instances of Mu are possible, so it would be useful when working on multiple IoTs that need to communicate with each other.

Thank you.

dybber:
I completely agree. Without the ability to override the automatically selected device, it's also problematic to communicate with certain devices. For instance, the currently very popular ESP32-based "M5StickC" uses a generic Vendor and Product ID, so it is not possible to detect it. The user has to be able choose to connect to unrecognized devices such as the M5StickC.

I've just created a few mockups of how it could work. We could add a dropdown in the statusbar, where the user can choose between devices:
![Screenshot 2020-02-14 at 12 40 24](https://user-images.githubusercontent.com/55204/74530101-8dd4ac00-4f29-11ea-8627-03d736ad64ee.png)
![Screenshot 2020-02-14 at 12 40 37](https://user-images.githubusercontent.com/55204/74530123-9e852200-4f29-11ea-819c-26eee2a2dabe.png)

For Mac and Linux it would be a bit more complex, for example:
![Screenshot 2020-02-14 at 12 53 34](https://user-images.githubusercontent.com/55204/74530058-7695be80-4f29-11ea-939e-c91b82cfd3b2.png)

Would this make sense? I could try to work on it, if we can agree on how it should work. Perhaps @ntoll has some input here?

We would still have to decide on a few other things as well:
 - Should the selected device follow the open buffer? What about the REPL, should it also support multiple open connections? Alternatively, only support one open connection and make the setting global - thus when you change device, the connection will close, and a new connection would be opened to the selected device. This alternative would be the simplest to do.
 - Should the dropdown menu only be shown when in one of the MicroPython modes? Or should it be shown as soon as there are any USB-serial devices? I think it would make sense to make it part of the MicroPython modes, rather than a global thing

dybber:
Further discussion along the same lines, is in Issue #851 and pull-request #855. Just mentioning @carlosperate, as he could probably also chip in here

the-stanely:
[...]

Issue #839, opened 2019-05-24
ESP-mode REPL: Off by one in error-message line numbers. 


dybber:
Error messages reported when executing a file using the "Run" button in ESP-mode, are reported on line N+1, when error is on line N.

E.g. with an error on line 14, the following error is reported:
```Traceback (most recent call last):
  File "<stdin>", line 15, in <module>```

dybber:
@ntoll: Do you remember why this newline was added:
https://github.com/mu-editor/mu/blob/master/mu/interface/panes.py#L309

ntoll:
No. I believe it may be an un-wanted part of the "recipe" for putting the device into raw mode, copied from the microfs command. I'm up to my eyes in creating a new web mode for Mu, but once that's ready to be a PR I can tackle the "low hanging fruit" bugs in preparation for the alpha.2 release.

dybber:
Okay, I think we should use one of the existing implementations for entering/exiting raw REPL (e.g. call the one in microfs.py), they include very specific timings and checks that all the right phases have passed, and I don't believe a newline here is necessary. I can look at it one of the next days.

dybber:
Okay, the code in microfs is harder to reuse than I anticipated as microfs uses pyserial, and Mu uses QSerialPort which are incompatible.

dybber:
I can see the reason for using QSerialPort to use Qt signals on incoming messages. Should I try porting the enter/exit raw REPL code from microfs.py or pyboard.py to use QSerialPort?

carlosperate:
@dybber did #839 fix this issue?

dybber:
Yes, this was fixed by #1017

Issue #810, opened 2019-04-03
Adafruit Metro 328 detected as ESP MicroPython Device


makermelissa:
This probably isn't a big deal and you can close this issue if you'd like, but I noticed on Mu Alpha 1.1 that the Adafruit Metro 328 comes up as being detected as an ESP MicroPython Device. The memory is probably too small to run MicroPython and it wouldn't really harm anything.

ntoll:
@makermelissa aha... I wondered when the first false positive would come up. My guess is the USB controller on the Metro 328 is the same as for an ESP based board, so reporting the same VID/PID..???

dhalbert:
Adafruit board detection is now based on the VID (0x239a) only. We thought this was a compromise: some false positives vs. not having to add new boards all the time.

makermelissa:
Makes sense.

carlosperate:
> Adafruit board detection is now based on the VID (0x239a) only. We thought this was a compromise: some false positives vs. not having to add new boards all the time.

This sounds like a good compromise. The main disadvantage is that Mu will assume all boards with that VID (including future ones) are CircuitPython boards. So there could be false positives, but Mu also assumes the board is running CircuitPython (and in some cases this boards could be running something else, like an Arduino program), so to some extent we are already trusting the user is plugging-in a board that is ready to be programmed with Python.

I think the Metro is based on the CP2014, which some ESP boards use as their UART-to-USB bridge as well. Is Adafruit changing the CP2014 VID to 0x239a as well?

dybber:
For the Adafruit Metro 328, I guess this actually doesn't matter that much, as it will probably never be able to run MicroPython/CircuitPython.

At some point however, we should consider making it possible to change modes independently of the selected device. On connection it should still suggest one mode, but if you really want to use another mode instead, I think you should be allowed to try to do that, with appropriate warnings. 

In the future we will probably have many such modes, there's already pull requests for at least 4 such MicroPython based modes (Pyboard, Makeblock Studuino:bit, snek-mode), and they might overlap in device compatibility due to the issue with the same VID/PID being reused by many vendors.

dybber:
Actually, that is the case as it is. It might be detected as one type of device, but you can choose another mode and use that. I will close this issue, as mentioned in my last comment the Adafruit Metro 328 will probably never be able to run either MicroPython or CircuitPython.

Issue #276, opened 2017-06-22
Single instance application


carlosperate:
We need to consider if this is something we should enforce with Mu. As discussed in https://github.com/mu-editor/mu/issues/268 several instances of the editor would be using the same config file and log.

The easiest solution would be to set the application to be single-instance only, so that would be my vote. But it'd be good to check if this is also the best solution for the Mu target audience, does anybody have an insight in this area?

ntoll:
I have absolutely no insight into this, nor do I know how to enforce single instance. Any idea how this would work? (I believe single instance is probably a helpful thing to do given the context you describe @carlosperate).

ZanderBrown:
I think we already have this?

ZanderBrown:
So in the Qt world the answer seems to be that it's doable but harder than it should be

tjguk:
The conventional approach is to attempt to acquire a mutex of some sort and, on failure, switch focus to the already-activated version of the app.

Trying to activate an existing app can be tricky, so an alternative to transparent activation of the existing instance is to force failure by putting up a modal dialog warning that an instance is already running.

The second is the less complex option, but even so I don't know whether it's necessary in comparison to the small risk of a session file being unexpectedly overwritten.

tjguk:
Alternatively, we can ignore the single-instance notion and work harder to *merge* session files when saving. Again, I don't know if this complexity outweighs the risk.

ZanderBrown:
This is why GApplication is so nice, I'd have thought everyone would have these niceties by now but it seems QApplication is relatively "dumb" in this area

Either solution is going to be complex but the current state of affairs isn't great either (especially as we are an open with the target now)

carlosperate:
Maybe we can use this: https://github.com/IARI/alsa_jack_gui/blob/master/qtsingleapplication.py

tjguk:
That's basically technique (1) for my post above. And it's good to have something already written and working. But it does complicate the startup of the app and leaves the door open for odd problems.

ntoll:
OK... so there's a thing in Qt called QLockFile... see: http://blog.aeguana.com/2015/10/15/how-to-run-a-single-app-instance-in-qt/ and https://doc.qt.io/qt-5/qlockfile.html

We'd just need to do:

`from PyQt5.QtCore import QLockFile`

Check/create lock file and if there's a problem complain with a modal error box. This feels the simplest solution.

carlosperate:
[...]

PR #1017, opened 2020-03-23
Support for multiple devices, and refactor of the device-interfacing code

dybber:
This pull-request relates to MicroPython-modes (ESP, micro:bit, CircuitPython), and how devices are detected and shown in the UI. In the beginning this was an attempt to fix issue #988, but in the process various refactorings was necessary and, I also ended up fixing another bug (#839) in the process.

The pull-request includes:
 - a refactor of USB/serial-interfacing code, to separate concerns all serial interaction is moved to the modes, away from the general UI-code, more specifically a REPLConnection class has been created to handle the interactions
 - support for multiple devices, with a drop-down where users can select between devices, when there are more than one device connected. The new UI element is shown below.
 - a fix for issue #839 
 - devices are now able to be recognized based on manufacturer string

The refactoring should make it easier to maintain the MicroPython-modes in the future, and through the new dropdown menu it should be easier to get a fix for issue #810 as well.

**I would like to get feedback on:**
 - how well this works with various devices and platforms (especially on Windows/Linux), I only have Mac. @the-stanely has made some tests on his Windows-machine, where it seems to work for his uses. (Thanks!)
 - the new UI for indicating device connectivity + device selection

**UI:**

In Python mode, no change:

In micro:bit mode, without any connected device:

In micro:bit mode, with a connected device:

With multiple devices connected:


dybber:
Well, it's not really multiple devices, but you can now select beween devices manually. This would in theory allow users to open multiple Mu-instances and interact with two devices in each their separate Mu-instance.

carlosperate:
> This would in theory allow users to open multiple Mu-instances and interact with two devices in each their separate Mu-instance.

I think right now Mu might assume in some places that is only a single instance running, things like sharing the settings file and log file, could be problematic if more than once instance was running at the same time.

For simplicity I would probably go the other way and force the editor being a "single instance" app (https://github.com/mu-editor/mu/issues/276), but that doesn't mean that we couldn't just track these issues and resolve them to support it.

the-stanely:
I tested this version with 3 simultaneous ESP32s from 3 different manufacturers.  It worked great.  Startup opened the most recent session.  I didn't see any problems opening different files in different instances.  Admittedly, I rarely use the log.  When I do, something went wrong.  This has been rare.

I think the ability to open multiple instances with several devices is more powerful than the ability to keep the log straight -- though I don't know if the log gets confused in that case.

When developing multiple devices that communicate with each other, being able to debug them simultaneously is an excellent feature.  This is especially true with ESP32 which come with WiFi and Bluetooth.

the-stanely:
Other features for Mu.  I realize that Mu is supposed to be targeted at beginners, but some advanced features would benefit everybody.  Adding an F3 key for 'search again' would be great.  This is a standard feature in many editors and would speed software development for users.

The search and search replace feature is cumbersome.  There is no edit menu, so users have to know in advance to use ^F.  This might be a mystery to new users, it doesn't affect me.

Occasionally I forget that I replaced something recently, then when I go to search new, I get a replace.  Would it make sense for the find/replace popup to have 'Find', 'Replace', and 'Replace All' buttons?

One of Mu's nice feature is that it integrates some of ampy with the 'Files' feature.  When upip is used, it creates sub-directories.  It would be nice if Mu's 'Files' capability was enhanced to navigate the file structure, especially on the left side.  Right side navigation enhancement would be great too.  Basically, navigating into sub-directories and parent directory '..' would be a valuable enhancement.

dybber:
Hi @the-stanely: 
Great that this have worked for you. I hope this might find it's way into a next release of Mu.

I'm don't have any good solutions for the multiple instances issue right now. I think @carlosperate is right, that the best solution and what everyone else seems to do, is to make it a single-instance application, and then from there try to make support for having different tabs running with different modes, or having multiple windows within the same app. I still think the contribution in this issue simplifies a lot behind that scenes, that perhaps would make this easier to do at least for the micro:bit/ESP-mode/CircuitPython modes.

Regarding your last comment on this issue, please add such other suggestions as individual issues, then they will not be forgotten when this issue is closed. Posting them here also might clutter the discussion around this particular issue. Thanks!

dybber:
Sorry, I shouldn't have merged, but rebased. I will see if I can redo this, and then try to do a force-push so we don't clutter the commit history too much!

dybber:
I've rebased with master and the pull-request should be ready to merge.

tjguk:
@dybber I'll leave others to comment technically as the devices aren't really my area, but from a quick perusal of the code changes I'd say it's a solid piece of work. I very much hope we can get it merged in without any difficulty.

dybber:
Okay, thanks anyway for the nice comment :-) 

Btw. I forgot to mention earlier, that I also have some Inkscape svg's of the chip-connected chip-disconnected icons, which I didn't know where to put, so they are actually yet to be committed. Should they be committed to `mu/resources/images`, even though they aren't used in the UI?

ZanderBrown:
If not the UI where are they used?

dybber:
The PNG’s are used in the UI, was just questioning whether I should commit the vector versions as well (svg format)

ZanderBrown:
Ah, with newer images I think we've just been using the SVG itself

ntoll:
OK... I've tested this on Linux with:

* CircuitPython CPX and Bluetooth version thereof. :heavy_check_mark: 
* Microbit. :heavy_check_mark: 
* ESP8266/ESP32 :heavy_check_mark: 

All works for me. I'm going to merge this.

Thank you so much to @dybber for his outstanding work (and even more outstanding patience with me reviewing it). Here's to a quicker turn around next time. :wink:

ntoll:
OK... hold on that merge. There's some linting and (micro:bit) unit test failures as the branch currently stands. Try it with:

```
$ make check
```

I'm just off to do some dad taxiing, but will try to fix the test fails tomorrow (then merge).

ntoll:
Hurrah...! Fixed and merged. :rocket: :shipit:

Issue #806, opened 2019-03-31
Unintuitive behavior of "Run" in esp8266/esp32 mode


HALtheWise:
If you are reporting a bug, we would like to know:

#### What you were trying to do
I am teaching a class soon using the esp8266 as a platform for controlling small robots. We would like to use Mu as the editor for this use case, and are very excited by the new support in the beta!

#### What steps you took to make this happen
Create a file called "main.py", write some code in it, then click the "Run" button to send it to the esp8266. Unplug the microcontroller and restart it, to test on the actual robot.

#### What you expected to happen
Like (I believe) happens with the micro:bit, I expected the "Run" button to flash code in a permanent way, such that when I restart the microcontroller it executes immediately.

In the particular case of the esp8266, this probably involves replacing the "main.py" file with the currently open file, then restarting the microcontroller. It is unclear exactly how other files already loaded on the microcontroller or open in the editor should be handled.

#### What actually happened
When I click "run" the microcontroller stops what it is doing and runs my code. However, after a restart, the microcontroller goes back to executing whatever code it had before.

#### Why this difference is problematic (it may not be a bug!)
This behavior is different than what happens with the micro:bit, and means we will need to give significantly more complicated instructions to students. For our use case, they would need to:
  - name their file `main.py`, rather than `challenge1.py` or whatever
  - use the "Run" button to test their code when plugged into the computer, so you can see the terminal output
  - close the REPL tab
  - open the "files" tab, then drag `main.py` to the board
    - If you forget this step, the board will run the previous version of the code with no indication what is happening, creating a lot of confusion.

I think this is a pretty typical use case, and the current UX is just going to confuse students.

#### Technical details
`master` at f297c49d9a58fbed2d5b47d2056797f6910c469a, Linux with esp8266

dybber:
Hi @HALtheWise 

Your suggestion was actually the original implementation of the Run-button (called Flash at the time), where it created a "main.py", with the motivation to make it similar to micro:bit mode.

However, after a discussion of the pull request here #676 (and also in the follow-up here #790) we decided against it for the following reasons:
 - When the file/tab is called e.g. "challenge1.py", it is confusing that it then writes to "main.py"
 - We shouldn't duplicate functionality (simplicity is core to Mu), and we have the File-manager to copy files over
 - Students should learn about the filesystem, and structure projects in directories
 - During development, it dramatically speeds up

[...]

PR #1018, opened 2020-03-24
Add right-click menu option "Write to main.py on device" in local filepane

dybber:
This is a possible and minimal fix for issue #806, where an extra menu item is introduced in the local file system view to make it easy to write files to "main.py" in MicroPython-modes (perhaps it should only be shown in ESP-mode?).

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -466,9 +466,8 @@ class LocalFileList(MuFileList):
         # Mu micro:bit mode only handles .py & .hex
         if ext == ".py" or ext == ".hex":
             open_internal_action = menu.addAction(_("Open in Mu"))
-            write_to_main_action = menu.addAction(
-                _("Write to main.py on device")
-            )
+        if ext == ".py":
+            write_to_main_action = menu.addAction( _("Write to main.py on device") )
         # Open outside Mu (things get meta if Mu is the default application)
         open_action = menu.addAction(_("Open"))
         action = menu.exec_(self.mapToGlobal(event.pos()))
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -418,6 +418,7 @@ class LocalFileList(MuFileList):
     """
 
     get = pyqtSignal(str, str)
+    put = pyqtSignal(str, str)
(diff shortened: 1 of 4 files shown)


mu/interface/panes.py
```diff
@@ -498,6 +499,9 @@ def contextMenuEvent(self, event):
         # Mu micro:bit mode only handles .py & .hex
         if ext == ".py" or ext == ".hex":
             open_internal_action = menu.addAction(_("Open in Mu"))
+            write_to_main_action = menu.addAction(
+                _("Write to main.py on device")
+            )
```
carlosperate, line 504: If this menu item is added for hex files as well, does that mean it will try to transfer the entire hex as a 'main.py' file?

dybber: Yeah, but that doesn't make any sense. I'll take a look.

dybber: I guess this only makes sense for the ESP-mode, so I probably need to add in a different way from within the mode, and only for .py files.

dybber: In principle, when the modes are detached from the editor, this will be easier to do. Right now, we'll risk adding even more spaghetti code if we try to allow modes to add a context menu item.

In the future I hope a mode will be able to define its own UI components, instead of the UI components have to be shared by many modes, e.g. provide a default filesystem pane, and allow modes to inherit and customize it (the same for the plotter, REPL, etc.).

So I think the best we could do now, is to add this for all the modes (but only for .py), instead of introducing tighter couplings between modes and panes as it is.

How does that sound?

dybber: @carlosperate

carlosperate: Yeah, doing it only for .py files for now sounds good.
Technically it could also make sense for the micro:bit mode, as a hex file can contain a Python script (that's how uFlash adds code to a hex right now), but as you mentioned, we could add that once we have more granular control from individual modes.

dybber:
Rebased on top of master. I believe it's ready to merge.

dybber:
Rebased again, to remove a `qtapp` fixture dependency that's no longer needed.

Issue #840, opened 2019-05-24
ESP-mode and Micro:bit mode: Problems with copy-pasting and selection in REPL


dybber:
This issue documents a few problems with copy-pasting in the REPL when using ESP-mode/micro:bit mode on Linux (the other modes are probably affected as well, and perhaps also on Mac/Windows). I believe we should have a discussion on how to address it (in a cross-platform maner).

**Background:**
Students are used to being able to CTRL+C/CTRL+V in most places, but CTRL+C sends an interrupt signal in REPL. We thus have to use other approaches. This also allows teachers to teach how to work in ordinary terminals (where CTRL+C/CTRL+V might not be available either.

An alternative to CTRL+C/CTRL+V is to use mouse allow "copy on select", and do paste on "middle mouse" (also SHIFT+INSERT), however, when pasting with this approach the pasted code can not be executed in the REPL.

Currently the only "fail-safe" approach to copy pasting in the REPL is to use the right-click menu.

**Problems**
 - It is not possible to select text using the keyboard alone while holding SHIFT and using arrow keys
 - The alternative solution with "copy on select"/paste with middle mouse click works partly, but when trying to execute the pasted code nothing happens
 - The "copy on select"-approach is perhaps not an ideal solution on Mac/Windows - and especially for users using touchpads

**Possible solutions:**
 - Allow selection using SHIFT and arrow keys
 - Add CTRL+SHIFT+C, CTRL+SHIFT+V as alternative shortcuts for copy pasting in the terminal
 - Address the issue on "copy on select" seemingly working on Linux, where the code is pasted but never actually send to the interactive Python process.

dybber:
I have worked on this and I believe I have solved all the selection issues, as well as several other issues with mouse movement of the cursor, etc.

The code is available here: https://github.com/dybber/mu/tree/bugfix/replcursor_movement

I still need to write tests, but it would be nice if anyone else could try it out and check that everything also works for them.

PR #1026, opened 2020-04-15
Support mouse movement, selection, and fix several bugs in micro:bit & ESP-mode REPL

### LINKED ISSUE #987: Up Arrow in REPL for Previous Command Problem
In REPL on Windows 10, problem with up arrow for previous command.

First up arrow is fine.  Subsequent keystrokes don't erase the previous command before printing the new one.  Instead, what looks like terminal control characters are printed, followed by the previous command.

The character sequence is like: [10D

[nnD  where nn is the decimal number of characters to the beginning of line.  So it looks like the terminal is asked to left-delete the previous nn characters.  But the sequence is not being interpreted in windows and instead is printed.

Here's an example:

print(mac)[10Dprint mac[9Dmac = ubinascii.hexlify(network.WLAN().config('mac'),':').decode()

dybber:
The micro:bit, CircuitPython and ESP-mode REPL have suffered from several bugs, which this pull requests addresses:
 - Partial reception of multibyte VT100 commands
 - Support for selecting text with SHIFT + cursors
 - Support for mouse movement and mouse text selection
 - General stability

This fixes #987 and #840.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -156,6 +156,8 @@ class MicroPythonREPLPane(QTextEdit):
         self.setUndoRedoEnabled(False)
         self.setContextMenuPolicy(Qt.CustomContextMenu)
         self.customContextMenuRequested.connect(self.context_menu)
+        self.last_cursor_position = self.textCursor().position()
+        self.cursorPositionChanged.connect(self.cursor_moved)
         self.setObjectName("replpane")
         self.set_theme(theme)
 
@@ -197,7 +199,10 @@ class MicroPythonREPLPane(QTextEdit):
         """
         key = data.key()
         msg = bytes(data.text(), "utf8")
-        if key == Qt.Key_Backspace:
+
+        if key == Qt.Key_Return:
+            msg = b"\x1B[F\r" # Return: End + Return
+        elif key == Qt.Key_Backspace:
             msg = b"\b"
         elif key == Qt.Key_Delete:
             msg = b"\x1B[\x33\x7E"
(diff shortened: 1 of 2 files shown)


dybber:
I've rebased with master and this should be ready to merge, however, I think it would be wise to wait with the merge until we have merged the larger refactor in #1017.

mu/interface/panes.py
```diff
@@ -189,118 +205,245 @@ def context_menu(self):
...
+        for every input received.
         """
-        tc = self.textCursor()
-        # The text cursor must be on the last line of the document. If it isn't
-        # then move it there.
-        while tc.movePosition(QTextCursor.Down):
-            pass
+        logger.debug(
+            "MicroPython REPL, received through serial: {}".format(data)
+        )
```
carlosperate, line 360: Same as the other comment, this can get quite busy in the log.

mu/interface/panes.py
```diff
@@ -189,118 +205,245 @@ def context_menu(self):
     def set_theme(self, theme):
         pass
 
+    def send(self, msg):
+        logger.debug("MicroPython REPL, writing to serial: {}".format(msg))
```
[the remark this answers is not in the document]
dybber: I agree. I'll remove those.

mu/interface/panes.py
```diff
@@ -156,8 +169,14 @@ def __init__(self, serial, theme="day", parent=None):
...
         self.setContextMenuPolicy(Qt.CustomContextMenu)
         self.customContextMenuRequested.connect(self.context_menu)
+        # The following variable maintains the position where we know
+        # the device cursor is placed. It is initialized to the beginning
+        # of the QTextEdit (i.e. equal to the Qt cursor position)
+        self.device_cursor_position = self.textCursor().position()
         self.setObjectName("replpane")
         self.set_theme(theme)
+        self.unprocessed_input = b""  # used by process_bytes
+        self.decoder = codecs.getincrementaldecoder("utf8")()
```
carlosperate, line 179: Oh, this is a great find in the standard library!
Does that mean that if an argument to the `decode` function call doesn't have all the bytes for a utf-8 char (so the last byte of the data array contains the first byte of a multi-byte characcter) it will keep state and concatenate incomplete characters into the next `decode` call?

dybber: Yes, it decodes a bytestring in multiple steps and keeps a state of bytes that wasn't parsed between the steps. The explanation in the Python docs is a bit cryptic.

carlosperate:
Raised https://github.com/dybber/mu/pull/1 to add a couple of minor performance tweaks to this PR.

dybber:
I've rebased this on top of the current master, it would be nice if we could do one more round of testing, to check that I didn't introduce any new bugs during rebase.

dybber:
Rebased again, to remove `qtapp` fixture dependencies which are no longer needed.

PR #1032, opened 2020-04-25
Add ESP32 Boards CH552

tanakamasayuki:
I hope CH552 will be added.
It is used in M5StickC etc.

https://docs.m5stack.com/#/en/core/m5stickc

### DIFF
--- a/mu/modes/esp.py
+++ b/mu/modes/esp.py
@@ -44,6 +44,7 @@ class ESPMode(MicroPythonMode):
         # VID  , PID
         (0x1A86, 0x7523),  # HL-340
         (0x10C4, 0xEA60),  # CP210x
+        (0x0403, 0x6001),  # CH552
         (0x0403, 0x6015),  # Sparkfun ESP32 VID, PID
     ]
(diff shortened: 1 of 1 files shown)


dybber:
VID 0x0403, PID 0x6001 are the default values for any USB device using FTDI's USB-serial interface. It would therefore not be appropriate for detecting ESP32 devices, as any other kind of FTDI USB device with the generic VID/PID could then also be recognized as an ESP32 device. Unfortunately M5Stack haven't changed the default VID/PID to something specific for their company. 

However, I really also think it is a platform worth supporting in Mu, and have tried to do the detection based on "manufacturer string" instead. I've tried to do just that over here: https://github.com/mu-editor/mu/pull/1010

I've only tested it on a Mac, where it does recognize the M5StickC. It would be nice with test on Windows/Linux, if you are able to do that.

dybber:
Thanks for the contribution! I've made the necessary changes to be able to merge this and made a new pull request at #1094, and I'm thus closing this PR.

PR #1048, opened 2020-05-24
Attempt at hot-fixing the issues with building Windows installers/Mac .app's

dybber:
Hi all

My hope is that the small fixes here, can perhaps be a temporary fix to the issues that makes AppVeyor unable to generate Windows installers for pull requests. Also makes it possible to build installers on Mac/Linux.

This is just an attempt, and if I have misunderstood something about the somewhat involved build process for Mu, please excuse me and ignore this pull request.

### DIFF
--- a/package/mu_nsist.py
+++ b/package/mu_nsist.py
@@ -22,7 +22,6 @@ import ntpath
 import os
 import shutil
 import sys
-import winreg
 import zipfile
 
 from nsist import InstallerBuilder
@@ -36,6 +35,8 @@ logger = logging.getLogger(__name__)
 
 def find_makensis_win():
     """Locate makensis.exe on Windows by querying the registry"""
+    import winreg
+
     try:
         nsis_install_dir = winreg.QueryValue(
             winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\\NSIS"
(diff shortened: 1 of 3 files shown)


dybber:
Seems AppVeyor was able to build the installer, would any Windows user be able to try build locally (`make win64`) and test the generated installer?

It still reports the following warnings/errors, which I will have to look at another time:
```
ERROR: flake8 3.8.2 has requirement pycodestyle<2.7.0,>=2.6.0a1, but you'll have pycodestyle 2.5.0 which is incompatible.
ERROR: flake8 3.8.2 has requirement pyflakes<2.3.0,>=2.2.0, but you'll have pyflakes 2.1.1 which is incompatible.```

dybber:
Just added an additional fix, that should make it possible to build Mac .app's again.

dybber:
Can someone explain why pycodestyle and pyflakes are mentioned in as requirements in setup.py with incompatible versions to those of the current flake8 version? Why not just depend on flake8, and get the pycodestyle/pyflakes versions flake8 uses?

dybber:
I can see I can see this has been adressed in another pull-request over at #959 with a better solution than my quick fixes, so I will close this.

PR #1056, opened 2020-06-11
Ensure that, if we're developing inside a venv, the venv python is used

tjguk:
This is hopefully a non-controversial PR to ensure that the Windows make.py uses Python from the running virtualenv, otherwise it won't pick up the installed packages

### DIFF
--- a/make.py
+++ b/make.py
@@ -60,7 +60,7 @@ def _process_code(executable, use_python, *args):
     in the codebase, skipping docs and build artefacts
     """
     if use_python:
-        execution = ["python", executable]
+        execution = [sys.executable, executable]
     else:
         execution = [executable]
     returncodes = set()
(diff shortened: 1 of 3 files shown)


tjguk:
Appveyor is happy but Travis is failing with an obscure core dump somewhere in py.test
Since I don't believe it'll have been caused by this change (unless as a result of the changes of version in PyQt etc.) I'm going to go ahead and merge this in, hoping that we'll be able to roll forward if anything does need fixing

dybber:
Perhaps the pytest issues is the one described here: https://github.com/pytest-dev/pytest/issues/6492 - maybe we should try, as suggested, to pin the version to latest one without that issue `pytest==5.3.2`

tjguk:
Thanks, @dybber. I've merged and closed this PR for now (as I'm working on the Windows stuff). But I'll open a separate one with only that pytest change we'll see what happens.

tjguk:
Generated #1057

Issue #969, opened 2019-12-06
Min max


JLFra:
Hello, in graph mode. You fixe min =-max. And that's create a problem reading the graph, because the signal is not so clear. So could you change the min and mut it to a real min and not -max. And them our students could really read signal from sensor. The actual accuracy off the graph is not correct. Because all signal don't have min=-max. thanks to help reading ours students.
Julien

dybber:
I've added a pull request with a fix, which you might want to check out.

dybber:
The issue should be fixed with the #1059 merged into master, which makes min and max y-coordinates on the graph independent of each other.

PR #1059, opened 2020-06-11
Attempt at fixing #969: Calculate Plotter y min independent of y max

dybber:
This is an attempt at what can be done for issue #969 regarding the plotter. I'm not completely sure this is the way to go, but now we can see what issues it brings up if we try to follow the suggestion of making y min independent of y max calculation.

Basically if a signal oscillates, between lets say 0 and 1000, then previously it half of the plot would be empty as y min would be set to -1000, but the signal never decrease below 0. This fix makes that better, but on the other hand if the signal oscillates between say -100 and 100, this fix would then very often change the axis range (some times between 0-100, other times perhaps between 0 and -100)

I think this fix perhaps should be coupled with a fix that delays the change of range when signals decrease, so it doesn't do it immediately (as it might get back up again very shortly), but perhaps if there have been about 500 iterations, with no signal above a threshold then the range should decrease.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -1224,6 +1224,7 @@ class PlotterPane(QChartView):
         # Holds the raw actionable data detected while plotting.
         self.raw_data = []
         self.setObjectName("plotterpane")
+        self.num_datapoints = 0 # Number of datapoints to show (caps at self.max_x)
         self.lookback = 500
         self.max_x = 100  # Maximum value along x axis
         self.max_y = 1000  # Maximum value +/- along y axis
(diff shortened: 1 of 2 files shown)


dybber:
Added a lookback parameter, which does what I described above - retains data after they are no longer shown, to avoid too quick changes in the plot y-axis range. I don't know if 500 is the right level for the parameter, could perhaps use some fine tuning.

dybber:
Rebased, to get to force Travis to run again with the new changes from master. Hopefully it will be all green this time.

dybber:
Oh, tests still needs to be updated.

Issue #1055, opened 2020-06-09
Tab height on fresh start vs restored session


carlosperate:
Whenever Mu starts up (in macOS Mojave and Mu from master commit e3b7191e223a02fbff296fa0ecc834382e16e2b4) and re-opens tabs saved from the last session, it configures the tab height to be larger than they used to be.

I can't remember if this was changed on purpose (I think i like the larger tabs better), but the tab height is not the same in these two situations:
- When no files are opened and a new tab is opened (this is the default state when Mu is installed, or when all tabs are closed)
- When mu re-open tabs from the previous session

<img width="1392" alt="image" src="https://user-images.githubusercontent.com/4189262/84133149-b6376780-aa3e-11ea-8b2e-2c7e2b152203.png">

<img width="1392" alt="image" src="https://user-images.githubusercontent.com/4189262/84133790-8046b300-aa3f-11ea-829f-1b32fd6db166.png">

```
$ pip list
Package                       Version    
----------------------------- ----------
alabaster                     0.7.12
appdirs                       1.4.4
appnope                       0.1.0
arrow                         0.15.6
attrs                         19.3.0
Babel                         2.8.0
backcall                      0.1.0
binaryornot                   0.4.4
biplist                       1.0.3
black                         19.10b0
bleach                        3.1.5
boto3                         1.13.24
botocore                      1.16.24
briefcase                     0.2.9
certifi                       2020.4.5.2
chardet                       3.0.4
click                         7.1.2
cookiecutter                  1.7.2
coverage                      5.1
decorator                     4.4.2
dmgbuild                      1.3.3
docutils                      0.16
ds-store                      1.1.2
flake8                        3.8.2
Flask                         1.0.2
idna                          2.9
imagesize                     1.2.0
importlib-metadata            1.6.1
ipykernel                     5.3.0
ipython                       7.15.0
ipython-genutils              0.2.0
itsdangerous                  1.1.0
jedi                          0.17.0
Jinja2                        2.11.2
jinja2-time                   0.2.0
jmespath                      0.10.0
jupyter-client                6.1.3
jupyter-core                  4.6.3
keyring                       21.2.1
mac-alias                     2.0.7
MarkupSafe                    1.1.1
mccabe                        0.6.1
more-itertools                8.3.0
mu-editor                     1.1.0a2
nudatus                       0.0.4
numpy                         1.18.5
packaging                     20.4
parso                         0.7.0
pathspec                      0.8.0
pexpect                       4.8.0
[...]

PR #1062, opened 2020-06-11
Fix issue #1055 - tab height difference on fresh start vs. reload of files

dybber:
Fix issue #1055 by reordering when set_theme is called. Seems like Qt in this case needs things to happen in a specific order.

When the tabs are loaded without a theme set, it will use a Qt defined default height of the tabs, and not change back to what our stylesheets say, if the theme is set before this issue doesn't happen.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -733,6 +733,7 @@ class Editor:
         """
         settings_path = get_session_path()
         self.change_mode(self.mode)
+        self._view.set_theme(self.theme)
         with open(settings_path) as f:
             try:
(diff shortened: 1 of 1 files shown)


ZanderBrown:
Well that's sure a weird one

tjguk:
Thanks @dybber. When I get a chance I'll apply that locally and check but on inspection it looks good to merge

tjguk:
I can't test this on Windows as such, but it doesn't seem to cause any issues and the code looks harmess either way. So let's merge and see how it fares in the wild

PR #1064, opened 2020-06-12
New versions of flake8 don't like variables named I, O, or l (E741)

dybber:
Really a minor thing, but `make check` started to fail, after an upgrade of flake8, which are now complaining if you use certain letters as variable names: https://www.flake8rules.com/rules/E741.html


tjguk:
Well that's fussy. Thanks for this as well, @dybber --keep them coming. I'm at work ATM but I'm hoping to have a merge-fest once I can do some due diligence on the changes :)

Issue #1066, opened 2020-06-13
Upgrade briefcase to 0.3.0


dybber:
Briefcase have been revamped, to be more of an independent tool, rather than a plugin for setuptools. It would probably make sense for us to upgrade, here's some instructions for how to upgrade:
https://briefcase.readthedocs.io/en/latest/how-to/upgrade-from-v0.2.html

It seems this will require some effort, e.g. dependencies now also has to be described in a `pyproject.toml` file, and perhaps we could then rework setup to read the dependencies from that file, or perhaps more of setup.py can be moved to the `pyproject.toml` file, and setup.py simplified

tmontes:
(added this note on PR #1067, but should have shared it here, I suppose -- here's some duplication, for the sake of "having the right things in the right place"!) :)

Hey @dybber, I see you're pushing packaging things forward... Nice!

WRT to `briefcase`, you may want to read through my recent 0.3 exploration, something I did when I joined the `briefcase` PyCon US sprint a while back. Here are a few issues I raised which are directly extracted from the experience of packaging Mu with it:

* Fails if dependencies include .pth files | https://github.com/beeware/briefcase/issues/381
* User installable MSI packages? | https://github.com/beeware/briefcase/issues/382
* Packaged Standard Library missing modules | https://github.com/beeware/briefcase/issues/383
* Windows MSI - Append version number to name? | https://github.com/beeware/briefcase/issues/385

While most of these issues are Windows-related, some will apply to macOS too, I suspect. Regardless of that, my vision would be to have `briefcase` address all of Mu's application packaging needs across the supported platforms.

This could simplify the whole packaging process substantially (once it works!), let us keep a single source of truth WRT dependences (all in `pyproject.toml`, no more `setup.*`, no more `pynsist`, no more extra building/packaging scripts to maintain, etc).

Thoughts?

dybber:
Hi @tmontes, thanks for the input and pointers to your write ups. I hadn't discovered that briefcase doesn't bring in pip, venv, and tkinter through its support package. This also affects macOS packaging, and e.g. makes the "Third Party Packages" tab unusable.

I agree the end goal should be to package for all platforms using briefcase, but as I see it now, it will not happen right now, due to the issues you've found in briefcase itself. My suggestion before reading through your writeups, would have been to start switching to briefcase 0.3 for macOS packaging now, and then over time also switch over to use it for the other platforms.

[...]

PR #1067, opened 2020-06-13
Attempt at moving to briefcase-0.3.1

dybber:
This is a first attempt at moving to briefcase-0.3.1.

Known issues:
 - currently dependencies are listed in both setup.py and pyproject.toml, I guess setup.py could be rewritten to parse pyproject.toml and read the settings from there. Alternatively, people seem to suggest switching from setuptools to something like poetry
 - doesn't use the Python from https://github.com/mu-editor/mu_portable_python_macos/releases/download/0.0.6/python3-reduced.tar.gz - as it seems incompatible with the new briefcase (perhaps this is a minor thing - directory layout of the contents of the tarball). However, it works with the Python version briefcase supplies by default. I don't know the history for having our own support package, so I'm not sure what to do here.
 - I added a --no-sign flag in the Makefile, as I don't know anything about code signing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,9 +61,9 @@ script:
   # Run the tests on macOS and package it: "make macos" runs checks+tests first.
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then make macos; fi
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then mkdir dist; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then zip -r -X dist/mu-editor.zip macOS/mu-editor.app; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export DMGFILE=macOS/mu-editor-*.dmg; fi
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then du -sk dist/; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then mv dist/mu-editor.zip dist/mu-editor_$(date '+%Y-%m-%d_%H_%M')_${TRAVIS_BRANCH}_${TRAVIS_COMMIT:0:7}.zip; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then mv "$DMGFILE" dist/mu-editor_$(date '+%Y-%m-%d_%H_%M')_${TRAVIS_BRANCH}_${TRAVIS_COMMIT:0:7}.dmg; fi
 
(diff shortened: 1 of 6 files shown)


dybber:
This relates to issue #1066

dybber:
I didn't change anything related Windows-packaging, still AppVeyor fails. @tjguk, perhaps you should look at the output provided by AppVeyor here.

dybber:
Alternatively, the pyprojects.toml could be auto-generated by setup.py, which is I guess not what is really intended with pyprojects.toml, but well at least an alternative we can consider.

dybber:
I'm asking for a solution at Stackoverflow: https://stackoverflow.com/questions/62362693/how-do-i-read-project-dependencies-from-pyproject-toml-from-my-setup-py-to-avoi

dybber:
To clarify, the 3 issues: duplication of dependencies, change of support package, and no code signing in my original description of the pull request are still relevant. The recent commits, have only been about getting AppVeyor and Travis happy again.

tmontes:
Hey @dybber, I see you're pushing packaging things forward... Nice!

WRT to `briefcase`, you may want to read through my recent 0.3 exploration, something I did when I joined the `briefcase` PyCon US sprint a while back. Here are a few issues I raised which are directly extracted from the experience of packaging Mu with it:

* Fails if dependencies include .pth files | https://github.com/beeware/briefcase/issues/381
* User installable MSI packages? | https://github.com/beeware/briefcase/issues/382
* Packaged Standard Library missing modules | https://github.com/beeware/briefcase/issues/383
* Windows MSI - Append version number to name? | https://github.com/beeware/briefcase/issues/385

While most of these issues are Windows-related, some will apply to macOS too, I suspect. Regardless of that, my vision would be to have `briefcase` address all of Mu's application packaging needs across the supported platforms.

This could simplify the whole packaging process substantially (once it works!), let us keep a single source of truth WRT dependences (all in `pyproject.toml`, no more `setup.*`, no more `pynsist`, no more extra building/packaging scripts to maintain, etc).

Thoughts?

carlosperate:
I was about to add a response here, but noticed the conversation has moved to https://github.com/mu-editor/mu/issues/1066, so adding the link here for visibility.

dybber:
Should we close this, or is anything in this PR relevant for the future of Mu? I guess perhaps moving to pyproject.toml could be something we should do even though we're moving away from briefcase towards pup

dybber:
I don't think it makes sense to keep this PR open, when we're moving to pup anyway, so I will close it now...

PR #1071, opened 2020-06-19
Allow updated version numbers for requirements.

lgautier:
When using python 3.8 (the default python3 on ubuntu 20.04), the version of pgzero in the requirements will not work. A fix on the `pgzero` size seems relatively easy (see issue lordmauve/pgzero#225 and PR lordmauve/pgzero#207), but this also means the version number in mu's `setup.py` file should be updated.

While at it I allowed the latest release for all dependencies and mu appears be passing tests (and working so far).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,24 +21,24 @@ with open(os.path.join(base_dir, "CHANGES.rst"), encoding="utf8") as f:
 
 
 install_requires = [
-    'PyQt5==5.13.2;"arm" not in platform_machine',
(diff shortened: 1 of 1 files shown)


carlosperate:
In the end we resolved this by doing this for the new user virtual environment dependencies:
https://github.com/mu-editor/mu/blob/9179f6415d6531ce43879c3c726a7912fd5af64f/mu/wheels/__init__.py#L41

This should be resolved now, but thank you very much for the PR @lgautier!

PR #1081, opened 2020-07-28
Fix Mu doesn't recognize Seeed Wio Terminal VID / PID #1080

d-c-d:
Mu doesn't recognize Seeed Wio Terminal VID / PID #1080

CircuitPython firmware in Wio Terninal needs to know about its VID/PID
  (0x2886, 0x802D),

This pull request enables serial access from Mu to Seeed Wio Terminal

PR #1036, opened 2020-05-04
xinabox support

luqmaanb:
XinaBox board support added for circuit python and esp32 (1.1.0-alpha.2)

PR #1033, opened 2020-04-29
Add support for Commander from 8086.net

dglaude:
The new board https://www.8086.net/product/commander does not use the Adafruit VID but has another VID/PID.
I made the following addition to this same file of the Mu I use, and it works for me.
This was suggested by @danh from #circuitpython channel on Adafruit Discord.

PR #1021, opened 2020-03-26
Add support for Commander

burtyb:
Add VID:PID for Commander board.

PR #1094, opened 2020-09-05
Add support for requested devices

dybber:
This pull request brings in a lot of new devices for CircuitPython mode and ESP-mode, and replaces several existing pull requests: #855 #916 #947 #975 #1008 #1021 #1032 #1033 #1036 #1081

Thanks to all contributors: @carlosperate, @deshipu, @sabas1080, @KalbeAbbas, @maholli, @burtyb, @dglaude, @tanakamasayuki, @luqmaanb and @d-c-d

The way to list valid devices was recently changed, and instead of getting all pull-requests to adhere to the new interface, I decided to fix all of them in one go. Going forward, the process for adding devices to Mu will be much more swift, please don't hesitate to send in PR's in future, even though it has taken quite some time this time around.

### DIFF
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -49,6 +49,9 @@ class CircuitPythonMode(MicroPythonMode):
         (0x1B4F, 0x8D23, None, "SparkFun SAMD21 Dev Breakout"),
         (0x1209, 0x2017, None, "Mini SAM M4"),
         (0x1209, 0x7102, None, "Mini SAM M0"),
+        (0x04D8, 0xEC72, None, "XinaBox CC03"),
+        (0x04D8, 0xEC75, None, "XinaBox CS11"),
+        (0x04D8, 0xED5E, None, "XinaBox CW03"),
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -45,6 +45,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x1209, 0xBAB1, None, "Electronic Cats Meow Meow"),
         (0x1209, 0xBAB2, None, "Electronic Cats CatWAN USBStick"),
         (0x1209, 0xBAB3, None, "Electronic Cats Bast Pro Mini M0"),
+        (0x1209, 0xBAB6, None, "Electronic Cats Escornabot Makech"),
         (0x1B4F, 0x8D22, None, "SparkFun SAMD21 Mini Breakout"),
         (0x1B4F, 0x8D23, None, "SparkFun SAMD21 Dev Breakout"),
         (0x1209, 0x2017, None, "Mini SAM M4"),
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -52,6 +52,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x04D8, 0xEC72, None, "XinaBox CC03"),
         (0x04D8, 0xEC75, None, "XinaBox CS11"),
         (0x04D8, 0xED5E, None, "XinaBox CW03"),
+        (0x3171, 0x0101, None, "8086.net Commander"),
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -55,6 +55,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x3171, 0x0101, None, "8086.net Commander"),
         (0x04D8, 0xED94, None, "PyCubed"),
         (0x04D8, 0xEDBE, None, "SAM32"),
+        (0x1D50, 0x60E8, None, "PewPew Game Console"),
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -57,6 +57,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x04D8, 0xED94, None, "PyCubed"),
         (0x04D8, 0xEDBE, None, "SAM32"),
         (0x1D50, 0x60E8, None, "PewPew Game Console"),
+        (0x2886, 0x802D, None, "Seeed Wio Terminal"),
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -58,6 +58,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x04D8, 0xEDBE, None, "SAM32"),
         (0x1D50, 0x60E8, None, "PewPew Game Console"),
         (0x2886, 0x802D, None, "Seeed Wio Terminal"),
+        (0x1B4F, 0x0016, None, "Sparkfun Thing Plus - SAMD51"),
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -53,6 +53,8 @@ class CircuitPythonMode(MicroPythonMode):
         (0x04D8, 0xEC75, None, "XinaBox CS11"),
         (0x04D8, 0xED5E, None, "XinaBox CW03"),
         (0x3171, 0x0101, None, "8086.net Commander"),
+        (0x04D8, 0xED94, None, "PyCubed"),
+        (0x04D8, 0xEDBE, None, "SAM32"),
     ]
     # Modules built into CircuitPython which mustn't be used as file names
     # for source code.
--- a/mu/modes/esp.py
+++ b/mu/modes/esp.py
@@ -38,15 +38,25 @@ class ESPMode(MicroPythonMode):
     icon = "esp"
     fs = None
 
-    # There are many boards which use ESP microcontrollers but they often use
-    # the same USB / serial chips (which actually define the Vendor ID and
-    # Product ID for the connected devices.
+    # The below list defines the supported devices, however, many
+    # devices are using the exact same FTDI USB-interface, with vendor
+    # ID 0x403 without reporting their own VID/PID
+
+    # In some instances we can recognize the device not on VID/PID,
+    # but on manufacturer ID, that's what the third column is for.
+    # These more specific device specifications, should be listed
+    # before the generic FTDI VID/PID's
     valid_boards = [
         # VID  , PID,    Manufacturer string, Device name
         (0x1A86, 0x7523, None, "HL-340"),
         (0x10C4, 0xEA60, None, "CP210x"),
-        (0x0403, 0x6015, None, "Sparkfun ESP32 Thing"),
         (0x0403, 0x6001, "M5STACK Inc.", "M5Stack ESP32 device"),
+        (0x0403, 0x6001, None, None),  # FT232/FT245 (XinaBox CW01, CW02)
+        (0x0403, 0x6010, None, None),  # FT2232C/D/L/HL/Q (ESP-WROVER-KIT)
+        (0x0403, 0x6011, None, None),  # FT4232
+        (0x0403, 0x6014, None, None),  # FT232H
+        (0x0403, 0x6015, None, None),  # FT X-Series (Sparkfun ESP32)
+        (0x0403, 0x601C, None, None),  # FT4222H
     ]
 
     def actions(self):
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1612,7 +1612,7 @@ class Editor(QObject):
                 self.device_disconnected.emit(connected)
                 logger.info(
                     (
-                        "{} device connected on port: {}"
+                        "{} device disconnected on port: {}"
                         "(VID: 0x{:04X}, PID: 0x{:04X}, manufacturer {})"
                     ).format(
                         connected.short_mode_name,


dybber:
Just tested that we also support M5StickV and M5Stack UnitV - however, they like many others devices, reuse the FTDI VID/PID and thus show up as "ESP MicroPython".

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,19 @@ class CircuitPythonMode(MicroPythonMode):
...
         (0x1209, 0xBAB2, None, "Electronic Cats CatWAN USBStick"),
         (0x1209, 0xBAB3, None, "Electronic Cats Bast Pro Mini M0"),
+        (0x1209, 0xBAB6, None, "Electronic Cats Escornabot Makech"),
         (0x1B4F, 0x8D22, None, "SparkFun SAMD21 Mini Breakout"),
         (0x1B4F, 0x8D23, None, "SparkFun SAMD21 Dev Breakout"),
         (0x1209, 0x2017, None, "Mini SAM M4"),
         (0x1209, 0x7102, None, "Mini SAM M0"),
+        (0x04D8, 0xEC72, None, "XinaBox CC03"),
+        (0x04D8, 0xEC75, None, "XinaBox CS11"),
+        (0x04D8, 0xED5E, None, "XinaBox CW03"),
```
carlosperate, line 55: @KalbeAbbas are these PIDs sublicense from Microchip? Or are they share with Microchip or other products?

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,20 @@ class CircuitPythonMode(MicroPythonMode):
         (0x1209, 0xBAB1, None, "Electronic Cats Meow Meow"),
         (0x1209, 0xBAB2, None, "Electronic Cats CatWAN USBStick"),
         (0x1209, 0xBAB3, None, "Electronic Cats Bast Pro Mini M0"),
+        (0x1209, 0xBAB6, None, "Electronic Cats Escornabot Makech"),
```
carlosperate, line 48: This one looks good, thanks @sabas1080!
http://pid.codes/1209/BAB6/

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,20 @@ class CircuitPythonMode(MicroPythonMode):
...
         (0x1B4F, 0x8D22, None, "SparkFun SAMD21 Mini Breakout"),
         (0x1B4F, 0x8D23, None, "SparkFun SAMD21 Dev Breakout"),
         (0x1209, 0x2017, None, "Mini SAM M4"),
         (0x1209, 0x7102, None, "Mini SAM M0"),
+        (0x04D8, 0xEC72, None, "XinaBox CC03"),
+        (0x04D8, 0xEC75, None, "XinaBox CS11"),
+        (0x04D8, 0xED5E, None, "XinaBox CW03"),
+        (0x3171, 0x0101, None, "8086.net Commander"),
```
carlosperate, line 56: This looks good too, thanks @burtyb and @dglaude!
https://www.the-sz.com/products/usbid/index.php?v=0x3171

mu/modes/circuitpython.py
carlosperate: Btw, these PID/VID have only been added to CircuitPython, is this what you expected? Do these boards also support standard MicroPython? If so, which mode would you prefer to support?

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,20 @@ class CircuitPythonMode(MicroPythonMode):
...
+        (0x04D8, 0xED94, None, "PyCubed"),
+        (0x04D8, 0xEDBE, None, "SAM32"),
```
carlosperate, line 58: Thanks for the original submission @maholli!
Are these PIDs sublicense from Microchip? Or are they shared with Microchip or other products?

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,20 @@ class CircuitPythonMode(MicroPythonMode):
...
+        (0x1D50, 0x60E8, None, "PewPew Game Console"),
```
carlosperate, line 59: This looks good too, thanks @deshipu!
https://raw.githubusercontent.com/openmoko/openmoko-usb-oui/master/usb_product_ids.psv

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,20 @@ class CircuitPythonMode(MicroPythonMode):
...
+        (0x2886, 0x802D, None, "Seeed Wio Terminal"),
```
carlosperate, line 60: This looks good too, thanks @d-c-d!
https://www.the-sz.com/products/usbid/index.php?v=0x2886

mu/modes/circuitpython.py
```diff
@@ -45,10 +45,20 @@ class CircuitPythonMode(MicroPythonMode):
...
+        (0x1B4F, 0x0016, None, "Sparkfun Thing Plus - SAMD51"),
```
carlosperate, line 61: Thanks for the original submission @nezra. Does this board also support MicroPython?

mu/modes/circuitpython.py
carlosperate: How are the CC03 and CS11 modules connected to the computer? Is there a USB module that needs to be plugged? If so, is that module just routing USB pins into a USB connector, or do they have additional USB device?
I couldn't find any info on the CW03, do you have any links?

https://xinabox.cc/products/cc03
https://xinabox.cc/products/cs11

mu/modes/esp.py
```diff
@@ -38,15 +38,25 @@ class ESPMode(MicroPythonMode):
...
+    # but on manufacturer ID, that's what the third column is for.
+    # These more specific device specifications, should be listed
+    # before the generic FTDI VID/PID's
     valid_boards = [
         # VID  , PID,    Manufacturer string, Device name
         (0x1A86, 0x7523, None, "HL-340"),
         (0x10C4, 0xEA60, None, "CP210x"),
-        (0x0403, 0x6015, None, "Sparkfun ESP32 Thing"),
         (0x0403, 0x6001, "M5STACK Inc.", "M5Stack ESP32 device"),
+        (0x0403, 0x6001, None, None),  # FT232/FT245 (XinaBox CW01, CW02)
```
carlosperate, line 54: @dybber is this going to be recognised as a M5Stack?

mu/modes/esp.py
```diff
@@ -38,15 +38,25 @@ class ESPMode(MicroPythonMode):
...
+        (0x0403, 0x6010, None, None),  # FT2232C/D/L/HL/Q (ESP-WROVER-KIT)
```
carlosperate, line 55: I think the main issue with this case was that the same device creates 2 virtual serial ports, and both will of course contain the same VID/PID pair.

@dybber  I believe with the latest changes in master two devices would appear in the dropdown with the same name, is that right? So a user could switch between the two of them, even if they look identical?

mu/modes/esp.py
dybber: Yes, the idea is that if a device matches multiple lines, the name reported will be the first in the list. This means that a device with VID/PID 0x0403 0x6001, and "M5Stack Inc." as manufacturer will be shown as "M5Stack ESP32 device", whereas anyother device with the same VID/PID will just be shown as "ESP MicroPython"

Here is a screenshot where I had a M5StickC device and a Sparkfun ESP32 Thing connected simultaneously (VID/PID 0x403 0x6015):

Screenshot 2020-09-05 at 18 59 55

mu/modes/esp.py
dybber: I actually just checked, and seems like that is still an issue. Will investigate, the intention is that two identical devices, each of them will show up in the menu, with the same name, but different ports will be shown when you hover over them.

dybber: I was wrong just before, it does work as intended. Two devices show up:

Screenshot 2020-09-05 at 19 12 03

And you can determine which port by hovering:
Screenshot 2020-09-05 at 19 12 47

mu/modes/circuitpython.py
dybber: I don't believe so.

mu/modes/circuitpython.py
dybber: For the SAM32 it seems like it, here's another PR from @maholli: https://github.com/adafruit/uf2-samdx1/blob/2e5ac4999a79cc2e5a840291d9f85e003230b2dd/boards/sam32/board_config.h#L11

mu/modes/circuitpython.py
dybber: A programmer is necessary, either Xinabox IP02 or Xinabox IP03. @KalbeAbbas has written a guide here: https://www.hackster.io/PragmaticPhil/loading-circuitpython-onto-an-xinabox-programmable-core-544749

Thus it might be that it's actually the IP02/IP03 which are recognized, but I don't believe these programmers are used for anything else than Xinabox devices.

I can't find info on the CW03 either, I guess it's no longer in production.

MicroPython seems to support the SAMD21 processor: https://github.com/micropython/micropython/tree/master/ports/samd

mu/modes/esp.py
dybber: Perhaps it should say "ESP MicroPython device" instead? Currently when the name is `None`, it defaults to the mode name. It could be mode name with `" device"` appended.

mu/modes/circuitpython.py
carlosperate: Great, I'll mark the comment as resolved.

mu/modes/esp.py
carlosperate: I guess I was surprised the manufacturer string was configured in FTDI chips.
So then in an M5Stack ESP32 the [QSerialPortInfo.manufacturer()](https://doc.qt.io/qt-5/qserialportinfo.html#manufacturer) returns "M5STACK Inc."?

mu/modes/esp.py
carlosperate: Yeah, "device" would make it more clear, but maybe it would be more useful to have the port name in brackets? To be able to tell devices apart.
Although we would have to shorten non-Windows ports, `COMXX` is short but the unix paths or port names can be quite long.

mu/modes/circuitpython.py
carlosperate: Great, thanks Martin!

mu/modes/circuitpython.py
carlosperate: So it looks like IP03 is just a breakout for the SWD/JTAG pins on the microcontroller modules.

IPO2 actually has an FTD232RL (by default 0x0403, 0x6001), so while they could be getting these VID/PID preprogrammed from factory by Microchip, I'm not sure how the FTDI module would change these values depending on the microcontroller module plugged. Also I think ths might be just a USB-to-UART chip.

I guess it is quite possible the IP03 module also has USB data pass-through to the microcontroller modules.

Maybe @pragmaticphil (Hi Phil! 👋 ) could also point us to the CW03 device info?

mu/modes/esp.py
dybber: Yes, that's a bit confusing, but I guess it's just because they don't want to pay the $5000 to get their own Vendor ID.

mu/modes/esp.py
dybber: Yes, I guess when multiple devices of the same type are connected, it would help to differentiate them, without having to mouseover. Teaching wise, it's probably best to just show the entire port-name, instead of trying to hide parts of it. That would just introduce confusion.

mu/modes/circuitpython.py
gotfredsen: **Re: CS11 and CC03**
Yes, the CS11 and CC03 is only supported via CircuitPython, not MicroPython. They have their own VID/PID and I believe has to be included like this:
```
add /Applications/mu-editor.app/Contents/Resources/app/mu/modes/circuitpython.py line 49:
    (0x04D8, 0xEC72),  # XinaBox CC03
    (0x04D8, 0xEC75),  # XinaBox CS11
```
the CW03 is not ready yet, so let us forget about that one for now. And correct, the IP03 just connected standard Micro-USB connector to the USB via a small connector to the CS11/CC03. It does not have any USB device on it. 

**Re: CW01 and CW02**
These two are using UART to USB chip, being the standard FTDI chip, with its own VID/PID combination. The FTDI chip sits on a separate module, either IP01 or IP02, they both use the same FTDI chip. The CW01 is based on the ESP8266 and the CW02 is based on the ESP32, and they both support the standard MicroPython implementation.

I believe the way these are being supported is like this
```
add to modes/esp.py line 47:
     (0x0403, 0x6001)   # XinaBox CW01, CW02 (FTDI) 
```

Hope this clarifies things a bit.

dybber: Hi @gotfredsen, thanks for the clarification. However, how devices are added to Mu has changed slightly, for example see https://github.com/mu-editor/mu/blob/master/mu/modes/circuitpython.py#L39

Or the proposed changes in this PR: https://github.com/mu-editor/mu/pull/1094/commits/b1764ec2ef4ef754a59ebc92ffd43825bd65d6db

When this PR is merged "XinaBox CC03" and "XinaBox CS11" will show up in the user-interface, so the user can see the name of the connected devices, however for CW01/CW02 we will not be able to detect which device it is, so they will just show up as "CircuitPython device" or similar, as long as they don't have their own VID/PID or a unique manufacturer-string.

gotfredsen: @dybber, I have accepted the propose change in PR #b1764ec

And with regards to CW01, CW02 and the use FTDI, is that also ok?

dybber: That change is also part of it, I forgot to link to it: https://github.com/mu-editor/mu/pull/1094/commits/75b65c577c3bc5459040c27b71b84b3319062a30#diff-e78b8abb0346fee662c3583eb227d9a6R54

dybber: The remaining questions is what @carlosperate wrote to begin with: 
> Are these PIDs sublicense from Microchip? Or are they shared with Microchip or other products? 

(For CC03, CS11, CW03)

gotfredsen: @dybber, I have accepted the propose change in PR #75b65c5 as well.

The PIDs are sublicensed from Microchip and are not shared with any other devices, not ours nor others.

dybber: Okay, I guess we can mark this as resolved then. Thanks!

mu/modes/esp.py
dybber: I will take a look at this when I can find the time, perhaps friday evening

gotfredsen: Just as a suggestion. We have a flashing tool we use to load precompiled source onto our CW01 and CW02. The tool is also done in Python. We use ESPTOOL, which as you know also is in Python, to flash the code, but we first use the tool to inquire if its a ESP8266 or a ESP32 plugged in, and we show the MAC address at the same time:
Screenshot 2020-09-09 at 12 20 40 PM

We also use, as shown in this example, the tool to load MicroPython to the xChips.

mu/modes/esp.py
carlosperate: So the FTDI chips in the M5 Stack have the USB manufacturer string set to "M5STACK Inc."?
I don't have one to check, so just wanted to make sure that is the case.

mu/modes/esp.py
carlosperate: I was thinking a string in the dropdown as `ESP MicroPython (/dev/cu.usbmodem1234567)` might be too long, but maybe that's fine if it helps avoid confusion.

I'll create a ticket to track the work and mark this as resolved, thanks Martin!

mu/modes/esp.py
dybber: Exactly, and otherwise (if they change that in the future), it will just show up as ESP MicroPython.

mu/modes/esp.py
dybber: Ok. How do you detect that it's a CW02 chip, when it just reuses FTDI's VID/PID (0x0403, 0x6001), and it doesn't have its own?

dybber: It will perhaps be even longer, if it becomes: `ESP MicroPython device (Port: /dev/tty/cu.usbmodem1234567)`, but let's test the idea.

mu/modes/circuitpython.py
carlosperate: Thank you @gotfredsen that definitely helps a lot!

I believe the `0x04D8` VID belongs to Microchip, and they do sublicense PIDs to third parties, but unfortunately there doesn't seem to be a public list online to be able to check this.
Are these PIDs sublicensed for these products?  Just want to make sure there might not be clashes with other USB devices with the same values.

mu/modes/esp.py
carlosperate: I guess on connection we can query the REPL to get the device info, but that would stop the user code every time the device is "detected".

mu/modes/circuitpython.py
carlosperate: My PR view wasn't refreshed when I sent my last message (I'm used to GH keeping the tabs synced, but that didn't happen this time).
My question has been answer already, sorry for the duplication!

mu/modes/esp.py
carlosperate: Great, thanks Martin!

mu/modes/esp.py
gotfredsen: > Ok. How do you detect that it's a CW02 chip, when it just reuses FTDI's VID/PID (0x0403, 0x6001), and it doesn't have its own?

Again, just a suggestion, if you want to know which ESP connected to the USB/UART controller: We simply fire off a `python -m esptool --port /dev/cu.usbserial-<actual-port> read_flash` and then scan for `Detecting chip type... ESP32`. We are _not_ determining whether it is a CW01 or CW02 based on VID/PID, we only use that t check that a FTDI chip is connected to the USB port, so the above suggestion is after that. But as I said, just a suggestion. I am happy that it just say ESP MicroPython, but you had an issue trying to determine what was inserted more specifically and this could help.

mu/modes/circuitpython.py
gotfredsen: > Thank you @gotfredsen that definitely helps a lot!
> 
> I believe the `0x04D8` VID belongs to Microchip, and they do sublicense PIDs to third parties, but unfortunately there doesn't seem to be a public list online to be able to check this.
> Are these PIDs sublicensed for these products? Just want to make sure there might not be clashes with other USB devices with the same values.

@carlosperate, the PIDs we got from Microchip are sub-licensed to us and are not used with any other device, neither from us, nor from anybody else. They are unique to our CS11, CC03 and CW03.

mu/modes/esp.py
carlosperate: Yeah, taking in consideration the long delays in reading the flash or querying the REPL it sounds like its fine to identify the devices as generic ESP MicroPython.
Thanks again everybody for the clarifications and discussion!

carlosperate:
I've restarted the travis builds.
macOS is failing for an issue outside of this PR, and I had to restart the Linux Py3.7 one a couple of times as it had some issues with the coverage module, but ended up passing.

carlosperate (approved):
All comments resolved, great work sorting all the new devicces @dybber!

sabas1080:
@dybber thanks!

PR #1097, opened 2020-09-10
Travis: Update Xcode version to 9.4 to fix brew build issues.

carlosperate:
This version still stays in macOS 10.13, so general build compatibility should be the same.
Turns out the macOS build failures were due to issues building stuff with brew and this tiny update fixes it.


carlosperate:
Thanks Tim!

PR #1100, opened 2020-09-18
ESP Firmware flasher device selection

dybber:
This PR replaced #1015. It is a rebased version on top of master, where I've added the device selector, so the user can decide which device to flash.

Thanks to @KKawase0104 for the majority of this work. @KKawase0104, could you perhaps test this version, then we will soon be ready to merge it into master.

Screenshots:

Screenshot 2020-09-18 at 19 25 12
Screenshot 2020-09-18 at 19 25 19
Screenshot 2020-09-18 at 19 25 30


KKawase0104:
Hello.

I have checked on Windows and ESP32.

When I try to write the firmware without a USB cable connected to the PC and ESP32, I get a problem.
None will be returned in the following process in the update_firmware method of dialogs.py, and the subsequent process will fail and Mu will fall.

`device = self.device_selector.selected_device()`

I think it needs to be fixed to either display a message or disable the button depending on the state of the USB cable connection.

Otherwise, I have checked that it works as intended.
- USB serial behavior when connected or unconnected.
- Writing and starting MicroPython
- behavior when the esptool package is not installed
- behavior when the esptool package is installed
- ESP8266 selected and write to esp32 board failed
- Firmware write failure with incorrect file name

Thank you.

dybber:
Good catch!

I think the "Erase & write firmware" button should be disabled when you are missing either the device or selecting .bin file, so I've tried to add that, as well as adding a bit to the stylesheet to make it visible when a button is disabled. 

However, I'm not sure I've got the colors for the disabled button right. Here are some screenshots from a first attempt (included in the commit)

Enabled, day.css
Screenshot 2020-09-23 at 08 17 40

Disabled, day.css
Screenshot 2020-09-23 at 08 10 26

Enabled, contrast.css
Screenshot 2020-09-23 at 08 17 17

Disabled, constrast.css
Screenshot 2020-09-23 at 08 16 41

Enabled, night.css
Screenshot 2020-09-23 at 08 20 55

Disabled, night.css
Screenshot 2020-09-23 at 08 21 06

KKawase0104:
Hello.
I have confirmed that it has been fixed.
I could see that contrast and night mode are invalid indications.
Day mode might work better if the text is white.

Thank you.

dybber:
Hi @KKawase0104, what do you mean with "invalid indications"?

Here is "day mode" with white text on the disabled button:

Screenshot 2020-09-23 at 12 33 54

tjguk:
I'm certainly no expert, and I think that any user will quickly work out that -- whatever the visible difference is -- it means "disabled". It's a fairly common UI idiom to disable a button until some things are completed, so I doubt many people will be fooled.

However, I'm not convinced by the white text suggested for day mode. Perhaps a very light grey? Anyway, I have no very strong feelings, and we can always change it later if we get negative feedback

KKawase0104:
Hello.
@dybber, Thanks for the change, but after @tjguk 's suggestion, I found it inconsistent that only day mode flips the color of the text. I'm sorry for making you change it so many times, but I think the text color should be black. I may have been too concerned about it.

@tjguk Thank you for your good suggestion.

Thank you.

dybber:
I didn't commit the change, only made the screenshot, so no worries :-)

tmontes:
> However, I'm not convinced by the white text suggested for day mode. Perhaps a very light grey? Anyway, I have no very strong feelings, and we can always change it later if we get negative feedback

...generally agree: lighter gray for the "light theme", darker white for the "dark themes" would be my intuition (as if the button itself became mostly transparent such that its own colours shift towards the background).

dybber:
Rebased on top of master, to make it mergeable.

dybber:
Thanks for contribution @KKawase0104, the firmware flasher will now be part of the next Mu release

KKawase0104:
@dybber 
I'm happy to contribute to Mu with your leading.
Thank you.

PR #1102, opened 2020-09-20
Update tests for environments without QtCharts

carlosperate:
The charts feature is optional in Mu, as it is not available in all platforms. This is mostly for Raspberry Pi with Raspbian Stretch, but other Arm-based devices like Chromebooks can have this dependency missing.

Right now the tests fail in those environments as they try to exercise the plotter feature, so this PR fixes that by skipping the tests if it cannot import QtChart. It also adds a couple of extra missing tests for checking the modes button creation with and without the charts available.

### DIFF
--- a/tests/interface/test_panes.py
+++ b/tests/interface/test_panes.py
@@ -3,18 +3,19 @@
 Tests for the user interface elements of Mu.
 """
 from PyQt5.QtWidgets import QMessageBox, QLabel
-from PyQt5.QtChart import QChart, QLineSeries, QValueAxis
 from PyQt5.QtCore import Qt, QEvent, QPointF
 from PyQt5.QtGui import QTextCursor, QMouseEvent
+from collections import deque
 from unittest import mock
 
 import sys
 import os
 import signal
-import mu
+import pytest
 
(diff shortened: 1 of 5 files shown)


dybber:
Looks good to me, but wouldn't the coverage check still fail on ARM devices, as part of the code will not be covered by any test-cases?

carlosperate:
Thanks Martin!
About the coverage, yes, the coverage will fall below 100% on the Arm tests, but I believe that doesn't currently cause a failure in the CI.

Alternatively we could mock QtCharts (and related dependencies), but we already have a lot of mocking going around and I think having Arm (only) under 100% coverage is not that bad.

dybber:
Okay, I don't have any problems with coverage decreasing below 100% on ARM specifically, it was more a worry about the CI and if whether they will start failing, but I guess we will have to try.

dybber:
Rebased this on top of master and removed the `qtapp` dependencies from in the process

carlosperate:
Thanks Martin!
Let's leave this open for a few more days more in case other people have any thoughts about using the test skips.

Issue #826, opened 2019-05-07
requirements.txt problem for ARM architecture


askvictor:
The requirements.txt includes some lines such as `qscintilla; "arm" not in platform_machine` which exclude the requirements for ARM devices. However, ARM can appear to python as aarch64 (at least my ARM Chromebook running crostini shows this when I call `platform.machine`). Is this exclusion supposed to be for all ARM devices?

carlosperate:
Yes, the exclusion is meant for all ARM devices, but because the string returned by `platform.machine` can vary broadly, is going to be difficult to catch all possibilities.

Checking for `arm` being inside the string was just a way that worked in Raspbian with Raspberry Pi 1/Zero (with armv6 processor) and Pi 2 (armv7 and armv8/arm64) and Pi 3 (armv8/arm64). I don't think we have tried this is any other Arm-based platform, so not sure what other strings we can expect apart from `aarch64` or something with `arm` in the title.

Looks like  `aarch64` is also a fairly common string we should be checking as well (https://stackoverflow.com/questions/31851611/differences-between-arm64-and-aarch64) although I'm not sure what would be the best way to do that. We'll have to check if there is an "or" operator available in environmental markers: https://www.python.org/dev/peps/pep-0508/

askvictor:
Looks like environmental markers do support "or". I've just tried `"arm" not in platform_machine and "aarch" not in platform_machine` which works fine to exclude those packages on those platforms. Might be worth updating the docs to specify ARM devices rather than just Raspberry Pi

carlosperate:
@askvictor this was automatically closed with PR https://github.com/mu-editor/mu/pull/1104, but it'd be great if you'd be able to check if it works in chromebooks?

PR #1104, opened 2020-09-21
Travis: Run tests in arm64 architecture

carlosperate:
This is not ready for merge quite yet, but it is getting there, it needs PR https://github.com/mu-editor/mu/pull/1102 to be merged first.

This PR adds the arm64 architecture to Travis CI to run the tests. This will be useful to help make sure we don't break things on Raspberry Pi, and the environmental markers edited in this PR should also enable being able to run Mu in other single board computers, possibly on Arm Chromebooks, and on systems like the [Pinebook Pro](https://www.pine64.org/pinebook-pro/).

Also fixes https://github.com/mu-editor/mu/issues/826.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ matrix:
       python: 3.5
       services:
         - xvfb
+      env:
+        - QT_QPA_PLATFORM="offscreen"
     - os: osx
       # match xcode version with earliest supported by current PyQt 
       # (earliest to maximize compatibility)
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,30 @@ matrix:
       sudo: required
       language: python
       python: 3.5
+      services:
+        - xvfb
     - os: linux
       dist: xenial 
       sudo: required
       language: python
       python: 3.6
+      services:
+        - xvfb
     - os: linux
       dist: xenial
       sudo: required
       language: python
       python: 3.7
+      services:
+        - xvfb
+    - os: linux
+      dist: xenial
+      arch: arm64
+      sudo: required
+      language: python
+      python: 3.5
+      services:
+        - xvfb
     - os: osx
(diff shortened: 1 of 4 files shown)


carlosperate:
Currently failing from updates introduced in https://github.com/mu-editor/mu/pull/1100.
@dybber do you know if there is an alternative to `QtProcess.errorOccurred()` (introduced in Qt5.6) available in Qt5.5.1?

dybber:
Seems like it was renamed from `QProcess.error` to `QProcess.errorOccurred` at some point.

See: https://doc.qt.io/archives/qt-5.5/qprocess.html

And now: https://doc.qt.io/qt-5/qprocess.html

carlosperate:
Thanks Martin!
With the changed signal method name the tests seems to work fine.  However I am not sure how to manually trigger an error to test that code-path. I tried disconnecting the board during flashing and raising an exception within the esptool itself by editing the module. The errors caused are captured correctly in the text area, but not via `esptool_error()` method call.

I guess the signal only gets raised when the actual process crashes, any thoughts on how that could be forced?

dybber:
I'm on my phone this weekend, so I can only browse the documentation. 

What if the esptool just isn't there, then it should fail with a `QProcess.FailedToStart`

dybber:
Just an idea, what if you change the `PATH` environment variable, to provoke the `QProcess.FailedToStart`?

carlosperate:
Changing Path, not having esptool available when it tries to running or forcing Mu to look at the wrong module all caused issues that were caught somewhere else before the QProcess.error could be trigger.

However, manually editing esptool and using [ctypes to make the Python process crash](https://stackoverflow.com/a/42544856/775259) did the trick, and I can confirm that both `QProcess.error` and `QProcess.error` capture that crash.

dybber:
Great!

PR #1106, opened 2020-09-25
Tests: Add more qtapp fixtures to fix random test crashes

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/1105, thanks @dybber for the tip about adding the fixture!

I've tested this locally in macOS and I haven't been able to get the same error when the affected tests are run in isolation (which before this patch always triggers the issue), or when running all the tests in a loop for about 1h.

Currently as a draft PR while I restart the tests in travis to ensure nothing else pops up. I also need to remove the first two commits (useful for Travis debugging) before we can merge this.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,11 @@
 import pytest
 from PyQt5.QtWidgets import QApplication
 
+# Keep global reference to avoid being garbage collected
+_qapp_instance = None
 
-@pytest.fixture(scope="session")
+
+@pytest.fixture(scope="session", autouse=True)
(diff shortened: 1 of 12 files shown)


carlosperate:
I've restarted the CI quite a few times and been running the tests in a while loop for more than half an hour and haven't encountered these issues anymore.
I'll remove the commits adding the -s and -q flags to pytests and this PR is ready.

dybber:
Looks good to me!

carlosperate:
Thanks Martin!

Issue #1099, opened 2020-09-17
mu on Linux Mint 20 - in case someone will have a problem


abixadamj:
This is only **INFORMATION**
_mu 1.0.3 is able to install on Linux Mint 20._ `pip3 install mu-editor==1.0.3`

Before I have installed some packages:
```
apt install --yes python3-freetype libfreetype-dev libfreetype6-dev
apt install --yes python3-pygame-sdl2 python3-dev
apt install --yes libsdl-image1.2-dev libsdl-mixer1.2-dev libsdl-ttf2.0-dev libsdl1.2-dev libsmpeg-dev
apt install --yes libportmidi-dev ffmpeg libswscale-dev libavformat-dev libavcodec-dev libfreetype6-dev
```

After that, I run the following:

```
root@zs-ceglow:~# pip3 install mu-editor==1.0.3
Collecting mu-editor==1.0.3
  Downloading mu_editor-1.0.3-py3-none-any.whl (2.0 MB)
     |████████████████████████████████| 2.0 MB 1.2 MB/s 
Collecting pyqt5==5.14.1
  Downloading PyQt5-5.14.1-5.14.1-cp35.cp36.cp37.cp38-abi3-manylinux2014_x86_64.whl (63.5 MB)
     |████████████████████████████████| 63.5 MB 26.1 MB/s 
Requirement already satisfied: Pillow>=5.2.0 in /usr/lib/python3/dist-packages (from mu-editor==1.0.3) (7.0.0)
Collecting nudatus>=0.0.3
  Downloading nudatus-0.0.4.tar.gz (4.5 kB)
Collecting semver>=2.8.0
  Downloading semver-2.10.2-py2.py3-none-any.whl (12 kB)
Collecting gpiozero>=1.4.1
  Downloading gpiozero-1.5.1-py2.py3-none-any.whl (138 kB)
     |████████████████████████████████| 138 kB 64.8 MB/s 
Collecting pycodestyle==2.4.0
  Downloading pycodestyle-2.4.0-py2.py3-none-any.whl (62 kB)
     |████████████████████████████████| 62 kB 947 kB/s 
Collecting qscintilla==2.11.4
  Downloading QScintilla-2.11.4-1-cp35.cp36.cp37.cp38-abi3-manylinux2014_x86_64.whl (1.9 MB)
     |████████████████████████████████| 1.9 MB 35.9 MB/s 
Collecting qtconsole==4.3.1
[...]

Issue #999, opened 2020-02-04
Installation in ubuntu 18.04 python 3.8


waqarrashid33:
I am been trying to install the mu-editor on ubuntu 18.04 but so far haven't been successful. I am getting the following error and I think I need to install some dependencies. It was giving error about SDL but then I searched on internet and somehow installed the .so file 
```
    ERROR: Command errored out with exit status 1:
     command: /usr/local/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-amxp4hv0/pygame/setup.py'"'"'; __file__='"'"'/tmp/pip-install-amxp4hv0/pygame/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-amxp4hv0/pygame/pip-egg-info
         cwd: /tmp/pip-install-amxp4hv0/pygame/
    Complete output (18 lines):    
    
    WARNING, No "Setup" File Exists, Running "buildconfig/config.py"
    Using UNIX configuration...
    
    Hunting dependencies...
    SDL     : found 1.2.15
    FONT    : not found
    IMAGE   : not found
    MIXER   : not found
    PNG     : found
    JPEG    : not found
    SCRAP   : found
    PORTMIDI: not found
    PORTTIME: not found
    FREETYPE: found 21.0.15
    Missing dependencies
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```
Below is the output for `python3 setup.py egg_info`. I think the problem is probably straight forward but I am not well versed in python so I would appreciate any help.

```
python3 setup.py egg_info
running egg_info
creating mu_editor.egg-info
writing mu_editor.egg-info/PKG-INFO
writing dependency_links to mu_editor.egg-info/dependency_links.txt
writing entry points to mu_editor.egg-info/entry_points.txt
writing requirements to mu_editor.egg-info/requires.txt
writing top-level names to mu_editor.egg-info/top_level.txt
writing manifest file 'mu_editor.egg-info/SOURCES.txt'
package init file 'mu/debugger/__init__.py' not found (or not a regular file)
reading manifest file 'mu_editor.egg-info/SOURCES.txt'
[...]

PR #1108, opened 2020-09-27
Package: Add PEP 345 Python version required.

carlosperate:
Another tiny PR.
Adding this metadata ensures PyPI, and other indexes like piwheels, don't install Mu in incompatible Python version.
This would prevent issues like https://github.com/mu-editor/mu/issues/999 and https://github.com/mu-editor/mu/issues/1099.


dybber (approved):

carlosperate:
Merging this, even if we decide to remove it in the future, as we are having more people encountering issues trying to pip install Mu in Python 3.8.

PR #1116, opened 2020-09-29
Show splash screen before mode selector, and fix theming on first run (fixes #429)

### LINKED ISSUE #429: Splash screen is shown after mode is selected
Just wondering if this is the desired behaviour.

* Install Mu
* Open Mu
* Window opens asking the user to select mode -> select any mode
* The mode window closes and then the splash screen opens

I would probably expect the splash screen to be shown before any action is taken on any of the Mu windows.

Tested from source on latest master at the time of writing (56fc1a3a83822ce30585d6adde1a646f96545c5d) on Raspbian Stretch.

dybber:
An attempt at fixing issue #429. However, I'm not certain how to add tests for the 3 lines which are currently not tested. Any suggestions for how to test that and get coverage back to 100% would be helpful.

The problem is that we need to keep calling `raise_()` on the splash screen, if we open them in the opposite order, otherwise the mode selection will still open on top of it.

The relevant lines from `app.run` is here, where my question is how to test `remove_splash`. I will try to look at it myself some other time.

```python
    # Display a friendly "splash" icon.
    splash = QSplashScreen(load_pixmap("splash-screen"))
    # splash.setWindowFlags(Qt.WindowStaysOnTopHint)
    splash.show()

    # Make sure the splash screen stays on top while
    # the mode selection dialog might open
    raise_splash = QTimer()
    raise_splash.timeout.connect(lambda: splash.raise_())
    raise_splash.start(10)

    # Hide the splash icon.
    splash_be_gone = QTimer()
    def remove_splash():
        splash.finish(editor_window)
        raise_splash.stop()
    splash_be_gone.timeout.connect(remove_splash)
    splash_be_gone.setSingleShot(True)
    splash_be_gone.start(2000)
```

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -136,26 +136,6 @@ def run():
     # unless this flag is set
     app.setAttribute(Qt.AA_UseHighDpiPixmaps)
 
-    # Display a friendly "splash" icon.
-    splash = QSplashScreen(load_pixmap("splash-screen"))
-    # splash.setWindowFlags(Qt.WindowStaysOnTopHint)
-    splash.show()
-
-    # Make sure the splash screen stays on top while
-    # the mode selection dialog might open
-    raise_splash = QTimer()
-    raise_splash.timeout.connect(lambda: splash.raise_())
-    raise_splash.start(10)
-
-    # Hide the splash icon.
-    splash_be_gone = QTimer()
-    def remove_splash():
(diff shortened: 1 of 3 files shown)


dybber:
Okay, found a way to write the test, but became pretty big for such a small thing. Had to make my own QTimer replacement and throw that in.

Issue #1117, opened 2020-09-29
switching between devices does't switch default file system 


CarlFK:
version: today's git master 

This is messy.  I suspect getting things to work right is going to be quite a bit of work as it isn't easy to figure out what usb device provides what tty and storage device.

I suspect the solution is make it easy for one device so that beginners have a good experience,  and consider 2 devices "expert mode" and give the user more options.

I'm not even sure if this should be 1 issue or many.  
so, here is one of the current problems:  

2 CPX boards running CircuitPython, mu sees 2, but switching between the 2 doesn't change the default when I click "Load":

log shows:
```
2020-09-29 11:27:04,825 - mu.logic:833(check_usb) INFO: circuitpython device connected on port: /dev/ttyACM2(VID: 0x239A, PID: 0x8019, manufacturer: 'Adafruit Industries LLC')
2020-09-29 11:27:04,826 - mu.logic:833(check_usb) INFO: circuitpython device connected on port: /dev/ttyACM0(VID: 0x239A, PID: 0x8019, manufacturer: 'Adafruit Industries LLC')
```

Select the ACM0 device, click load:
`2020-09-29 11:29:56,722 - mu.logic:1198(get_dialog_directory) INFO: Using path for file dialog: /media/carl/CIRCUITPY1
`
Select the ACM2 device, click load:
`2020-09-29 11:32:32,422 - mu.logic:1198(get_dialog_directory) INFO: Using path for file dialog: /media/carl/CIRCUITPY1`

In both cases, the dialog shows the same CPX.

It does switch between the 2 device's serial (good).

dybber:
Oh yeah, that can indeed get complicated. It's probably not easy to figure out which serial device and which volume corresponds to the same device. I guess you can't tell a CircuitPython or micro:bit (e.g. through serial) to change its volume name to something more specific?

I guess we could change the CircuitPython and micro:bit modes to transfer files through the serial interface, like in the ESP MicroPython mode, then we wouldn't have that problem. I'm no expert in CircuitPython and micro:bit, however, so perhaps someone else should chip in with ideas for possible solutions.

Currently everything should work for only one device, so at least the large majority will not have any problems.

CarlFK:
> Oh yeah, that can indeed get complicated. It's probably not easy to figure out which serial device and which volume corresponds to the same device.

Basically.  it is possible, but likely difficult to work on most os/distro/platform reliably. 


[...]

PR #1118, opened 2020-09-29
Add check for PYBFLASH.

CarlFK:
the MP device I have has a volume label of PYBFLASH, and a 2nd CP device will show up as CIRCUITPY1.

this PR helps with these cases. 

carl@twist:~/src/mu/carlfk/mu$ mount |grep media 
/dev/sdc1 on /media/carl/PYBFLASH type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)
/dev/sdd1 on /media/carl/CIRCUITPY type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)
/dev/sde1 on /media/carl/CIRCUITPY1 type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)

however, vaguely related to #1117 it does not properly support multiple devices.

The "break" changes it from defaulting to the last device found in all cases (not great) to defaulting to the first device found (also not great.)  The break is there just to have a place in the code to look at to help see why in all cases it returns the same mount point.   I'm not sure how to do this right, but this PR is at least better.

### DIFF
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -125,8 +125,11 @@ class CircuitPythonMode(MicroPythonMode):
                     mount_output = check_output(mount_command).splitlines()
                     mounted_volumes = [x.split()[2] for x in mount_output]
                     for volume in mounted_volumes:
-                        if volume.endswith(b"CIRCUITPY"):
+                        tail =  os.path.split(volume)[-1]
+                        if tail.startswith(b"CIRCUITPY") or \
+                                tail.startswith(b"PYBFLASH"):
                             device_dir = volume.decode("utf-8")
+                            break
                 except FileNotFoundError:
                     next
         elif os.name == "nt":


dybber:
Thanks. PYBFLASH sounds like it's a `pyboard`, not a CircuitPython device, but I guess it's not a problem to have it both here and in the coming `pyboard` mode.

CarlFK:
more tricky: how to call the same thing running different python builds: 

Same hardware runs both MP and CP:
https://learn.adafruit.com/adafruit-stm32f405-feather-express

I'm also not sure it matters for mu, but it makes discussions like this hard.

dybber (approved):

PR #1114, opened 2020-09-29
Increase timeout for filelistings slightly to 2 seconds (Fixes #858)

dybber:
Increasing the timeout for file listing to 2 seconds fixes issue #858, this does not address #985, which requires us to introduce a very long delay of around 15 seconds. A future option might be to make it possible to change this timeout from configuration files, but at least this fixes it for certain devices.

Issue #858, opened 2019-06-13
Problem with esp32 and mu-editor on windows 7


sbelin:
Hello,
I use esp32 (M5stack) device with mu_editor version 1.1.0. alpha.1 on macos.
It works very well.
When I use the same device with the same version under windows 7, I can not access the files.
I do not understand.
Thank you in advance,

Firmware on the M5stack : UIFlow-v1.2.3 (MicroPython ESP32_LoBo_v3.2.24 - 2018-09-06 on M5Stack with ESP32)

Here is the contents of the log file:
2019-06-13 15: 31: 54,112 - mu.logic: 1162 (sync_package_state) INFO: Synchronize package states ...
2019-06-13 15: 31: 54,112 - mu.logic: 1163 (sync_package_state) INFO: Old: set ()
2019-06-13 15: 31: 54,112 - mu.logic: 1164 (sync_package_state) INFO: New: set ()
2019-06-13 15: 31: 55,516 - mu.modes.esp: 184 (toggle_files) INFO: Toggle filesystem off.
2019-06-13 15: 31: 56,079 - mu.modes.base: 229 (find_device) INFO: Found on COM3
2019-06-13 15: 31: 56,079 - mu.modes.base: 230 (find_device) INFO: Serial number: 013EFE32
2019-06-13 15: 31: 56,086 - mu.modes.esp: 180 (toggle_files) INFO: Toggle filesystem on.
2019-06-13 15: 31: 57,155 - mu.modes.base: 403 (ls) ERROR: Could not enter raw REPL.
Traceback (most recent call last):
File "C: \ Users \ Desktop \ AppData \ Local \ Mu \ pkgs \ mu \ modes \ base.py", line 400, in ls
result = tuple (microfs.ls (self.serial))
File "C: \ Users \ Desktop \ AppData \ Local \ Mu \ pkgs \ mu \ contrib \ microfs.py", line 181, in ls
], serial)
File "C: \ Users \ Desktop \ AppData \ Local \ Mu \ pkgs \ mu \ contrib \ microfs.py", line 132, in execute
raw_on (serial)
File "C: \ Users \ Desktop \ AppData \ Local \ Mu \ pkgs \ mu \ contrib \ microfs.py", line 82, in raw_on
raise IOError ('Could not enter raw REPL.')
OSError: Could not enter raw REPL.
[...]

PR #1123, opened 2020-09-30
Serial timeout config, for code review.

WayneKeenan:
Allow the serial timeout to be overridden in `settings.json`, related to #858 and #985, then subsequently #1114 

Includes a tiny refactor related to obtaining the contents of `settings.json`.

Passes the self tests but I do not have any hardware to actually test against with me right now.

### DIFF
--- a/mu/modes/base.py
+++ b/mu/modes/base.py
@@ -668,7 +668,7 @@ class FileManager(QObject):
         """
         # Create a new serial connection.
         try:
-            self.serial = Serial(self.port, 115200, timeout=self.settings.get("serial_timeout", 1), parity="N")
+            self.serial = Serial(self.port, 115200, timeout=self.settings.get("serial_timeout", 2), parity="N")
             self.ls()
         except Exception as ex:
             logger.exception(ex)
--- a/mu/modes/base.py
+++ b/mu/modes/base.py
@@ -668,7 +668,12 @@ class FileManager(QObject):
         """
         # Create a new serial connection.
         try:
-            self.serial = Serial(self.port, 115200, timeout=self.settings.get("serial_timeout", 2), parity="N")
+            self.serial = Serial(
+                self.port,
+                115200,
+                timeout=self.settings.get("serial_timeout", 2),
+                parity="N",
+            )
             self.ls()
         except Exception as ex:
             logger.exception(ex)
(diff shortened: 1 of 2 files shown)


tjguk:
Amusingly I literally have the code for getting/settings settings data open in front of me right now on my venv branch!

I'll let others comment on the approach to setting the timeout. But I'm broadly supportive of the refactor for settings. (I have a larger refactor in mind, but this is a good start)

dybber:
Looks good, I guess this one should replace #1114.

Seems that Travis only fails due to some code style check, but the error from `black` is really not informative. Try running a `make tidy` (@carlosperate: perhaps we should use `pycodestyle` to do the check?).

WayneKeenan:
The mock test also needs updating as the timeout argument value is now unexpected, will sort.

```
E           AssertionError: Expected call: Serial('/dev/ttyUSB0', 115200, parity='N', timeout=1)
E           Actual call: Serial('/dev/ttyUSB0', 115200, parity='N', timeout=2)
```

dybber:
Great, I can approve this, thank you for the contribution! I think it's ready to merge.

dybber:
> I'll let others comment on the approach to setting the timeout. But I'm broadly supportive of the refactor for settings. (I have a larger refactor in mind, but this is a good start)

@tjguk, yes, let's do a refactor of that after this one is merged, perhaps in your branch, I also find it a bit strange that reading the settings file is done in the modes.

tjguk:
> 
> 
> > I'll let others comment on the approach to setting the timeout. But I'm broadly supportive of the refactor for settings. (I have a larger refactor in mind, but this is a good start)
> 
> @tjguk, yes, let's do a refactor of that after this one is merged, perhaps in your branch, I also find it a bit strange that reading the settings file is done in the modes.

Sure. I'll leave you to merge this one into master when you're happy, then I'll merge it into my branch and see where to go with it.

carlosperate:
I also had a look at the settings not that long ago, I believe almost everything has been moved into the session.json file at some point. So we should either unify these two files or separate some of entries from session back into settings.

Something like "serial_timeout" feel more like a settings.json entry than a session.json, so we could merge this as it is and unify/separate afterwards as needed.

PR #1129, opened 2020-10-06
Fix issue with inconsistency in device detection on startup (fixes #1121)

### LINKED ISSUE #1121: Bug: device detected, but not allowed to open REPL
An error in the new device-detection mechanism is causing the following behaviour: A device is connected, and shows up in the bottom right corner, but the dialog still shows when trying to interact with device, e.g. through opening the REPL:

Screenshot 2020-09-30 at 10 13 47

A restart of Mu fixed the problem. I will assign myself to investigate this further.

dybber:
This is a fix for issue #1121, which was caused by some signals not being sent on startup, due to `restore_session` being called before those signals where connected. The fix is just a simple reordering in `app.py`

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -176,13 +176,14 @@ def run():
     # Setup the window.
     editor_window.closeEvent = editor.quit
     editor_window.setup(editor.debug_toggle_breakpoint, editor.theme)
-    # Restore the previous session along with files passed by the os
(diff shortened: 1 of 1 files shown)


carlosperate (approved):
LGTM 👍 

I was able to replicate without the patch, and then confirm the fix with it.
Thanks Martin!

PR #1151, opened 2020-11-23
Make sure we use en_GB.utf8 as locale when executing tests. Fixes #736

### LINKED ISSUE #736: dev-setup: Makefile should set locale to `en_GB` for running tests
This relates to the developer setup on Linux (Ubuntu 18.10)
I cloned the mu-repo and installed the virtual environment in venv.
The default locale on my machine is `de_DE.utf8` *(German)*

When running the `make coverage` target I get an `Assertion Error` for every test that checks localized messages *(example at the end of the issue)*

The reason for these errors is that the `make` process on my machine runs in locale `de_DE.utf8` but the test expects it to run in locale `en_GB.utf8`. `coverage` tests (the default) english messages against the german translations and fails.

A solution would be to force `make` to use `en_GB.utf8` during execution by adding 
`export LANG=en_GB.utf8`
to the Makefile.  Here is an excerpt from `git diff`:
```
diff --git a/Makefile b/Makefile
index b0c740a..2b8b7af 100644
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 XARGS := xargs -0 $(shell test $$(uname) = Linux && echo -r)
 GREP_T_FLAG := $(shell test $$(uname) = Linux && echo -T)
 export PYFLAKES_BUILTINS=_
+export LANG=en_GB.utf8
```
As a side effect `make run` will now also run the editor in the english locale. But `run.py` can still be used to start `mu` in the default locale of the machine. *(Which is good if you want to localize the program for your language)*

**Example Error Message**

```
___________________________________________________________________________________ test_debug_mode ____________________________________________________________________________________

    def test_debug_mode():
        """
        Sanity check for setting up of the mode.
        """
        editor = mock.MagicMock()
        view = mock.MagicMock()
        dm = DebugMode(editor, view)
>       assert dm.name == 'Graphical Debugger'
E       AssertionError: assert 'Grafischer Debugger' == 'Graphical Debugger'
E         - Grafischer Debugger
E         ?    ^ - ^^^
E         + Graphical Debugger
E         ?    ^^  ^^

tests/modes/test_debug.py:17: AssertionError
```

dybber:


carlosperate:
Thanks Martin!

This is probably okay for an initial fix, but I think we should keep #736 open until we are able to update the tests to not expect a specific locale.

Although for this PR to be ready we should also update make.py.

Issue #1155, opened 2020-11-27
Dependency conflict in macOS with briefcase dependencies


carlosperate:
We currently have a dependency conflict with the current setup.py, which is the reason the Travis macOS builds are failing with the latest pip:

https://travis-ci.org/github/mu-editor/mu/jobs/745708216
```
pkg_resources.ContextualVersionConflict: (urllib3 1.23 (/Users/travis/.pyenv/versions/3.6.5/lib/python3.6/site-packages), Requirement.parse('urllib3<1.27,>=1.25.4; python_version != "3.4"'), {'botocore'})
```

Using `pipdeeptree` we can see this is due to briefcase:
```
$ pipdeptree  
Warning!!! Possibly conflicting dependencies found:
* botocore==1.19.25
 - urllib3 [required: >=1.25.4,<1.27, installed: 1.23]
```

This is a reverse tree to see what depends on urllib3
```
$ pipdeptree -p urllib3 -r
Warning!!! Possibly conflicting dependencies found:
* botocore==1.19.25
 - urllib3 [required: >=1.25.4,<1.27, installed: 1.23]
------------------------------------------------------------------------
urllib3==1.23
  - botocore==1.19.25 [requires: urllib3>=1.25.4,<1.27]
    - boto3==1.16.25 [requires: botocore>=1.19.25,<1.20.0]
      - briefcase==0.2.9 [requires: boto3>=1.4.4]
    - s3transfer==0.3.3 [requires: botocore>=1.12.36,<2.0a.0]
      - boto3==1.16.25 [requires: s3transfer>=0.3.0,<0.4.0]
        - briefcase==0.2.9 [requires: boto3>=1.4.4]
  - briefcase==0.2.9 [requires: urllib3<1.24]
  - requests==2.25.0 [requires: urllib3>=1.21.1,<1.27]
    - briefcase==0.2.9 [requires: requests<3.0]
    - cookiecutter==1.7.2 [requires: requests>=2.23.0]
      - briefcase==0.2.9 [requires: cookiecutter>=1.0]
    - requests-toolbelt==0.9.1 [requires: requests>=2.0.1,<3.0.0]
      - twine==3.2.0 [requires: requests-toolbelt>=0.8.0,!=0.9.0]
    - Sphinx==3.3.1 [requires: requests>=2.5.0]
    - twine==3.2.0 [requires: requests>=2.20]
```

[...]

PR #1156, opened 2020-11-27
Temporarily pin boto3 to v1.15.18 to fix urllib3 version conflicts

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/1155, as there is conflicts with required urllib3 versions by different dependencies.

This will likely not be needed once we update all the dependencies (specially if we use pup instead of briefcase), but in the meantime we'll need this fix (or other way to fix the issue) for the CI to pass.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,6 +65,10 @@ extras_require = {
         # Windows native packaging (see win_installer.py).
         'requests==2.23.0;platform_system == "Windows"',
         'yarg==0.1.9;platform_system == "Windows"',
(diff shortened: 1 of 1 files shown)


dybber:
I would be comfortable merging this. We also depend explicitly on docutils to avoid conflicts with briefcase.

carlosperate:
Thanks Martin, your comment also made me realise this dependency makes more sense in the `package` section, right on top of briefcase, so I've moved it there and works as well.

PR #1163, opened 2020-12-07
Clean up Pygame Zero naming in a few places, fixes #341

### LINKED ISSUE #341: Pygame [+Zero] naming consistency?
I'm not actually sure which option Daniel Pope prefers (or which the Pygame devs prefer) but we have a mishmash of:

"Pygamezero" (the mode as displayed in the bottom right of the screen)
"Pygame Zero" (the mode as displayed when you *click* that name)
and even the list of modes seems conflicted, with the header using "Pygame Zero / PyGame Mode" and then going on to say that it should be "pygame". [https://mu.readthedocs.io/en/latest/modes.html#pygame-zero-pygame-mode]

Even the pygame website seems unsure with, eg, the announcement of pygame 1.9,3 headed "pygame 1.9.3 released" while two lines further down, the download link has "Pygame 1.9.3 released"

FWIW I feel we should just find one and stick to it regardless. The common denominator seems to be "Pygame" and "Pygame Zero"

(tagged as "trivial")

dybber:
A small cleanup to fix issue #341, however, it already seems to have been fixed in most places. Now I just went through with `grep` to find the last spots.


dybber:
I guess this is pretty uncontroversial, so I will just go ahead and merge.

PR #1164, opened 2020-12-07
Implement Find Again (F3), fixes #995

### LINKED ISSUE #995: Feature Request: Add F3 function key support for Find Again.
F3 function key is available in most editors.  It would be really great if Mu could also support it.  The only other option is the Ctrl-F to pop up the search window and click OK.

NOTE: This feature should not do the replace part of the find again.

Thank you.

devdanzin:
Basic implementation for the Find Again feature suggested in #995: an `Editor.find_again() `  method that uses the  'find' parts of `find_replace()`. Testing is a copy-and-adapt deal right now. 

Refactoring the duplicated code into two helper methods (one for empty find string, one for present string) could make it more elegant. But that gets in the way if we want Find Again behavior to diverge from Find and Replace.

Currently, an empty find string will display a modal like for find and replace. We could mention `Ctrl+F` to provide a string (since the user didn't press that shortcut ). Or we could have F3 never pop up a modal, just a status bar message.

Open to and welcoming any suggestions, criticisms, alternative designs and requests.

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -205,6 +205,7 @@ def run():
     # Connect the various UI elements in the window to the editor.
     editor_window.connect_tab_rename(editor.rename_tab, "Ctrl+Shift+S")
     editor_window.connect_find_replace(editor.find_replace, "Ctrl+F")
+    editor_window.connect_find_again(editor.find_again, "F3")
     editor_window.connect_toggle_comments(editor.toggle_comments, "Ctrl+K")
     editor.connect_to_status_bar(editor_window.status_bar)
 
--- a/mu/app.py
+++ b/mu/app.py
@@ -205,7 +205,9 @@ def run():
     # Connect the various UI elements in the window to the editor.
     editor_window.connect_tab_rename(editor.rename_tab, "Ctrl+Shift+S")
     editor_window.connect_find_replace(editor.find_replace, "Ctrl+F")
-    editor_window.connect_find_again(editor.find_again, "F3")
+    # Connect find again both forward and backward ('Shift+F3')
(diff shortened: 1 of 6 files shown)


dybber (approved):
Looks good, and works as expected. I can approve this.

devdanzin:
This has a problem: F3 is already used to "Flash your code onto the micro:bit" in that mode. Sorry about the oversight. Any suggestions for resolving this conflict?

tjguk:
@devdanzin I'd be interested in seeing this PR extended to include Shift-F3 which is a common key combo for reverse-search. If you had the opportunity, feel free. If you can't get to it, I'l merge this anyway as it is within the next day or so. Thanks :)

I've already pushed the question of the conflict with m:b mode to issue #1219 as I don't think it should stop this going ahead.

devdanzin:
Adding Shift+F3 done:
* added another shortcut;
* connect it to `find_again_backward`;
* made `find_again_backward` pass `forward=False` to `highlight_text`;
* added a workaround because `findFirst(forward=False)` doesn't  work, needs explicit line and index numbers.

I'd like to revisit this tomorrow, code is rather unpolished and tests need to check more details.

It seems we currently do a Find as if the input was a regex. If this isn't on purpose (I'll check history), it's a simple fix.

One of the changes brings a comment cheatsheet for `QsciScintilla.findFirst()` from `EditorPane.find_next_match`. It seems we don't use that method anywhere. We call `findFirst` with its opaque flags from `highlight_text` and `replace_text` instead. It might make sense to refactor this to at least always document `findFirst()` positional parameters.

tjguk:
Thanks, @devdanzin -- I hope to find time to review this tomorrow

Issue #751, opened 2019-01-30
Debugger run prints to output panel a logger.warn


carlosperate:
Tested on macOS, on release 1.0.2.

When running the debugger it prints to the panel this logger entry:
https://github.com/mu-editor/mu/blob/3cb427fdfe09cf8b2ea0b75db763da424814b60d/mu/interface/themes.py#L50

<img width="1328" alt="image" src="https://user-images.githubusercontent.com/4189262/51976134-d7b43680-247b-11e9-93b4-4f1ecccfcd0f.png">

ntoll:
@carlosperate I cannot recreate this on my Mac. I'm running on OSX 10.13.6.

carlosperate:
My version of macOS is 10.14.2 (Mojave).

I was able to recreate this with the app bundle downloaded from the website and also running from master branch. Perhaps this is some kind of global setting or environmental variable? I don't think I've set up anything like that for Python, but something to look into.

carlosperate:
I also tested this on @microbit-giles macOS 10.14.2 Mojave laptop with Mu 1.0.1 from the official installer and it shows the same message in the debugger terminal.

carlosperate:
Still an issue in the current master branch:
<img width="952" alt="image" src="https://user-images.githubusercontent.com/4189262/97237394-ea89a580-17de-11eb-8c7d-ee6fc265b1a9.png">

carlosperate:
Fixed via https://github.com/mu-editor/mu/pull/1165

PR #1165, opened 2020-12-07
Remove temporary font workaround for #552

carlosperate:
The code removed here was added in PR https://github.com/mu-editor/mu/pull/567/ and it is no longer necessary, as updating the PyQt/QScintilla packages had resolved the problem a long time ago.


dybber:
I guess this also fixes https://github.com/mu-editor/mu/issues/751, but I'm not sure why that one even happens...

Issue #1140, opened 2020-11-02
TravisCI pricing changes


carlosperate:
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

Sounds like for open source projects we'll get the equivalent of 1000 Linux minutes, and one thing to keep in mind is that macOS is x5 as expensive than the Linux builds.
<img width="1062" alt="image" src="https://user-images.githubusercontent.com/4189262/97929888-1c13eb00-1d62-11eb-8498-87f7b9b138d4.png">

A build currently consumes about 50mins in total, so that gives us about 20 builds per month more or less.

That might be enough, but we can consider moving to GitHub actions, which also offers Windows, macOS and Linux builders. Travis does have Arm builds, which we can keep running as that runner on its own is unlikely to consume all the available minutes.

carlosperate:
Travis builds reduced in https://github.com/mu-editor/mu/pull/1176.

PR #1150, opened 2020-11-23
Attempt at fixing #921 through some changes to HiDPI settings

dybber:
This is an attempts at fixing the issue with HiDPI scaling on Windows 10 as mentioned in #921. I've not been able to test that this fixes the issues on Windows, as I'm on a Mac. I'm thus looking for someone who can test it.

PR #1166, opened 2020-12-07
CI: Add initial GitHub Actions workflows.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/1140.

Now that there a reduced minutes in TravisCI we will likely hit the max rate fairly quickly, so I've prioritised this since it was a relatively quick win and to ensure we still have CI running all the tests.

The test workflow runs the `make check` tests in Python 3.6 and 3.7 in the oldest and newest version available of Windows, macOS and Linux.

The build workflow creates the executables for macOS and Windows and uploads them to GitHub as a workflow artifact. This one is currently using PyNsist and Briefcase, as Travis and AppVeyor are doing, but can be easily switched in the future to pup.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build Mu
 
 on:
   push:
-    branches: '*'
+    branches: 'master'
   pull_request:
     branches: '*'
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,5 +47,5 @@ jobs:
       - name: Upload Mu installer
         uses: actions/upload-artifact@v1
         with:
-          name: mu-editor-${{ runner.os }}-${{ github.sha }}.app
+          name: mu-editor-${{ runner.os }}-${{ github.sha }}
           path: dist
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,51 @@
+name: Build Mu
+
+on:
+  push:
+    branches: '*'
+  pull_request:
+    branches: '*'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [macos-10.15, windows-2016]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    name: Build ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+      - name: Display Python info
+        run: |
+          python -c "import sys; print(sys.version)"
+          python -c "import platform, struct; print(platform.machine(), struct.calcsize('P') * 8)"
+          python -c "import sys; print(sys.executable)"
+          python -m pip --version
+          pip --version
+          pip config list
+          pip freeze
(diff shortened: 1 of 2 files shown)


carlosperate:
That was timely, right after this PR was created this was posted to hacker news: https://news.ycombinator.com/item?id=25338983

dybber:
Looks good, should we merge this?

dybber:
Would this allow us to make builds for pull requests? E.g. for a PR like https://github.com/mu-editor/mu/pull/1150 it would be helpful if an installer was created which Windows users could try out, and confirm whether it works, without them having to have a complete development setup.

dybber:
Ah, they are available at https://github.com/mu-editor/mu/actions - however, both the Mac and Windows builds have an .app extension. Shouldn't it be .exe for Windows?

.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
+name: Build Mu
+
+on:
+  push:
+    branches: '*'
+  pull_request:
+    branches: '*'
```
dybber, line 7: I'm not sure it makes sense to run on both push and pull_request. At least it seems that this execute all the actions twice for every pull request. Perhaps it should be:

```
  push:
    branches: 'master'
  pull_request:
    branches: '*'
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
+name: Run tests
+
+on:
+  push:
+    branches: '*'
+  pull_request:
+    branches: '*'
```
dybber, line 7: Same comment here, it seems all the tests are executed twice on every platform

carlosperate:
> Would this allow us to make builds for pull requests? E.g. for a PR like #1150 it would be helpful if an installer was created which Windows users could try out, and confirm whether it works, without them having to have a complete development setup.

Yeah it would work for new PRs, or if a PR rebases with master after merge.

The current build outputs are automatically stored here from Travis and AppVeyor: http://mu-builds.s3-website.eu-west-2.amazonaws.com/

> Ah, they are available at https://github.com/mu-editor/mu/actions - however, both the Mac and Windows builds have an .app extension. Shouldn't it be .exe for Windows?

Yes, good catch! I'll remove the 'app' from the uploaded artefact name

.github/workflows/build.yml
carlosperate: I think we want the test to run in all pushes to all branches, as that is useful during development before creating a PR.

But yes, we could only build the executables on master and PR 👍

dybber (approved):

dybber:
Do we want to continue running tests on AppVeyor and Travis?

carlosperate:
Not sure about AppVeyor, we can make the can when moving to pup soon.
About Travis, is still provides the Arm platform, which is quite useful, so I would like to keep that one if possible.

dybber:
Makes sense, everyone contributing also needs to adapt, perhaps we can phase them out gradually.

I just look forward to see this merged, because it will make it easier for users to test development builds.

carlosperate:
Yeah, let's merge this now and transition Travis before we run out of free credits (the macOS builds consume A LOT of credits).
The Actions are basically already active due to this PR and we can always delete this if end up deciding to go a different route.

Issue #1160, opened 2020-12-01
Support .pyw files


tjguk:
We don't currently support .pyw files when opening a file loader

devdanzin:
The change to make it work is trivial indeed, seen below. But some choices aren't as clear to an outsider like me, such as:
* should we have a single test for a .pyw file, or somehow run (a subset of?) normal .py tests for .pyw files too?
* should we hardcode .pyw along with .py (which arguably is the base file extension supported) or put it into Python3 mode (which maybe could also grow, say, .pxi someday) ?
* what about saving a new file as .pyw, is it in scope (might lead to file saved in new version not opening in old version)?


```diff

diff --git a/mu/logic.py b/mu/logic.py
index a96a36f..675e8d3 100644
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1105,7 +1105,7 @@ class Editor(QObject):
                 return
         name, text, newline, file_mode = None, None, None, None
         try:
-            if path.lower().endswith(".py"):
+            if path.lower().endswith(".py") or path.lower().endswith(".pyw"):
                 # Open the file, read the textual content and set the name as
                 # the path to the file.
                 try:
@@ -1204,7 +1204,7 @@ class Editor(QObject):
         extracts a Python script from a hex file.
         """
         # Get all supported extensions from the different modes
-        extensions = ["py"]
+        extensions = ["py", "pyw"]
         for mode_name, mode in self.modes.items():
             if mode.file_extensions:
                 extensions += mode.file_extensions
```

devdanzin:
Tim, adding .pyw as a file extension to Python3 mode is not trivial because some of what we'd need to handle it is directly implemented in logic.py instead, so the basic editor always works with Python files. Maybe adding it to places like mentioned above is good enough, but I can work on moving logic into Python mode if it's the better path.

Tidy, check and serving from web mode also do a check for `endswith(".py")`, so we should add .pyw there too.

Saving with any extension is already supported, so that's covered. Starting the REPL and importing from .pyw files works. 

Should I submit an exploratory PR from the changes I tried?

tjguk:
Thanks for giving this more consideration than I obviously have, @devdanzin  :)

I would happily look at a tentative PR moving the file-extension logic into a mode -- although I think there are several plans in the air around making modes more... well, modular / pluggable etc. So if you want to move things around, don't go *too* far. ;)

dybber:
[...]

PR #1169, opened 2020-12-09
Support .pyw files, explore a solution for #1160

devdanzin:
This PR changes a few places in logic.py that expected .py files to also accept .pyw files, allowing opening, checking, and tidying (saving already worked).

[The first commit](https://github.com/mu-editor/mu/commit/f9ce9d630aa4c2514764e7ae5b67546ce77a11a4#diff-50021fc2523f19f6dbad358e978343012cd08531c4675a8f087973a5547220ee) just duplicates the checks, but I refactored that into a helper method. 

Regardless of whether this will be the approach we take, we need a design for tests: lots of duplication? Check both extensions in cases only .py was tested? Just a couple specific .pyw tests?

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1105,7 +1105,7 @@ class Editor(QObject):
                 return
         name, text, newline, file_mode = None, None, None, None
         try:
-            if path.lower().endswith(".py"):
+            if path.lower().endswith(".py") or path.lower().endswith(".pyw"):
                 # Open the file, read the textual content and set the name as
                 # the path to the file.
                 try:
@@ -1204,7 +1204,7 @@ class Editor(QObject):
         extracts a Python script from a hex file.
         """
         # Get all supported extensions from the different modes
-        extensions = ["py"]
+        extensions = ["py", "pyw"]
         for mode_name, mode in self.modes.items():
             if mode.file_extensions:
                 extensions += mode.file_extensions
@@ -1303,7 +1303,7 @@ class Editor(QObject):
         return False.
         """
         logger.info('Checking path "{}" for shadow module.'.format(path))
-        filename = os.path.basename(path).replace(".py", "")
+        filename = os.path.basename(path).replace(".pyw", "").replace(".py", "")
         return filename in self.modes[self.mode].module_names
 
     def save(self, *args, default=None):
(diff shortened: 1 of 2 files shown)


mu/logic.py
```diff
@@ -1204,11 +1205,12 @@ def load(self, *args, default_path=None):
...
         """
         # Get all supported extensions from the different modes
-        extensions = ["py"]
+        extensions = [ext.strip("*.") for ext in self.python_extensions]
         for mode_name, mode in self.modes.items():
             if mode.file_extensions:
                 extensions += mode.file_extensions
-        extensions = set([e.lower() for e in extensions])
+        # Sort extensions to get deterministic output
+        extensions = list(sorted(set([e.lower() for e in extensions])))
```
dybber, line 1213: I guess you don't need set() here?

mu/logic.py
```diff
@@ -1859,3 +1863,13 @@ def tidy_code(self):
                 "these problems."
             )
             self._view.show_message(message, information)
+
+    def has_python_extension(self, filename):
+        """
+        Check whether the given filename matches recognized Python extensions.
+        """
+        file_ends = filename.lower().endswith
+        for ext in self.python_extensions:
```
dybber, line 1872: Explicit loop can be avoid using any()

mu/logic.py
```diff
@@ -1303,7 +1305,9 @@ def check_for_shadow_module(self, path):
         return False.
         """
         logger.info('Checking path "{}" for shadow module.'.format(path))
-        filename = os.path.basename(path).replace(".py", "")
+        filename = (
+            os.path.basename(path).replace(".pyw", "").replace(".py", "")
```
dybber, line 1309: Maybe we should actually do a case-insensitive replace here? Not sure it's worth the extra work though.

dybber (approved):
Looks good! I added a few suggestion, but nothing important. Didn't know the modes already had a file_extensions field.

mu/logic.py
devdanzin: As things stand, no, using a set() here doesn't help at all and introduces unstable ordering. It can be argued that it might help when adding modes, but IMO it'd need a test exercising extension deduplication to justify the set() staying.

Having the results sorted or ordered seems desirable, removing the set and sort gives us order. I'll try it like that.

mu/logic.py
devdanzin: Sweet!

mu/logic.py
devdanzin: Makes sense, we accept uppercase extensions so IMO we should do case-insensitive replace now rather than wait for a user to hit this corner case. Trying it.

devdanzin:
Thanks for the review, @dybber! Changes pushed. 

Should we add a couple of new tests for these changes?

dybber:
Great, I don't think we need any tests of these small changes

Issue #889, opened 2019-07-13
EP2019 BD | Webmode hard crash on macOS


tmontes:
### Context

* Observed during EP2019 Beginners' Day sessions.
* User running macOS 10.14.x with UI set to German.
* Mu 1.1.0.alpha2 (probably commit 4678e35c309849f6ec0e14422af34e00ba362e81) installed from the DMG image distributed at https://codewith.mu.

### Behaviour

* Mu launched by double clicking its icon in the Finder.
* Mu set to web mode.
* The it either crashed when (sorry, can't remember fully):
  * Starting the web server.
  * Having it respond to the first request (inclined to believe this was the actual case).

### More info

* I have collected the full log which Michael (the user) was kind enough to email me.
* Won't paste it in full here, because of its size (but keeping it, in case anyone wants to dig deeper).
* A few suspect lines show up repeatedly:

```
2019-07-09 11:04:33,196 - mu.modes.web:182(stop_server) DEBUG: Stopping Flask app.
2019-07-09 11:04:33,196 - root:96(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
File "/Applications/mu-editor.app/Contents/Resources/app/mu/modes/web.py", line 122, in run_toggle
self.stop_server()
File "/Applications/mu-editor.app/Contents/Resources/app/mu/modes/web.py", line 186, in stop_server
os.kill(pid, signal.SIGINT)
KeyboardInterrupt
2019-07-09 11:04:37,056 - root:113(run) INFO:
...
2019-07-09 11:05:16,080 - mu.modes.web:175(start_server) DEBUG: Starting Flask app.
2019-07-09 11:05:52,312 - mu.modes.web:182(stop_server) DEBUG: Stopping Flask app.
2019-07-09 11:05:52,312 - root:96(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
File "/Applications/mu-editor.app/Contents/Resources/app/mu/modes/web.py", line 122, in run_toggle
self.stop_server()
File "/Applications/mu-editor.app/Contents/Resources/app/mu/modes/web.py", line 186, in stop_server
os.kill(pid, signal.SIGINT)
KeyboardInterrupt
2019-07-09 11:05:56,421 - root:113(run) INFO:
...
2019-07-09 11:06:25,540 - mu.modes.web:175(start_server) DEBUG: Starting Flask app.
[...]

PR #1174, opened 2020-12-10
Use a fixed unicode locale (en_GB.UTF8) when starting Flask server

dybber:
Should be a fix for issue #889, but needs confirmation from someone experiencing the problem.


dybber:
Seems like Travis is failing, because it's can't download the branch. I'm not sure why that's the case. It's trying to grab it from a branch in this repository (mu-editor/mu), even though the branch is in my repos (dybber/mu).

I've tried restarting the job, but it fails in the same way. I'm not sure about the AppVeyor error.

tjguk:
I propose to merge this anyway just to move it forward. If it doesn't solve the problem (or adversely affects someone who didn't have the problem, we can revisit). In any case Flask mode isn't even released yet :)

PR #1175, opened 2020-12-10
Stretch ESP flasher logarea when admin dialog is resized

### LINKED ISSUE #1126: Settings ESP Flasher: Log text area should take all available height
If the settings window height is increased, the log text are should take all the available vertical space:

image

dybber:
Fixes #1126


carlosperate (approved):
Tested and LGTM 👍 
Thanks Martin!

PR #1177, opened 2020-12-11
CI: Separate Win32 installer build and ensure it's uploaded.

carlosperate:
Before this PR the workflow was uploading the `dist` folder, which got cleaned (and deleted the 32 bit installer) when the 64 bit installer was build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,19 +33,24 @@ jobs:
         run: |
           pip install .[dev]
           pip list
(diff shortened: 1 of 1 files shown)


carlosperate:
"failed" CI checks are AppVeyor and Travis builds I've manually stop to save credits, as this PR only affects Github Actions.

As this is a fairly uncontroversial PR I'll merge now to fix the Windows 32 bit builds not being correctly uploaded.

PR #1178, opened 2020-12-12
micro:bit flash() refactor (in preparation for V2 work)

carlosperate:
The flash method in the micro:bit mode is too long (over 250 lines) with a lot of different paths.
Before we do any V2 work it would be quite beneficial to refactor it to make it more readable, without changing the behaviour (unless we fix bugs).

~~This PR is not yet finished,~~ and will quite likely touch every single line of the method.
The best way to review it is to **go commit by commit**, which I will try to make as atomic as possible and include info in the commit message to help with the review as well.

### DIFF
--- a/mu/contrib/uflash.py
+++ b/mu/contrib/uflash.py
@@ -3,7 +3,7 @@
 This module contains functions for turning a Python script into a .hex file
 and flashing it onto a BBC micro:bit.
 
-Copyright (c) 2015-2018 Nicholas H.Tollervey and others.
+Copyright (c) 2015-2020 Nicholas H.Tollervey and others.
 
 See the LICENSE file for more information, or visit:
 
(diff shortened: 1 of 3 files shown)


carlosperate:
Related PR in uFlash: https://github.com/ntoll/uflash/pull/69

carlosperate:
To make sure I understood how all the decisions where made in the original method I created this flowchart.
It can probably help with the review as well:
code2flow_ZE9Jye

carlosperate:
There are currently conflicts with the main branch, but these are related to the settings feature, so I prefer to do a single update with the review comments and fixing the conflict, as I'd like to merge this PR with the git history since the git diff touch every single line of this feature and the commits are atomic with the information about what individual changes have been made.

carlosperate:
@ntoll have you had a chance to look at this? There are more V2 specific updates I would like to add in a new PR on top of this work.

ntoll:
Hi @carlosperate been so busy with work. So many apologies.

This looks good to me..! Thank you..!

I like how you've refactored the micro:bit mode and separated concerns from the mega function. I also (strangely) enjoyed reading your flowchart... sort of like revisiting an old school and finding your own classroom - a tad familiar in a reminiscent manner. ;-)

All tests pass and it flashed devices as expected when I gave it a spin with my laptop.

Merging now.

carlosperate:
No worries, thanks for the review!

PR #1198, opened 2020-12-18
Venv test coverage

tjguk:


tjguk:
I've now achieved all the test coverage I can with the exception of the tricky-to-test slot/signal stuff. Assuming the current GH actions pass on this push, I'll merge this soon after so I can focus on some of the more functional work elsewhere

Issue #1184, opened 2020-12-16
Refactor settings & session


tjguk:
The code to handle settings and sessions is scattered around the codebase a little, exacerbated by the need to add baseline packages for the venv PR #1072. There are also other more or less directly-related issues which might be gathered into an overall PR:

#976 -- override UI language
#693 -- start-up switches
#960 -- amnesia mode
#625 -- multiple workspace directories
#1185 -- bring baseline packages into settings

tjguk:
The main work has now been done via PR #1200 -- the other issues identified above can be addressed separately.

PR #1200, opened 2020-12-20
Refactor settings

tjguk:
Addressing #1184 

Organise access to settings / session data so it happens via global singletons and saves automatically
(Looking forward) probably need to allow override of filenames and "amnesia mode"


mu/settings.py
```diff
@@ -0,0 +1,99 @@
...
+    DEFAULTS = {}
+
+
+#
+# Create global singletons for the user & session settings
+# Load these from their default files and register exit
+# handlers so they are saved when the Mu process exits
+#
+user = _UserSettings()
+user.load(os.path.join(config.DATA_DIR, "settings.json"))
```
devdanzin, line 94: Since we can fail due to the open as well as the load, needing some guards, maybe add a topic about failure modes and strategies for missing, invalid, unreadable etc. files? Amnesia or reset modes could be triggered from these errors.

devdanzin, line 94: Since we can fail due to the open as well as the load, needing some guards, maybe add a topic about failure modes and strategies for missing, invalid, unreadable etc. files? Amnesia or reset modes could be triggered from these errors.

tjguk: Heh :) Great minds and all that. I already have some uncommitted code which does some useful guarding both around the load and the save elements. Definitely a good idea to note the direction of travel in the design docs, though.

tjguk: I've added some notes here:

https://github.com/tjguk/mu/blob/refactor-settings/docs/design/settings-and-sessions.rst#failure-modes

docs/design/settings-and-sessions.rst
```diff
@@ -0,0 +1,132 @@
...
++++++++
+
+* A failure to find/open a settings file is considered usual: it's expected
+  that, the first time around, a user settings file won't exist to be read.
+  The loader will log a warning and carry on as though it had found it empty
+* A failure to read the JSON from a settings file is more complicated. For
+  pragmatic purposes, the intention is here is: log a warning; quarantine the
+  file; and carry on as though it had been found empty. That way the editor
+  continues to work, albeit in "reset" mode, and the failing file is available
+  for debugging.
```
carlosperate, line 109: Although not something to be done by this module itself, I think we should show an error to the user.
The most likely situation of a malformed JSON file is the user trying to change a setting, and in that case loading the defaults without immediate visible warning can be confusing and/or misleading.

tjguk: @carlosperate agreed, but that's part of a wider piece: how to keep the user informed during startup. I have a notion about essentially extending the current "Splash" frame to have a status line with a slot which startup processes can signal, indicating progress or errors / failures etc.

Of course, we might have a failure even to start up a UI at all. But I regard that as a slightly different problem

tjguk:
@carlosperate  one of the functions I relocated as part of the work on settings was, in effect, your commit:

https://github.com/mu-editor/mu/commit/8a3b9ae576d47d51771bdf63d0bf90bdf1b8e149

Do you know / can you outline the need here? I mean: the code's there and it works, so at one level it doesn't matter. But neither @ntoll nor I could remember the use case behind it.

carlosperate:
> @carlosperate one of the functions I relocated as part of the work on settings was, in effect, your commit:
> 
> [8a3b9ae](https://github.com/mu-editor/mu/commit/8a3b9ae576d47d51771bdf63d0bf90bdf1b8e149)
> 
> Do you know / can you outline the need here? I mean: the code's there and it works, so at one level it doesn't matter. But neither @ntoll nor I could remember the use case behind it.

It was added when the app was packaged as a single executable. In that case this was useful to be able to add a settings.json file in the same folder as mu.exe and it would pick up the settings from that file, so users could configure some of the "hidden" features.

I would argue that a feature like this could still be useful, but not necessary. If we were to update this it should perhaps look for the settings file in the `mu_code` folder instead and maybe as a read-only file, to overwrite other defaults without adding a lot of noise to the file.

At the time this was introduced it had certainly helped people easily change settings without having to figure out where the local "app settings" directory was located, which is usually a multi-step process.

tjguk:
> It was added when the app was packaged as a single executable. In that case this was useful to be able to add a settings.json file in the same folder as mu.exe and it would pick up the settings from that file,..

Thanks, @carlosperate; that's just the kind of information I was looking for. For now, the functionality remains there unchanged, but we might revisit in any later reconsideration of where settings file are searched.

Issue #917, opened 2019-09-12
Code checker reports too many whitespace errors


martybillingsley:
The Check command reports too many irrelevant issues when checking Python code. The whitespace comments, which Python doesn't care about, make it very difficult for beginning coders to see actual syntax errors. 

For example:
x=1
is legal Python, yet the check reports 
"Missing whitespace around operator"

This issue was raised -- and closed -- in June 2018, but the latest Mu download doesn't include the fix.

Any way we can configure Mu so the check feature reports only syntax errors?

tjguk:
Per my comment on gitter, I think that the addition of the [Tidy] button (invoking Black [1]) means that we can now disable layout-only checking on the [Check] button, leaving only the syntax / semantic issues.

My experience with youngsters is that their code is sufficiently untidy that pressing [Check] means they're bewildered by an array of "you've got the wrong amount of space..." messages and miss the essential "You haven't defined this name" messages.

I'll assign to myself with a view to a PR for feedback. @martybillingsley, feel free to prepare a change yourself. If you'd like to, and need guidance on the codebase, please ask back here.

[1] https://github.com/psf/black

Issue #655, opened 2018-09-14
The "Check" tool should check for tabs (tabulators)


aoloe:
The Mu editor indents by four space and there seems to be no way to change that.

If someone copy pastes some code from an external sources, tabs can sneak in.

The "Check" Tool should detect them.

aoloe:
tabulators are probably ok, if inside of strings...

carlosperate:
Some editor automatically switch the tabulation to your predefined preference, I wonder if that is worth investigating for Mu (this might be a terrible idea, but worth mentioning just in case).

Does the check tool currently detect if there are mixed tabs?

aoloe:
imo, forcing four spaces is a good behavior and i'm not sure that nowadays there is a need for being able to set it to something else (except when working on legacy code).

the python style guide also says that:
https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces
(i promise, i've found the link after formulating my comment) 

when further thinking about it, i have no idea if there might be a reason to allow tabs anywhere in a python script...

on the other hand, yes, the check tools does detects tabs... but only if they are mixed with spaces, not if a full block is in tabs.
if you paste in a full leftmost block  (a function, a loop, ...) it will work correctly (only tabs).
but as soon as the user starts adding lines it will break, and a user will have a hard time understanding why (most beginners to intermediates mostly don't get the difference between them...).

in my eyes, if it's easy to avoid that tabs can sneak in, then that's the better option...

PropGit:
The unfortunate situation is

1. spaces unnecessarily add up to more memory usage (easier to run out of space for your code) in memory-constrained devices like the micro:bit
2. users will end up pressing Tab or Space at different times, sometimes unknowingly
3. code copied from other sources may contain a mix of leading tabs and spaces the just happen to line up visually in the previous (and maybe also mu) editor or source location (web page?) but will play havoc with the Python interpreter

The end result is the user has a frustrating time and begins to think Python is no fun or that the tools are broken.

Mu currently replaces Tab key presses with 4 space characters, but if code containing tabs is copied from another source and pasted into Mu, in does not replace those, resulting in mixes that appear correct but the demons are lurking just out-of-sight.

[...]

PR #1204, opened 2020-12-23
Ignore whitespace and blank line checks for pycodestyle

### LINKED ISSUE #977: space after #
When error checking in Mu, if there is not a space after #, Mu reports an error:
Block comment should start with a '# '

There is a lot of python code that has been written that does not have a space after a pound sign.
It would be very helpful if this requirement could be turned off.

devdanzin:
Disable checks for all [whitespace (E2) and blank line (E3) errors](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) for the Check button, to avoid spamming the users with low-reward error messages. Should fix #977 and #917, also addressing #655 as in leaving tabs to be solved by the Tidy button.

We might also want to ignore most indentation checks, there are a lot of low value ones (like indentation in a comment is not a multiple of four), some that might help improving style (like misalignment of continuation lines and their braces), and at least two that seem worth keeping (because they flag and provide context to IndentationError , which prevents Tidy from running):

* E112: expected an indented block
* E113: unexpected indentation

These helpful indentation checks often get de-emphasized by "not multiple of four" errors in the same message box.

Unpatched result for very badly mis-spaced code (also used for a new test added):
unpatched_ignore_whitespace

Patched result for the same very badly mis-spaced code:
patched_ignore_whitespace

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -469,11 +469,8 @@ def check_pycodestyle(code, config_file=False):
         "E121",
         "E123",
         "E126",
-        "E226",
-        "E203",
-        "E302",
-        "E305",
-        "E24",
+        "E2",
+        "E3",
         "E704",
         "W291",
         "W292",
@@ -515,8 +512,6 @@ def check_pycodestyle(code, config_file=False):
             line_no, col, msg = matcher.groups()
             line_no = int(line_no) - 1
             code, description = msg.split(" ", 1)
-            if code == "E303":
-                description += _(" above this line")
             if line_no not in style_feedback:
                 style_feedback[line_no] = []
             style_feedback[line_no].append(
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -620,13 +620,27 @@ def test_check_pycodestyle():
     """
     Ensure the expected result if generated from the PEP8 style validator.
     """
-    code = "import foo\n\n\n\n\n\ndef bar():\n    pass\n"  # Generate E303
+    code = "import foo; print(foo.__file__)"  # Generate E702
     result = mu.logic.check_pycodestyle(code)
     assert len(result) == 1
-    assert result[6][0]["line_no"] == 6
-    assert result[6][0]["column"] == 0
-    assert " above this line" in result[6][0]["message"]
-    assert result[6][0]["code"] == "E303"
+    assert result[0][0]["line_no"] == 0
+    assert result[0][0]["column"] == 10
+    assert " on one line (semicolon)" in result[0][0]["message"]
+    assert result[0][0]["code"] == "E702"
+
+
+def test_check_pycodestyle_no_whitespace_newlines():
+    """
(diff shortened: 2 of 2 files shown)


tjguk:
I'm very much with @devdanzin here, especially now that we have the [Tidy] button.

dybber:
To me this would disable one of the more useful features of Mu. I recommend students not to use "Tidy", as they don't learn anything from pressing it. It's like relying on Word grammar checks to teach students about grammar.

For me it would be better with separate "Check syntax" and "Check style" buttons, or make it configurable if users wants to disable style checks, or make the code style warnings as only underlines, so you have to mouse over to see what the problem is.

devdanzin:
Sorry @dybber , I should have raised an issue to discuss before sending a PR (and read previous discussion more carefully). Adapting this into two buttons should be trivial, maybe named something like "basic check" and "through check". The other options don't seem too complicated either, but we need some sort of consensus for a design.

We may be able to serve equally well complete beginners that get confused by a wall of style errors, and beginners that are able to produce mostly well-formatted code and benefit a lot from the style checks. For both groups it might make more sense to first check for more serious issues, fix them, then check for simple style issues in separate steps.

Since I've jumped the gun on this, I'd like to close this PR for now and create an issue to explore our options, would that work? Or should we continue here?

dybber:
It would be nice with a discussion, but perhaps I also overstated my opinion in my immediate reaction. We should find the right balance and adding yet another button should be last resort.

I can also see your point that it can become quite a big "wall of warnings/errors", that will also be hard to get an overview of for a beginner.

I like to extend on one of your suggestions. What if:
 - First the button will check for syntax errors and report those
 - If the button is pressed and no syntax errors is found it will do a PEP8 check (except white space/indentation issues)
 - If the button is pressed and neither syntax nor reduced PEP8 check is done, it will do a full PEP8 check

I guess then it will useful with some kind of UI indication of how far it got. With emojis:

🟧✔️
🟨✔️
🟩✔️

carlosperate:
> I guess then it will useful with some kind of UI indication of how far it got. With emojis:
> 
> 🟧✔️
> 🟨✔️
> 🟩✔️

Probably better to use SVGs in case emojis on Qt are also OS-dependent. This is what I see 😆 
image

Just a quick note to add that I would also prefer a setting or a multi-purpose button (if it's easy to understand), as I'd rather  avoid an extra button. Even with the new resizing feature, we have quite a few buttons already.

carlosperate:
This issue also has relevant discussion about button UI: https://github.com/mu-editor/mu/issues/762

dybber:
Sorry about the emoji-confusion, I was replying using the iOS app, but somekind of UI indication of how far the checker got.

I think it would be good to discuss this at one of the online meetings.

ntoll:
Agreed about discussion at next online meeting.

We're getting into "vi vs EMACS" territory here when it comes to code checking... ;-) However we resolve it, we should be able to get the end result into Mu for future releases in 1.1.

PR #1205, opened 2020-12-23
Update check_flake regex for handling builtins and add "const" builtin, fixes #788

### LINKED ISSUE #788: Add builtins to address Undefined Names
Would it be possible to allow adding terms to the linter? For example, Mu flags the MicroPython _const()_ method as 'undefined name'.

In Atom or VS Code, I use Flake8 which allows me to create a config file.  Adding builtins = const to the config file prevents warnings regarding _const()_.

devdanzin:
Fix #788 by adding a "const" builtin to MicroPythonMode. For the fix to work, solve a issue similar to  #1161, as our regexes were missing flake's new output formatting and not matching builtins. Moved the regex to a module constant so it's easier to spot it when fixing other regexes.

Updated existing test and added test using real check_flake output to catch any regressions.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -52,6 +52,8 @@ LOG_FILE = os.path.join(LOG_DIR, "mu.log")
 STYLE_REGEX = re.compile(r".*:(\d+):(\d+):\s+(.*)")
 # Regex to match flake8 output.
 FLAKE_REGEX = re.compile(r".*:(\d+):(\d+)\s+(.*)")
+# Regex to match undefined name errors for given builtins
+BUILTINS_REGEX = r"^undefined name '({})'"
 # Regex to match false positive flake errors if microbit.* is expanded.
 EXPAND_FALSE_POSITIVE = re.compile(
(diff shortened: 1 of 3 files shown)


dybber:
Great. Looks good to me. 

Both AppVeyor and Travis are failing but it seems unrelated to your modifications, instead it seems to be related to virtual environments.

tjguk:
@devdanzin  could you rebase this -- or otherwise rework -- against current master, please? (Which includes the fix in #1207)

devdanzin:
Sure, done.

It turns out the regex problem here was fixed by #1207 (I think), so the only real benefit we now get from this PR is adding the `const` builtin to `MicroPythonMode` and a test. We only had a "new" format because the previously applied `FLAKE_REGEX` failed to match (due to a real new format). I kept the moving of `BUILTINS_REGEX` into the code that has other regex constants, but can remove it as well.

dybber:
Great. I can also approve of these changes. I will merge.

PR #1207, opened 2020-12-25
Fix FLAKE_REGEX to match new flake 'path:lineno:col message' format.

devdanzin:
Mu currently misplaces "undefined name" errors from flake because the message format changed, like in #1161 and #1205.
before_flake_regex
after_flake_regex


tjguk:
Looks like we still have an issue with columns but let's merge this to get the lines right...

PR #1210, opened 2020-12-29
CI: Add GH Action workflow to run tests on Arm.

carlosperate:
Moving away from Travis, the main thing left over there are the tests running in the Arm platform as done in https://github.com/mu-editor/mu/pull/1104.

This GH Actions workflow job uses docker with a QEMU backend to emulate armv7 and run a Debian Buster container.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,3 +45,41 @@ jobs:
       - name: Run tests
         if: runner.os != 'Linux'
         run: python make.py check
+
+  build-arm:
+    runs-on: ubuntu-latest
+    name: Test Py 3.7 - arm-debian-buster
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: 'linux/arm64,linux/arm/v7,linux/arm/v6'
+      - name: Check Debian image info
+        uses: docker://multiarch/debian-debootstrap:armhf-buster
+        with:
+          args: /bin/bash -c "uname -a && cat /etc/os-release"
+      - name: Install dependencies and run tests
+        uses: docker://multiarch/debian-debootstrap:armhf-buster
+        with:
+          args: >
+            bash -c "
+              apt-get update &&
+              apt-get install -y python3 python3-pip python3-virtualenv &&
+              apt-get install -y python3-pyqt5 python3-pyqt5.qsci python3-pyqt5.qtserialport python3-pyqt5.qtsvg python3-pyqt5.qtchart &&
+              apt-get install -y libxmlsec1-dev libxml2 libxml2-dev libxkbcommon-x11-0 libatlas-base-dev xvfb &&
+              python3 -m virtualenv venv --python=python3 --system-site-packages &&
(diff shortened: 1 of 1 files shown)


carlosperate:
@tjguk there are some virtualenv related test failures in this Arm CI.


All the errors are the same:
```
Subprocess.CalledProcessError: Command '['/github/workspace/venv/bin/python', '-m', 'virtualenv', '-q', '/github/home/.local/share/mu/mu_venv']' returned non-zero exit status 3.` 
```
- https://github.com/mu-editor/mu/pull/1210/checks?check_run_id=1619580180#step:6:2735
- https://github.com/mu-editor/mu/pull/1210/checks?check_run_id=1619580180#step:6:2860
- https://github.com/mu-editor/mu/pull/1210/checks?check_run_id=1619580180#step:6:2971
- https://github.com/mu-editor/mu/pull/1210/checks?check_run_id=1619580180#step:6:3083

https://github.com/mu-editor/mu/blob/42418443e034378693e77111373198c13380f1bc/mu/virtual_environment.py#L389-L393

Executing this command in the docker image provides:
   ```
   $ /venv/bin/python -m virtualenv -q /root/.local/share/mu/mu_venv 
  The path python2 (from --python=python2) does not exist
  ```

It looks like virtualenv (installed via apt from package `python3-virtualenv`) might be defaulting to the "python2" executable if not provided.  If this is not the default behaviour it might be a debian patch.

Maybe we need to pass sys.executable as the python flag?

carlosperate:
I can confirm that the tests pass if I uninstall the `python3-virtualenv` (v15.1.0) package and installing `virtualenv` (v20.2.2) via pip.

If I install the same version (v15.1.0) with pip that also works.

carlosperate:
Yep, debian patch: https://sources.debian.org/patches/python-virtualenv/1.11.6+ds-1/


> python2 default.patch
>  even though /usr/bin/virtualenv is a python 3 script now, in  Debian we still want to default to using Python 2 unless the -p/--python  argument is given.

carlosperate:
Now that the python flag is provided there are other virtualenv issues.
It looks like the `pvenv.cfg` is not created.

- https://github.com/mu-editor/mu/runs/1619718148?check_suite_focus=true#step:6:2693
- https://github.com/mu-editor/mu/runs/1619718148?check_suite_focus=true#step:6:2740

Having a look at the issue tracker it seems virtualenv is not meant to create `pvenv.cfg`?
https://github.com/pypa/virtualenv/issues/719

tjguk:
Thanks for investigating @carlosperate. Your `-p sys.executable` fix looks good to me. Re detecting a virtual env --- it's a bit of a pain, but I'm happy to try for some sort of workaround. I'm still not quite clear why this fails on Debian, since it works on every other platform we're trying on. But... distros will be distros, I imagine.

tjguk:
I've raised a separate PR to address these together. I'll watch the Travis results. Assuming all is good, I hope my PR should be mergeable into yours with any conflicts. (Or, since the changes are fairly trivial, you could just cherrypick)

carlosperate:
Great, thanks @tjguk! I've rebased and all the Arm tests (Pythonnn 3.7) in GH Actions are passing 👍

carlosperate:
Okay, all the GH workflows have passed, so I'll merge this.
Getting close to be able to fully drop TravisCI.

PR #1211, opened 2020-12-29
Address virtualenv issues on Debian

tjguk:
Address issues raised by @carlosperate in https://github.com/mu-editor/mu/pull/1210#issuecomment-751905774 and following:

* Use the `-p` switch to specify the current `sys.executable` when creating the virtualenv
* Use the presence of a python executable to detect whether a directory is a virtual env even if a pyvenv.cfg doesn't exist

### DIFF
--- a/mu/virtual_environment.py
+++ b/mu/virtual_environment.py
@@ -334,6 +334,27 @@ class VirtualEnvironment(object):
         else:
             return self.process.run_blocking(self.interpreter, args)
 
+    def _directory_is_venv(self):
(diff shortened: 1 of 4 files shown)


tjguk:
Ok; I'm about to merge this. I'm getting a clean bill of health *except* for the issue of trying to install black on Python < 3.6. I'll need to give some thought to the best way to manage this and I don't want to delay the slew of other changes in this PR  which should at least reduce the issues of building on Debian and/or Python 3.5

PR #1229, opened 2021-01-08
Bump Python version up to include 3.8

tjguk:
Now that pygame has wheels built for Python 3.8 on MacOS bump up our dev and testing range to include 3.8


ntoll:
Yup. :slightly_smiling_face: :+1:

PR #1233, opened 2021-01-10
Add support for Spresense board

baggio63446333:
Add VID and PID of Spresense board to CircuitPython mode.


carlosperate (approved):
Thanks @baggio63446333! LGTM 👍 

Can confirm the Sonny VID/PID from the CircuitPython config: https://github.com/adafruit/circuitpython/blob/main/ports/cxd56/boards/spresense/mpconfigboard.mk

Issue #970, opened 2019-12-08
True dark mode


bythckr:
This is a suggestion. First off, love the project and a nice tool to learn python. Much better than IDLE. Hope its become standard in python soon.

In the dark mode/theme,
- The terminal at the bottom when you run/debug/repl/plotter a code doesn't turn dark.
- The Mu administration opened by the cogwheel, the input background of the "Current Log" & "Python3 Environment" is white.
- The button background (between the yellow ring & icon) is white, [issue is already created](https://github.com/mu-editor/mu/issues/847). Instead of an image with a background, wouldn't an image with transparent background be better?
- Also if the buttons on the top could be smaller, it would be great. An option to make it small, if there is a need for big buttons for some.

dybber:
Just checked what the status is now.

- [ ] The Python 3 REPL still has a white background in dark-mode
- [x] It seems the background in the administration Window is now fixed
- [ ] Buttons not following the theme setting is still an issue I guess, but I'm not a dark-mode user. I can see the linked issue #847 is now closed without a solution implemented. Perhaps chime in over there what you think, to make sure it's recorded that multiple people are requesting this.
- [ ] Smaller buttons might be linked to this suggestion, where the Zoom buttons also affects the buttons https://github.com/mu-editor/mu/pull/955 and the general discussion of making buttons have the same size no matter what resolution computer they are displayed.

devdanzin:
Making the Python 3 REPL follow the theme seems to be a simple enough change. It seems we could even use the current theme CSS files if we want matching colors.

For the REPL, High contrast is trivial, it's white on black and only the prompt color can be argued. Night theme can be turned into a fully-colorful dark theme (using monokai below), or kept much closer to its current black-on-white styling (but dark). 

Here's a quick demo from the minimal changes below:

https://user-images.githubusercontent.com/74280297/104253619-b9228c80-5453-11eb-9dce-6b208fe44571.mp4

The Night->Day colors aren't working well, but from High Contrast and Day flipping black/white, it seems to be simply a matter of redefining those colors in the right places.

```diff
diff --git a/mu/interface/panes.py b/mu/interface/panes.py
index c1deb6e..e7f31dc 100644
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -56,6 +56,8 @@ from PyQt5.QtGui import (
     QStandardItem,
 )
 from qtconsole.rich_jupyter_widget import RichJupyterWidget
+from qtconsole import styles
[...]

PR #1238, opened 2021-01-12
WIP: Dark REPL and buttons to match Night and High Contrast themes

devdanzin:
This is a simple implementation of making the REPL and buttons match the themes, as mentioned in #970, offered as a PR to make it easier to test locally.

It uses Monokai as the Night theme syntax coloring and the "transparent icons" trick, but only for dark themes. These icons aren't good enough for actually using in Mu, just for the proof of concept. More discussion is needed regarding whether this is desirable and, if it is, what design choices we'd like to make. 

day
night
highcontrast


dybber:
Hi @devdanzin, I decided to split this in two parts: the REPL issues with dark-mode and the buttons. We've now merged a fix for the REPL-part of this, i did the same as you with monokai for the dark-mode syntax style, but instead of making a new HighConstrast style sheet, I just applied the existing stylesheets to the QtConsole widget

We still need to find a way forward with the buttons. I don't think just making the background transparent is good enough, but we should instead recolor all the buttons. I don't know if we have vector versions or some other format with individual layers (GIMP, .psd, Krita or similar), but that would be the easiest way for us to recolor them rather than work with the PNG-images we have now.

ntoll:
SVGs of button images can be found here: https://github.com/mu-editor/Mu-tate/tree/master/mu/icons (not all images are the same as what we currently use - but they're close enough). Linking here for visibility.

ZanderBrown:
I have thought for some time we should use SVG icons, easier to tweak and even without HiDPI we use them at multiple sizes

dybber:
> SVGs of button images can be found here: https://github.com/mu-editor/Mu-tate/tree/master/mu/icons (not all images are the same as what we currently use - but they're close enough). Linking here for visibility.

Great, I'm not much of a graphical designer, but at least there's something to work with here, if we can find someone who could do it. I can try and ask around, unless one of you would like to give it ago.

Rybec:
I've created a whole set of tool bar buttons in SVG format for this.  Find them in Pull Request #1603.

devdanzin:
Thanks for the ping, @carlosperate. I think it's simpler to just close this one and start a new one, since both the code and the resources (and the requirements) have changed a bit. Good news is we can recolor the SVGs using just a text editor, so the bikeshedding can happen efficiently.

Maybe it's better as two PRs, one for the changing of icons when themes change, one for the final pick of colors. I'll try to do it soonish.

carlosperate:
That would be great, thank you so much @devdanzin!

Issue #1239, opened 2021-01-13
Python3(Jupyter) REPL "Kernel died, restarting"


litodemesa:
If you are reporting a bug, we would like to know:

* What you were trying to do,
* What steps you took to make this happen,
* What you expected to happen,
* What actually happened,
* Why this difference is problematic (it may not be a bug!),
* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 

Please remember to attach a **copy of the full log files for Mu**. You can get
the logs by clicking on the cog icon in the bottom right of the editor window.
Click on the logs and use CTRL-A to select all, then CTRL-C to copy and CTRL-V
to paste the contents into the issue.

Thank you for contributing to Mu! :-)

litodemesa:
Just try to run REPL
[log_file.txt](https://github.com/mu-editor/mu/files/5810126/log_file.txt)

devdanzin:
Thank you again! Can't test this right now, but from the log I see Mu is working with a `python3.6.zip` and the kernel is trying to start from a `python3.7` executable. I believe Mu shouldn't be invoking any external Python install.

devdanzin:
OK, I can reproduce this by passing a bad PYTHONPATH in the Python3 Environment tab (the one close to the log view). If your case is as simple as mine, just deleting a PYTHONPATH line from that tab will solve it.

Still not sure how exactly it gets recorded there, it somehow happens on its own when I repeatedly alternate between running one of two Mu versions (1.1.0-alpha.2 and current git) that were installed to different places, using different Python versions. Probably related to running the script or opening the REPL.

devdanzin:
It seems simply making Mu unable to receive PYTHONPATH from Editor.envars (both for REPL and Run actions) fixes the issue I can reproduce. 

We can also check whether the Python used to start the kernel is the same as `venv.interpreter` and use the latter if they differ (only requires a simple subclassing of `QtKernelManager`). Or just always use it.

Very similar to #944 (except in that case there was no Python3.7 installed) and #584 (version mismatch, but on MacOS).

dybber:
I will reopen this, at the fix is still not merged

dybber:
Fix merged. Closing

PR #1240, opened 2021-01-14
WIP: Fix REPL using wrong interpreter, bad PYTHONPATH from envars breaking things

devdanzin:
This records two tentative fixes for #1239:
* Avoid using any `PYTHONPATH` values from envars ("Python3 Environment" tab), as that is prone to breaking stuff and way more likely if user has more than one Mu version installed.
* Check whether `QtKernelManager` figures out the right interpreter to run the kernel on. If it doesn't, we log the issue and tell it to use `venv.interpreter` instead. Since I couldn't actually reproduce this issue, not sure the value added goes beyond better diagnostics (i.e., kernel might still fail to run).

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -862,6 +862,7 @@ class PythonProcessPane(QTextEdit):
         """
         if not envars:  # Envars must be a list if not passed a value.
             envars = []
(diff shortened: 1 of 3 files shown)


ntoll:
Looks good to me. You're on fire @devdanzin :fire: :+1:

PR #1241, opened 2021-01-18
Add a Lego Spike mode.

ntoll:
This branch adds the (mythical?) Lego Spike mode. :-)

As you'll see it's simply a Lego version of the ESP mode, so the actual changes to code are minimal. Things to note:

* I've made a slight update to the ESP mode so that it uses the `board_name` attribute when displaying device names to the user.
* The Lego mode simply overrides names and other Lego specifics.
* The `LEGO_API` stuff used for auto-complete is *old*. I've reached out to my contacts with Lego to try to get something more up-to-date.


carlosperate (approved):
Been playing with this mode this afternoon and seems to work as expected, code looks good as well 👍 
I guess we still need to wait for the latest API.

Apart from that a couple of questions:
- Is there any new MicroPython build for the SPIKE? Could not find anything online
- Is the current "run" button action the preferred mode to execute code? It sends and executes the code via serial, but there is no permanence, so users would have to save their script as a `main.py` file and save it using the files panel. Assuming most of these devices will be moving around I would expect users to prefer a flash-like button like the micro:bit mode?

dybber:
> Is there any new MicroPython build for the SPIKE? Could not find anything online

Probably completely closed source

> Is the current "run" button action the preferred mode to execute code? It sends and executes the code via serial, but there is no permanence, so users would have to save their script as a main.py file and save it using the files panel. Assuming most of these devices will be moving around I would expect users to prefer a flash-like button like the micro:bit mode?

That was also the original implementation in the ESP-mode. Why we went the other way is explained here (and the following comments in the thread): https://github.com/mu-editor/mu/pull/676#issuecomment-465596177

carlosperate:
> That was also the original implementation in the ESP-mode. Why we went the other way is explained here (and the following comments in the thread): [#676 (comment)](https://github.com/mu-editor/mu/pull/676#issuecomment-465596177)

Yeah, I partially agree with the reasoning for the ESP mode, but for Lego Spike most user's will be playing with code that is meant to run untethered and the process to transfer code via the files panel will get tedious fast.

Apart from if the code runs when the users click the run button, but not when it is unplugged, it doesn't feel very discoverable to have to create a main.py file to transfer via "files". It makes sense once you know why that is the process, but it's the kind of thing new users can struggle with.

dybber:
I argued the same way for the ESP-mode, so I guess @ntoll should chip in on how he sees it.

ntoll:
I've just merged simply to get this into the pending release for two reasons:

* Folks (Tufts University) want to run classes and test this.
* We're releasing not-final, so we can refine when the docs become available and given feedback from Tufts.

After my initial concerns about the "Run" button... I actually quite like it now and prefer it over saving a file. I believe @dybber was right in making it explicit to save a file to the "file system" on boards since this is changing the state of the device in a way that's reminiscent of updating a server by deploying a web-app. It's a thing that should be done on purpose.

Furthermore, Lego are in a state of flux and their "native" API for interacting with the device (via a JSON RPC like mechanism) is being retired/revised (so I'm led to believe), so were we to use it, we'd only be making work for ourselves that we'd have to re-write/revise later. Ergo, "Run" and copying via "Files" is definitely the way to go since they completely bypass the native API.

I hope this makes sense.

As always, many many thanks for the work and contributions involved in this stuff...! :+1:

dybber:
I also don't mind the Run-button as it is now. I can see why Carlos suggests it, but I think it will work out okay. At least we can get it tested in this version, and then perhaps in the future change it if too many people complain that it's too cumbersome a workflow

PR #1244, opened 2021-01-22
Remember debugger layout and expanded items (fixes #979)

### LINKED ISSUE #979: Debugging windows don't retain formatting changes
Hi! New to Mu! New to programming. Just started using the "Debugger" in Python 3 mode. When the debugger stops at a breakpoint, the "Names" and "Values" pane opens up to the right. So far so good. But, I like to resize the "Names" and "Values" column (so there's more space between the two), and expand any expandable items (like a dictionary, for instance), so that I can see everything quickly and clearly at a glance. However, when I click on the "Step Over" button to execute the next line of code, my reformatting goes away. The columns are now scrunched up again, and any expanded items get collapsed. So I end up reformatting that window each time I step through my code.
Is there a way to retain my formatting choices while stepping through my code?
Windows 10, btw, if that makes a difference.

devdanzin:
Proposed fix for #979 by:
1. remembering column widths on each step of the debugger;
2. saving column widths on closing debugger and using them on reopening;
3. recording a set of variables the user has expanded and keeping them that way.

Needs tests for each of these features and the docs/comments wording is subpar.


ntoll:
Great stuff..! Thank you again (again) @devdanzin :-)

PR #1245, opened 2021-01-22
Added FeatherS2 and TinyS2 to CircuitPython supported boards

UnexpectedMaker:
Just adding some of my CircuitPython boards to the list of supported boards


mu/modes/circuitpython.py
```diff
@@ -64,6 +64,8 @@ class CircuitPythonMode(MicroPythonMode):
         (0x04D8, 0xEAD2, None, "DynOSSAT-EDU-OBC"),
         (0x1209, 0x4DDD, None, "ODT CP Sapling M0"),
         (0x1209, 0x4DDE, None, "ODT CP Sapling M0 w/ SPI Flash"),
+        (0x239A, 0x80AC, None, "Unexpected Maker FeatherS2"),
```
carlosperate, line 67: If this is the Adafruit PID? In that case the board will be automatically detected without this addition.

mu/modes/circuitpython.py
```diff
@@ -64,6 +64,8 @@ class CircuitPythonMode(MicroPythonMode):
...
+        (0x303A, 0x8002, None, "Unexpected Maker TinyS2"),
```
carlosperate, line 68: Was this allocated via this repo? https://github.com/espressif/usb-pids

carlosperate (commented):
Thanks for the PR @UnexpectedMaker!
Added a couple of questions.

mu/modes/circuitpython.py
UnexpectedMaker: Yes, TinyS2 was issued from Espressif as part of their new allocating PIDs to their customers

ntoll:
@UnexpectedMaker thanks for the contribution. As @carlosperate pointed out, if you're just using an Adafruit VID "it should just work" :tm: In any case, happy to support new boards. Many congratulations..! :tada:

PR #1234, opened 2021-01-10
Using pup to package under macOS

tmontes:
NOTES:

* As agreed last Monday, here is the first take at this.
* As of now, macOS packaged Mu is not able to run `turtle`/`tkinter` things, see:
  * https://github.com/mu-editor/pup/issues/122
  * https://github.com/indygreg/python-build-standalone/issues/68
* Not being 100% familiar with the recently merged venv branch:
  * I experienced a few odd behaviors resulting from running "in dev mode" vs "packaged".
  * Did not investigate in full but they went away once I removed any previously existing Mu venv.
  * Currently packaged macOS Mu **does not bundle** venv needed wheels -- shouldn't it?

IMPORTANT:

* I disabled the `check` dependency for the `macos` target:
  * Two tests were failing on my system.
  * I really wanted to see how this goes on CI. :-)

PR #1247, opened 2021-01-26
Readding `check` dependency to `macos` make target.

tmontes:
...should have gone into #1234. :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ win64: check
 	@echo "\nBuilding 64bit Windows installer."
 	python win_installer.py 64 setup.py
 
-macos:
+macos: check
 	@echo "\nPackaging Mu into a macOS native application."
 	python -m venv venv-pup
 	# Don't activate venv-pup because:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,10 +29,9 @@ jobs:
           pip --version
           pip config list
           pip freeze
-      - name: Install Mu dependencies
-        if: runner.os == 'Windows'
+      - name: Install Mu test dependencies
         run: |
-          pip install .[dev]
+          pip install .[tests]
           pip list
       - run: mkdir upload
       - name: Build Windows 32 bit
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,9 +29,9 @@ jobs:
           pip --version
           pip config list
           pip freeze
-      - name: Install Mu test dependencies
+      - name: Install Mu test and package dependencies
         run: |
-          pip install .[tests]
+          pip install .[tests,package]
           pip list
       - run: mkdir upload
       - name: Build Windows 32 bit


dybber:
Tests are failing on MacOS-11.0, Python 3.7, but it seems they already did that on master branch. Github Actions doesn't show any log, so it's hard for me to understand why it's failing. Is this due to the lack of notarization necessary for Big Sur?

Makefile
```diff
@@ -115,7 +115,7 @@ win64: check
 	@echo "\nBuilding 64bit Windows installer."
 	python win_installer.py 64 setup.py
 
-macos:
+macos: check
```
carlosperate, line 118: Actually I wouldn't mind if "check" was left out the build commands, it reduces the CI time. But this could be a breaking change for developers that expect this to be part of the process.

@ntoll @tjguk @dybber @devdanzin @ZanderBrown thoughts?

devdanzin: IIUC, check would still run in tests, so the effect for code that passes check would be faster builds and same amount of information. For code that doesn't pass check, it would mean the build doesn't abort earlier. So I think removing check from build is OK as the usual case is passing it. If we'd like to avoid wasting build minutes on rare check fail cases, maybe we could delay build until check passes in tests.

carlosperate: Yeah, I think removing it for the CI builds is the right call, since otherwise we are running the tests twice.

My main concern if people that run builds locally would prefer to keep the tests as part of the build command, so that it continues to aborts the build if the local tests fail.

tmontes: I thought about that, too... I'm trying to keep things as much as they were before, though, and forgot to keep the `check` dependency on the `macos` target.

Future `pup` should be able to run tests **within the packaged application** but that's not the case yet... Thus, inline with @carlosperate thoughts WRT local builds, I figured I'd "play it safe" and keep the dependency.

tmontes:
> Tests are failing on MacOS-11.0, Python 3.7, but it seems they already did that on master branch. Github Actions doesn't show any log, so it's hard for me to understand why it's failing. Is this due to the lack of notarization necessary for Big Sur?

No, tests run against the installed source -- no depdendency on signing or notarization. Those same tests passed in the last commit in PR #1234, which I'm kind of amending with this one! (see https://github.com/mu-editor/mu/actions/runs/508253731)

More: not really sure why the **Run testst / Test Py 3.7 - macos-11 (pull_request)** check is listed 3x -- could it have been a CI glitch?

carlosperate:
Yeah, macOS 11 looks a bit buggy in GH, it's quite possible the error is about running the instance.
I'd restart the tests until we can see an output (and it will likely be a pass).

tmontes:
> Yeah, macOS 11 looks a bit buggy in GH, it's quite possible the error is about running the instance.
> I'd restart the tests until we can see an output (and it will likely be a pass).

...any hint on how to do that from within GH without pushing a "blank" commit? :-)

carlosperate:
I take it you don't have this button?
image

tmontes:
> I take it you don't have this button?
> image

...hmmm, felling silly! Maybe I missed it. Will look for it! Thanks. :-)

Makefile
tmontes: FYI: Logged #1248 to track / promote this discussion.

tmontes:
Merging decision (same as PR #1234):

* `pup` is packaging Mu on macOS -- run `make macos` at the repo root on macOS.
* Travis and AppVeyor failures were there before.
* macOS 11 on Python 3.6 failed now, when it failed before for 3.7 -- declaring macOS 11 CI as brittle.

PR #1249, opened 2021-01-27
Appveyor: Fix current issue by removing the built exe rename

carlosperate:
It just uploads with the default name.
We'll be removing this soon, but in the meantime we can get another green tick on the CI.


carlosperate:
AppVeyor tests have passed, I'll merge.

PR #1250, opened 2021-01-28
Using pup to package under Windows

tmontes:
This is the Windows counterpart of #1234 and, much like then, for now, the `check` dependency is commented out.

The `build.yml` build spec packages both 32 and 64-bit MSIs but, IIUC, both files end up with the same name. Needs review.


tmontes:
First attempt was nearly there:

* The conflicting file names are leading to a Windows packaging failure on CI.
* A few leftovers made all other checks fail. :-)

Next run should be better, but fixing the conflicting filenames is still to be done... (needs thinking?)

tmontes:
Second attempt still failing a lot due to `flake8` strictness WRT to the long URL added to `make.py`.

Let's go again!

tmontes:
Outstanding failures:

* Duplicate artifact filenames.
* AppVeyor - `pup` does not support 3.6:
  * Update to running on 3.7?
  * Do not package, just test?
  * Drop AppVeyor given that we have Windows 3.6 under GitHub actions?

To be addressed some other day (soon...). :-)

carlosperate:
Let's just drop AppVeyor, we don't need to continue using it as we have GH Actions on Windows already.

tmontes:
Good progress:

* Windows 32/64 bit name clash addressed.

Next:

* Drop AppVeyor.

tmontes:
Note:

* Expected that removing the `.appveyor.yml` would disable the AppVeyor check.
* Removed the webhook from the repo settings.

tmontes:
This looks merge-able:

* All green...
* ...except dear old Travis.

Will continue tomorrow.

tmontes:
Merge decision:

* All checks green...
* ...except Travis with Python 3.5 on ARM, IIUC -- completely unrelated, was already failing under #1234.

Going forward!

Issue #1122, opened 2020-09-30
Mu should remember UI pane positions and sizes


tmontes:
(I could have sworn this topic has been raised in a past issue, but could not find it, so here it goes)

## Facts

* Depending on how it is used, one or more UI panes are displayed within Mu's window: REPL, debugger, charts, and maybe more.
* Qt (and thus Mu) allow users to resize these panes and also re-attach them to differente window sides (example: REPL is shown attached to the bottom but can be dragged to the right in order for it to become more like a "column", instead of a "row").
* In general, such user-driven UI adjustments are not remembered by Mu (some, across restarts, others, across pane close/open cycles).

## Reproducing | Attachment

* Lanch Mu in Python 3 mode.
* Bring up a Python 3 REPL by clicking the REPL button.
* The REPL pane is displayed, attached to the bottom of the window.
* Drag that pane (from its "title bar") towards the right-hand side of the window until it attaches there.
* The result is now a UI with two large columns: editor on the left, REPL on the right.
* Close the REPL and bring it up again by clicking twice on the REPL button.
* The REPL pane is shown attached to the bottom of the window.

### Expected behaviour
* The REPL pane should have been displayed attached to right-hand side and with the same width as when it was closed.

Notes:
* Suspect such re-attachment mis-behaviours are observed for all of Mu's UI panes.

## Reproducing | Sizing

* Launch Mu in Python 3 mode.
* Create a new file with, say, a few `print` statements.
* Initiate a Python 3 debugging session by clicking the DEBUG button.
* Two panes are displayed:
  * The running pane, attached to the bottom.
  * The debug inspector pane attached to the right.
* Adjust the width of the debug inspector pane.
* Adjust the width of the *name*/*value* columns within the debug inspector pane.
* Stop debugging.
* Start debugging again.
* The debug inspector pane is still the same width... (Good!)
* ...the *name*/*value* column widths are some how reset.

### Expected behaviour
* The  *name*/*value* column widths shouln't have been changed from what we have just set.

### Continuation

* Quit Mu.
* Relaunch Mu.
* Initiate the same debugging session as before.
* Two panes are displayed, again, but the width of the debug inspector pane was forgotten.

Expected behaviour:
* The width of the debug inspector pane should have been remembered, along with the width of the  *name*/*value* column within it.

Notes:
* Suspect such sizing mis-behaviours are observed for all of Mu's UI panes.

## Epiloge

[...]

PR #1251, opened 2021-01-28
Remember UI pane positions for session

devdanzin:
Add simple support for reopening the UI panes (inspector, plotter, repl, runner) in the same dock area (bottom, left, right) they were previously moved to, as discussed in #1122.

As can be seen below, a given pane's position isn't stable in each of these areas. As cannot be seen below (sorry!) both the debugger inspector and debugger runner remember the areas they belong to, and moving a debugger runner doesn't change normal runner's recorded area.

As of right now REPLs, plotters and non-debugger runners clobber the positions of equivalent panes in other modes.

The filesystem pane isn't currently movable, but can easily be added if we decide to change that.

remember_panes2

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -319,6 +319,12 @@ class Window(QMainWindow):
 
     def __init__(self, parent=None):
         super().__init__(parent)
+        # Record pane area to allow reopening where user put it in a session
+        self._fs_area = 0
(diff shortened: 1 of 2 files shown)


ntoll:
Looks good to me. Merging. Many thanks @devdanzin :tada:

PR #1252, opened 2021-01-29
Display startup information on a splash screen

devdanzin:
This PR adds way to follow what's going on during Mu's initialisation. The current approach is to forward log entries into a new splash screen that is only used for showing messages (aka an infoscreen) and position it below the pixmap splash screen.

I'll be opening a corresponding issue so we can easily decide to go with another approach/design.

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -48,13 +48,16 @@ from .interface.themes import NIGHT_STYLE, DAY_STYLE, CONTRAST_STYLE
 from . import settings
 
 
-class SplashScreenHandler(logging.Handler):
+class ForwardingHandler(logging.Handler):
+    """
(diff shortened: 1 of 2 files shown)


ntoll:
Superseded by #1260 (but the logging code from this is a great start to enhance the work over there..!).

I feel I must apologise for not running with this PR in the first place. I actually started to do so, but realised (after discussions with @tjguk) that given the time-constraints for the release, we needed something simple yet animated that worked while the potentially long-running background tasks were in flight.

Like I said, the logging of the output of the `pip` command in the set-up task could be a great textual addition to such a screen.

devdanzin:
No worries, feels good to explore options, thanks for adding an indicator to the splash that we'll ship.

PR #1257, opened 2021-01-31
Support for micro:bit V2

carlosperate:
With the updated uFlash that uses a Universal Hex:
- https://github.com/ntoll/uflash/pull/70
- https://github.com/carlosperate/uflash/pull/1/files

The main thing that changes is that when a custom runtime is used:
- It will always flash it and then send the files via serial
- If a local location has been selected by the user an error is shown
- If Mu cannot find a serial port an error is shown
- Before, the last fallback could try to attach a script to a custom hex and flash that, in this case that will not happen as uFlash cannot do the attachment

I'll update the test later on Sunday.


carlosperate:
If anybody could test this with their micro:bits in the meantime that will be really useful as well :).

carlosperate:
Okay, I've updated the tests.
I still need to add more, but if we are cutting a beta soon let's try to get this in and I can continue adding tests later.

ntoll:
Hi @carlosperate -- this is great to see. I've tested it on a micro:bit v1 and v2 and all seems to be good.

PR #1258, opened 2021-01-31
Ensure a default exists for the workspace directory

tjguk:


tjguk:
(Just about to fix up some failing tests)

PR #1260, opened 2021-01-31
Add an animated splash screen that displays until the blocking setup tasks are done.

ntoll:
This PR is "inspired by" #1252 by @devdanzin. My initial intention was to simply take #1252 and run with it but realised, after chatting with @tjguk that an animated general purpose splash we could iterate on would be simpler and easier in the long run. Ergo this branch was born.

The GIF animation used is:

splash_screen

... It'll do for now, but I'll ask a designer friend (Steve H.) to take a look and do something that's less inclined to make you cross-eyed. Further improvements could be to add textual updates somewhere based on the output of the log files (see #1252 for a wonderful example of how this could be done).

How does it work..?

* An `AnimatedSplash` sub-class of QSplashScreen is used to manage the animation.
* A `StartupWorker` class is used to call the blocking long running tasks Mu needs to do before the editor can start properly.
* In the `run` function, once the splash screen is displayed, a new `QEventLoop` is created (since the one for the application doesn't exist yet), and a `StartupWorker` instance is run on a thread that will stop the event loop once the startup worker is finished.
* Obviously, this loop blocks from when we call its `exec` method, until the startup worker is finished.
* Then the splash screen is closed and the app continues as before.

I've also added unit tests for the new classes and functions created to fulfil this work. The GIMP file used to create the animation has also been added.


ntoll:
Rebased from master. Doh.

tjguk (approved):
tl;dr it works ok, both with first-time run with no existing venv; and later.

However... I'm seeing console-logged messages "QObject: Cannot create children for a parent that is in a different thread". It would be nice to get to the bottom of these, but I don't think it should stop us going ahead with this change

ntoll:
Yes... I see these same error messages, but had exactly the same thought as you. We can revise / refine / revisit once we've released. :tada:

PR #1262, opened 2021-01-31
Find using plain text instead of regex

### LINKED ISSUE #1222: Find uses regex mode by default
It seems we search using regex syntax and I can't find anything to suggest it's on purpose. Unclear we'd want to change it now as some users might be depending on this behaviour. To reproduce:

1. Open Mu, paste the code below;
```python
for line in open('test.txt', 'r'):
    print(line)
```
2. Hit `Ctrl+F` and search for `t.` (tee dot);
3. If the bug is real, it should highlight `te`, `t.`, `t'` and `t(` each time you hit `Crtl+F` then `Enter`.

Sending a trivial PR as soon as we agree this should change. Or a different PR with just a test to ensure we don't change it by accident.

devdanzin:
Fixes #1222 by calling `EditorPane.findFirst` with the regex flag set to `False`.

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -1174,7 +1174,7 @@ class Window(QMainWindow):
         if global_replace:
             counter = 0
             found = self.current_tab.findFirst(
-                target_text, True, True, False, False, line=0, index=0
+                target_text, False, True, False, False, line=0, index=0
             )
             if found:
                 counter += 1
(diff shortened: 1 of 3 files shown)


ntoll:
LGTM :tada: :shipit:

ntoll:
OK... tests all pass for me and my manual testing shows it does what it's supposed to do. Beginners can encounter the joys of regex via not-Mu at some future date. ;-) Nice save @devdanzin

Issue #1267, opened 2021-02-01
Reduce the Python compatible versions to up to 3.7


carlosperate:
As we currently cannot install from only wheels in macOS Python 3.8 due to pgzero locking a version of pygame not available for this platform.

- pygamezero locks pygame to be < 2.0.0: https://github.com/lordmauve/pgzero/blob/1.2.post5/setup.py#L11
- The latest version of pygame that matches that is v1.9.6 and doesn't Python 3.8 wheels: https://pypi.org/project/pygame/1.9.6/#files

So if a user tries to pip install mu-editor in 3.8 it will try to build from source, which will likely fail in most system (this also applies to the packaging).

PR #1268, opened 2021-02-01
Reduce Python version compatibility up to 3.7.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/1267.

In essence we need to remove Python 3.8 at the moment due to the pygame wheel not being available for macOS (pinned by pgzero).
We can restore this once we can update pygamezero or pygame to a version with the wheels we need.


carlosperate:
Current issue seems to be to do with GH Action setting up the job, so i'll have to restart the entire pipeline (odd that they don't have an option to restart only the failed jobs...)
image

ntoll:
Thanks..! :-)

PR #1271, opened 2021-02-02
Make: Fetch wheels as first packaging step.

carlosperate:
@tmontes @tjguk Is this how you expected the inclusion of this step in the CI builds?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,8 @@ win64: check
 	python make.py win64
 
 macos: check
+	@echo "\nFetching wheels."
+	python -m mu.wheels
 	@echo "\nPackaging Mu into a macOS native application."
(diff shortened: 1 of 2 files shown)


tjguk (approved):
Yes; that's the kind of thing I was thinking of...

carlosperate:
The CI failures are from the macOS environment having issues starting up.

Wheels from the CI packages are present in the macOS build (cannot uncompress the msi into something that resembles the app files):

image

The zip files have not really increased in size compared with previous builds, I think it's because the wheels are downloaded during test runs, so they have been included all along.

This PR does make it more explicit, so it still seems worth including it.

carlosperate:
Cool, I'll close this one then 👍

PR #1273, opened 2021-02-03
WIP - Allow the baseline packages to include git installations

tjguk:
Not working yet -- tests currently failing when creating virtual env

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,12 +34,6 @@ jobs:
           pip install .[tests]
           pip list
       - run: mkdir upload
-      - name: Build Windows 32 bit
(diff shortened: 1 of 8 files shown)


carlosperate:
Just saw this. Change looks good, makes things more clear.
I think all the tests (except the Arm Debian, which should be fixed after a rebase) are likely to pass.

tjguk:
The weird thing is that I can't get all the tests to pass locally! I'll try resetting the environment.

carlosperate:
Btw, the error currently experienced in the Arm Debian tests (also failing in maser now) is due to issues in DockerHub:
https://status.docker.com/pages/533c6539221ae15e3f000031

PR #1275, opened 2021-02-03
CI: Add Python 3.5 checks to GitHub Actions.

carlosperate:
As I'm updating make.py @tjguk would you be able to review this PR?

### DIFF
--- a/make.py
+++ b/make.py
@@ -172,7 +172,7 @@ def black():
     print("\nblack")
     # Black is no available in Python 3.5, in that case let the tests continue
     try:
-        _ = subprocess.run([TIDY, "--version"])
+        subprocess.run([TIDY, "--version"])
     except FileNotFoundError as e:
         python_version = sys.version_info
         if python_version.major == 3 and python_version.minor == 5:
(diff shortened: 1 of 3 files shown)


make.py
```diff
@@ -170,6 +170,17 @@ def tidy():
 def black():
     """Check code with the 'black' formatter."""
     print("\nblack")
+    # Black is no available in Python 3.5, in that case let the tests continue
+    try:
+        _ = subprocess.run([TIDY, "--version"])
```
tjguk, line 175: No need to assign the result of `subprocess.run` if it's not being used. (Just do `subprocess.run(..)`

tjguk (approved):
Basically the approach looks fine. I've left one tiny comment for something which seemed unnecessary. (But which obviously works fine)

make.py
carlosperate: Yeah, it's a bit cleaner that way, originally I was looking at the output, but ultimately we only care if it throws an exception.

carlosperate:
The CI failures are unrelated to this PR, so I'll merge this.

make.py
dybber: assigning to _ is also a bit ugly, when we use _ for localization :)

PR #1278, opened 2021-02-06
Hide admin tabs not relevant for the current mode. Fixes #1125

### LINKED ISSUE #1125: Only show administration tabs relevant for the current mode
To simplify the administration panel, I suggest that only tabs relevant for the current mode are shown. We currently do this for the ESP Firmware flasher, but might also be relevant to do for:

 - micro:bit tab only shown in micro:bit mode
 - Python3 Environment + Third Party Packages, only in Python 3 and Web mode

At least that's my suggestions, but I'm not sure if that would introduce any problems to hide them? In the future when more and more modes are added it could be come unwieldy.

dybber:
Suggested fix for #1125. This hides the administration tabs not related to the current mode, making it harder for users to things like trying to use the "Third Party Packages" tab to install packages on microbit/other MicroPython devices (I've seen that one a lot, even by professional developers teaching at our code club...)

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -480,17 +480,18 @@ class AdminDialog(QDialog):
         self.tabs.addTab(self.log_widget, _("Current Log"))
         self.envar_widget = EnvironmentVariablesWidget(self)
         self.envar_widget.setup(settings.get("envars", ""))
+        self.microbit_widget = MicrobitSettingsWidget(self)
+        self.microbit_widget.setup(
+            settings.get("minify", False),
+            settings.get("microbit_runtime", ""),
+        )
+        self.package_widget = PackagesWidget(self)
+        self.package_widget.setup(packages)
         if mode.short_name in ["python", "web", "pygamezero"]:
             self.tabs.addTab(self.envar_widget, _("Python3 Environment"))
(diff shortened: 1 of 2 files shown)


tjguk:
I'm broadly in favour of the idea. In the longer term I'd like this to become part of any mode-refactoring we care to do, including things like mode-specific downloads etc. But I don't object at all to a short-term implementation

carlosperate:
I must admit I'm not too sold on this change.
I do agree that having too many tabs makes the settings window harder to navigate, but I think only showing the settings for the current mode makes them less discoverable.
That being said, a change like this is a lot simpler (and that's a good thing) than making something like a "modes" tab with  a second level tab system (either horizontal or vertical), so this would be an improvement.

----- 

Gave the installer from this PR a test and saw this in the Python mode settings:
image

The same on Windows looked fine.

ntoll:
Yes... this is another corner of Mu that started as a simple flower bed dug quickly for growing daffodils and has now grown into something more like Kew Gardens. Settings should definitely be better managed and I think this is a pragmatic and sensible first step. I have the same concerns as @carlosperate when it comes to discoverability, but these can be addressed in a later ticket where we've thought out what the most appropriate solution should be.

I'll merge this in the meantime.

Issue #1266, opened 2021-02-01
Tests flakiness and some test seems to be locking up in CI


carlosperate:
The GitHub Actions jobs have been queuing since yesterday.
The grey icons are test runs I had to cancel as they were stuck there for 4h+:
- https://github.com/mu-editor/mu/runs/1802206901?check_suite_focus=true
- https://github.com/mu-editor/mu/runs/1802207019?check_suite_focus=true
- https://github.com/mu-editor/mu/runs/1802207054?check_suite_focus=true
<img width="945" alt="image" src="https://user-images.githubusercontent.com/4189262/106436805-70fff400-646c-11eb-9349-c589d70e74f4.png">

Other CI runs with a similar lock up:
- https://github.com/mu-editor/mu/runs/1802309139?check_suite_focus=true
- https://github.com/mu-editor/mu/runs/1802424340?check_suite_focus=true
- https://github.com/mu-editor/mu/runs/1802424319?check_suite_focus=true
- https://github.com/mu-editor/mu/runs/1802285428?check_suite_focus=true

## Update

There was originally 2 issues, one to do with dependencies in Python 3.8 (already fixed) and another problem not  yet identified where some runs get stuck: https://github.com/mu-editor/mu/issues/1266#issuecomment-772083787

carlosperate:
Apart from that it looks like we are seeing test flakiness when the CI runs the tests twice for the same commit (one for the tests and one for the executable builds) and the tests fail in one of the runs:

- https://github.com/mu-editor/mu/actions/runs/525833317
<img width="946" alt="image" src="https://user-images.githubusercontent.com/4189262/106437623-5c702b80-646d-11eb-9d39-0b2138243917.png">

- https://github.com/mu-editor/mu/runs/1802161458?check_suite_focus=true
<img width="928" alt="image" src="https://user-images.githubusercontent.com/4189262/106437930-b2dd6a00-646d-11eb-8446-f49769dfcfe7.png">

- https://github.com/mu-editor/mu/runs/1801797634?check_suite_focus=true
<img width="937" alt="image" src="https://user-images.githubusercontent.com/4189262/106438070-e28c7200-646d-11eb-97d9-2d21ada2bf32.png">

However, due to the issues in the first comment we haven't run the full CI tests in the last few commits, so it is possible this could be fixed in the latest master. We'll have to look into it.

carlosperate:
[...]

PR #1288, opened 2021-02-10
make: Use virtualenv instead of venv.

### LINKED ISSUE #1286: pup-venv: Using venv instead of virtualenv sometimes does not include pip
We had the same issue with the mu virtualenv feature.

```
Packaging Mu into a macOS native application.
python -m venv venv-pup
# Don't activate venv-pup because:
# 1. Not really needed.
# 2. Previously active venv would be "gone" on venv-pup deactivation.
./venv-pup/bin/pip install pup
make: ./venv-pup/bin/pip: No such file or directory
make: *** [macos] Error 1
```

The venv does indeed not contain the pip exec:
image

Might be related to the creating this venv from a different venv. If I manually create a venv within my mu-venv I have the same  issue. But if I deactivate and try again it does add pip.

carlosperate:
A small PR to fix #1286, @tmontes could you have a look once you get a chance?
Low priority, as it is working fine in CI and other environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,4 @@ target/
 .venv*/
 /mu/wheels/*.whl
 /baseline_packages.json
+venv-pup/
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ macos: check
 	@echo "\nFetching wheels."
 	python -m mu.wheels
 	@echo "\nPackaging Mu into a macOS native application."
-	python -m venv venv-pup
+	python -m virtualenv venv-pup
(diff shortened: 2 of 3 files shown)


tmontes:
> A small PR to fix #1286, @tmontes could you have a look once you get a chance?

Looks good to me! :)

carlosperate:
Thanks Tiago!
The build CI passes and the only failure is Travis and https://github.com/mu-editor/mu/issues/1266#issuecomment-772083787, so I'll merge.

Issue #1287, opened 2021-02-10
Black is not bundled in the macOS app


carlosperate:
So when clicking the tidy button we get an error:

```
2021-02-10 21:12:46,049 - root:106(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
  File "/Users/microbit-carlos/Downloads/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/logic.py", line 1777, in tidy_code
    from black import format_str, FileMode, PY36_VERSIONS
ModuleNotFoundError: No module named 'black'
2021-02-10 21:12:46,053 - mu.settings:169(save) DEBUG: Saving to /Users/microbit-carlos/Library/Application Support/mu/session.json
2021-02-10 21:12:46,053 - mu.settings:169(save) DEBUG: Saving to /Users/microbit-carlos/Library/Application Support/mu/venv.json
2021-02-10 21:12:59,872 - root:172(run) INFO: 
```

Looking at the app bundle black is not present.

The  guards we have check the Python version instead of the package presence:
https://github.com/mu-editor/mu/blob/7282e7d11faca3e64b91af54efb43a469cae6e1e/mu/logic.py#L1772-L1773

PR #1289, opened 2021-02-10
Move black from a modes dependency to a Mu dependency.

carlosperate:
@tjguk as we need black for the tidy button I assume this is the best way to solve #1287?


tjguk (approved):

tjguk:
Thanks, @carlosperate  -- at this point I don't remember why I thought it would be a mode dependency. Doesn't seem to make any sense.

carlosperate:
Thanks Tim!
The only CI failure is Travis and testing the macOS CI installer I can confirm the Tidy button now works, so I'll merge.

PR #1296, opened 2021-02-12
Clear wheels before download

tjguk:


carlosperate (approved):
Haven't tested it locally, but LGTM 👍

Issue #1299, opened 2021-02-15
Settings tab in Python mode show other settings view


carlosperate:
Originally reported in https://github.com/mu-editor/mu/pull/1278#issuecomment-775832810.
Just tried this from master and seeing the same running from source code:
<img width="836" alt="image" src="https://user-images.githubusercontent.com/4189262/107937138-70428400-6f7b-11eb-9214-f57c27d9cd05.png">

It's hard to tell, but that's basically the micro:bit settings at the top left of the Python 3 settings tab bar:
<img width="261" alt="image" src="https://user-images.githubusercontent.com/4189262/107937242-949e6080-6f7b-11eb-98b6-7ec2eb3d28e6.png">

I can actually toggle the tick box from that tab area and it will apply the setting.

PR #1300, opened 2021-02-15
Fix for #1299 with microbit minify toggle

dybber:
This fixes issue #1299


ntoll:
@dybber could you run `make tidy` and push again. It looks like that is the cause of the failing tests.

dybber:
> @dybber could you run `make tidy` and push again. It looks like that is the cause of the failing tests.

Done. Previously "make check" also checked for PEP8 issues like these, but now it seems I only get these warnings after I make the pull request. I will have to remember to run both `make check` and `make tidy` in the future before pushing

carlosperate:
> Done. Previously "make check" also checked for PEP8 issues like these, but now it seems I only get these warnings after I make the pull request. I will have to remember to run both make check and make tidy in the future before pushing

The Makefile should only skip these tests if you are running in Python 3.5, as black is >=3.6. Is that the version you are running? Otherwise let's open a new issue.

dybber:
> > Done. Previously "make check" also checked for PEP8 issues like these, but now it seems I only get these warnings after I make the pull request. I will have to remember to run both make check and make tidy in the future before pushing
> 
> The Makefile should only skip these tests if you are running in Python 3.5, as black is >=3.6. Is that the version you are running? Otherwise let's open a new issue.

Oh sorry, maybe I used `make coverage` not `make check`, my bad. I will immediately add a pre-commit hook to automatically run `make check` in the future.

carlosperate:
No worries, the makefiles were updated recently to ensure `make check` doesn't fail in 3.5 (by skipping black) and was wondering if maybe there was still an issue.

PR #1301, opened 2021-02-15
Add esptool as dependency to simplify user experience in ESP-mode (fixes #1279)

### LINKED ISSUE #1279: ESP Firmware flasher: add button to install esptool dependency
Currently the esptool is not included with Mu, but is a dependency if users wants to flash firmware onto ESP-devices.

Thus users are required to install the package through the "Third Party Packages" tab:
Screenshot 2021-02-06 at 20 50 10

It would be nice if we instead had a button for "Install esptool", especially if #1278 is merged and the Third Party Packages tab is hidden when in ESP-mode.

dybber:
I hope it is enough to just add the 'esptool' dependency in `setup.py`, please check that I'm not missing something

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -241,18 +241,6 @@ class ESPFirmwareFlasherWidget(QWidget):
         widget_layout = QVBoxLayout()
         self.setLayout(widget_layout)
 
-        # Check whether esptool is installed, show error if not
-        if not self.esptool_is_installed():
-            error_msg = _(
-                "The ESP Firmware flasher requires the esptool' "
-                "package to be installed.\n"
-                "Select \"Third Party Packages\", add 'esptool' "
-                "and click 'OK'"
-            )
-            error_label = QLabel(error_msg)
-            widget_layout.addWidget(error_label)
-            return
-
         # Instructions
         grp_instructions = QGroupBox(
             _("How to flash MicroPython to your device")
--- a/mu/wheels/__init__.py
+++ b/mu/wheels/__init__.py
@@ -19,6 +19,7 @@ mode_packages = [
     ("pyserial", "pyserial==3.4"),
     ("qtconsole", "qtconsole==4.7.4"),
     ("nudatus", "nudatus>=0.0.3"),
+    ("esptool", "esptool"),
 ]
 WHEELS_DIRPATH = os.path.dirname(__file__)
 
--- a/mu/wheels/__init__.py
+++ b/mu/wheels/__init__.py
@@ -19,7 +19,7 @@ mode_packages = [
     ("pyserial", "pyserial==3.4"),
     ("qtconsole", "qtconsole==4.7.4"),
     ("nudatus", "nudatus>=0.0.3"),
-    ("esptool", "esptool"),
+    ("esptool", "esptool==3.*"),
 ]
 WHEELS_DIRPATH = os.path.dirname(__file__)
 
--- a/setup.py
+++ b/setup.py
@@ -54,10 +54,6 @@ install_requires = [
     # Needed for packaging
     #
     "wheel",
-    #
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -54,6 +54,10 @@
     # Needed for packaging
     #
     "wheel",
+    #
+    # Needed for ESP-mode
+    #
+    "esptool",
```
carlosperate, line 60: I think this would need to go in `mu.wheels.__init__` I thinnk

dybber: Ok

dybber: Done.

mu/wheels/__init__.py
```diff
@@ -19,6 +19,7 @@
     ("pyserial", "pyserial==3.4"),
     ("qtconsole", "qtconsole==4.7.4"),
     ("nudatus", "nudatus>=0.0.3"),
+    ("esptool", "esptool"),
```
carlosperate, line 22: Looks like esptool follows semver? v3.0 was released in November, I assume that version has been tested and works fine?

While I don't think we need to look a specific version we can probably set it to only install version 3.x, in case future (or previous) versions introduce breaking changes.

mu/interface/dialogs.py
```diff
@@ -241,18 +241,6 @@ def setup(self, mode, device_list):
...
-            error_msg = _(
-                "The ESP Firmware flasher requires the esptool' "
-                "package to be installed.\n"
-                "Select \"Third Party Packages\", add 'esptool' "
-                "and click 'OK'"
-            )
-            error_label = QLabel(error_msg)
-            widget_layout.addWidget(error_label)
-            return
-
```
carlosperate, line 255: Having thought about this as part of the discussion in the other issues I think it's probably a good idea to keep this as well. 
Users will now be able to remove the esptool package (accidentally or on purpose), so this provides a great fail mode if it's missing.
In fact, maybe we need to add similar fallbacks for the other modes as well.

dybber: Would that really be possible? Wouldn't the packages be part of the baseline packages that can't be removed? Otherwise users could also remove PyQt I guess...

carlosperate: PyQt is part of the baseline packages defined in setup.py that are used to run Mu and can't be removed, but mode packages defined in mu.wheels get installed in the virtualenv and will appear in the settings, so they could be removed.

We probably need to revisit how we expose/explain all these changes to the user, as I think it will be relatively easy to accidentally delete something like pygamezero and then suddenly the mode is no longer in the modes menu.

tjguk: In principle, anything installed as part of the baseline install (including deps of the top-level packages) shouldn't appear in the list of 3rd-party packages. It *is* possible (as I think @dybber pointed out elsewhere) to *add* them via the packaging window and then remove them -- at which point they will be gone from the venv.

dybber: Yeah, I don't think these mode packages show up in the "Third party packages"-tab unless you type them in yourself. I hope that's the behavior we're going for. I think all packages installed as part of clean installation of Mu, should never be possible to remove. If a potential user-installed dependency requires a different version of one of these package, I actually think we should just throw and error and say that the requested package is incompatible. But I guess that's another discussion...

dybber: Anyway, I can add it back in, it's just a couple of lines, and it will not do any harm having them there...

carlosperate: Sorry, you are right, I thought all the venv packages would appear in the settings window, but I guess we don't show anything stored in the venv.json file.

In that case, yeah, we can leave this out, if we ever need to bring it back in we can just have a look at the git history 👍

dybber: Okay, ready to merge I guess!

PR #1303, opened 2021-02-16
Add initial draft for friendly crash reporting.

ntoll:
This is a draft PR so folks can see how we may do the friendly crash reporting via a browser.

**Feedback most welcome.**

Here's what I've done:

* Created a self contained HTML template.
* Created translatable strings for the HTML content.
* Added, within a `try/except` in the `excepthook` code to assemble the HTML, drop it into a temp file, open up the default browser with the new HTML and exit.

The resulting web page:
* Explains there's a problem,
* Displays the immediate error message,
* Has a button to copy the error message,
* Has a link to the log file (for further investigation),
* A link to open up a new ticket on GitHub with a friendly template.

:warning: I've stuck in a division by zero in there too... so you get to see how it works when you run Mu... :boom: 

Here's what it looks like:

mu_crash_report

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -24,10 +24,10 @@ from logging.handlers import TimedRotatingFileHandler
 import os
 import platform
 import sys
+import urllib
 import webbrowser
-import tempfile
-import pathlib
 import traceback
+import base64
 
 from PyQt5.QtCore import (
     Qt,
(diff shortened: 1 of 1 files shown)


ntoll:
Tests will fail because, umm... Mu is supposed to fail. ;-) I'll remove the `1/0` when we're happy with the process and web page.

dybber:
Looks good! It would be great if the issue were created using the same template as our other issues, so users will also answer those questions when reporting an issue ("What were you doing?" etc.)

mu/app.py
```diff
@@ -104,6 +176,34 @@ def excepthook(*exc_args):
...
+            logfile="file://" + LOG_FILE,
+            error=error,
+            report=CRASH_REPORT,
+            thanks=CRASH_THANKS,
+        )
+        with tempfile.NamedTemporaryFile(
+            suffix=".html", delete=False
+        ) as crash_report:
+            crash_report.write(content.encode("utf-8"))
+        webbrowser.open(crash_report.name)
```
carlosperate, line 204: I haven't had a chance to look properly to the rest, but as a quick note, in the past I've had issues opening temp files with the browser where in some cases the tmp file was erased soon after handler was closed, so by the time the browser was opened (which can be a really long time forr some users) it couldn't find it any more.

I can't remember exactly which constructor/factory/function from tempfile I was using at the time, I think the (ugly) workaround I implemented was to manually manage the file deletion on a long timer.
Perhaps not closing the handler until the browser is open might be good enough, but we have no idea how long that might be.

ntoll: @carlosperate I've thought about this already.... note the `delete=False` flag when `NamedTemporaryFile` is instantiated... that ensures the file isn't deleted (until, I believe the OS clears out temp files when the user shuts down).

ntoll:
OK... the approach of using a `file://` is broken because Chromium based browsers don't allow it. :-/

But I have a plan. ;-)

mu/app.py
carlosperate: Oh yeah, I think I remember looking into that as well, and since the deletion is left to the OS, the files ended up hanging around in my dev machine forever (doing a search found tmp files created by my programme months and months in the past).
I guess we could consider that an OS issue and something outside Mu's scope, but that's the reason I ended up manually deleting the files in my example.

ntoll:
This approach won't work since Chromium based browsers won't open HTML directly from the user's filesystem. When using a Chrome based browser as your default browser, this is what you see:

wat_chrome

Such a file **very much does exist** and it's possible to open the file with Firefox.

More thought is required.

carlosperate:
I was fairly sure this was enabled in Chrome. I'll give this PR a try to see if I can replicate.

@ntoll are you using Chrome or Chromium from deb repositories? I woudn't be surprised if the debian package has more security flags enabled compared with the default Chrome installer.

PR #1307, opened 2021-02-19
Reduce baudrate when flashing firmware to make it more stable

dybber:
Currently when flashing firmware with our built-in ESP firmware flasher, we use a  medium/high baudrate (460800), which makes it possible to go wrong in some cases. We don't provide the user with the option to try a lower baudrate, in the case it doesn't work.

I'm not sure if we should add a dropdown widget for selecting baudrate, so for now I just think we should lower the baudrate to 115200, where it's most likely to work, and then users will have to wait for the slower process to complete, without the ability to speed it up.


dybber:
This is a small change, I think I will just merge it immediately.

Issue #238, opened 2017-01-22
Zooming should change the width of the line number column


carlosperate:
When the zoom is changed, the column on the left of the editor displays the line numbers stays the same width (the font size in that column does change as expected). Not much a problem when zooming out, but when the font size is large enough some of the characters are obscured due to the limited space:

![image](https://cloud.githubusercontent.com/assets/4189262/22185386/841a29f6-e0dc-11e6-96ed-0ec7f50c2874.png)

Not quite sure how much control we've got over this one, but we should check if this width can be changed accordingly.

ntoll:
We don't have control except for the fixed width of the column. I guess saying "don't do that" is the best we can do. :-/

carlosperate:
Can't the width of the column not be changed after it has been set? If so the zoom level could change this value as well?

carlosperate:
Can't we change the margin values while it is running by using [setMarginsBackgroundColor](http://pyqt.sourceforge.net/Docs/QScintilla2/classQsciScintilla.html#a419ab8aed49ea1711ce4ffcf19146df1) and [setMarginWidth](http://pyqt.sourceforge.net/Docs/QScintilla2/classQsciScintilla.html#a1a8b798b6b36db2b5f60b0cdccba88a1)?

ZanderBrown:
From my experience no

carlosperate:
Was there any specific issue with it? A simple hack on the code seems to change the GUI as expected (I just picked a different predefined theme item to the margin background):

![mu-hack](https://user-images.githubusercontent.com/4189262/40933200-b32aabce-6828-11e8-941d-cd6a891a2e8b.gif)

ntoll:
@carlosperate good stuff (on resizing the width). Please create a PR. ;-)

ntoll:
Care to link to your branch with the spike code / proof-of-concept..?

carlosperate:
Pushed the ugly hack for reference: 7df7af9123c763b88924d0153a27c0a9bf6370fd

ntoll:
This appears to have been fixed!

carlosperate:
Just tested the latest CI build from master and it looks more or less the same to me. There is a bit more margin, but still not dynamic width: 

<img width="1456" alt="image" src="https://user-images.githubusercontent.com/4189262/50173231-bbef3980-02ee-11e9-86b7-d6cf627511d0.png">

ntoll:
Doh... sorry... my misunderstanding... will re-open.

[...]

PR #1310, opened 2021-02-19
Make line number column width dynamic and follow zoom-levels

dybber:
Fix for #238

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -283,7 +283,17 @@ class EditorPane(QsciScintilla):
             "xxl": 16,
             "xxxl": 48,
         }
+        margins = {
+            "xs": 30,
+            "s": 35,
+            "m": 45,
+            "l": 50,
+            "xl": 60,
+            "xxl": 75,
+            "xxxl": 85,
+        }
         self.zoomTo(sizes[size])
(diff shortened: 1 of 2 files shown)


ntoll:
I'll merge this. But the debugger (breakpoint, current line) related images in the column don't also change size. I remember this is controlled by QScintilla and I'm not even sure if it's possible.

dybber:
> I'll merge this. But the debugger (breakpoint, current line) related images in the column don't also change size. I remember this is controlled by QScintilla and I'm not even sure if it's possible.

Do you want me to investigate that part as well? I think QScintilla is pretty customizable, so it should be doable

ntoll:
Please do.... I agree that QScintilla is very customizable... except the documentation can be VERY confusing. My intuition isn't so much "could we do this..?" as "can we work out the simplest incantation in code to make this work?" ;-)

It may just be a couple of lines of Python or the mother-of-all hacks. If the latter, let's discuss.... if the former just do it..! :-)

dybber:
So it was pretty easy, I just had to change the margins around the debugger-marker as well, and then the image resized as it had less space to fill. At least it works on Mac. See the new PR #1311

PR #1311, opened 2021-02-19
Make margins around debugger-marker follow zoom-levels

### LINKED ISSUE #238: Zooming should change the width of the line number column

dybber:
Follow up on #1310 to fix #238

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -283,6 +283,7 @@ class EditorPane(QsciScintilla):
             "xxl": 16,
             "xxxl": 48,
         }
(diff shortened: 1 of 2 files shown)


ntoll:
Looks good to me..! :-)

Issue #328, opened 2018-02-11
why not using system colors?


reithi:
I am a visually impaired user and I am using different color settings then the default user. I need always  to code in night mode (dark background and light text) because the other ways round I see nearly nothing.
Meanwhile many Smartphone Apps implement that feature because its much easier to see even for normal users. 
I wonder why the application isn't using my qtsettings which all the other programs are using.
Would be very helpful if it would.
Thx a lot!!

ZanderBrown:
Unfortunately as we have fairly extensive theming to achieve a consistent look across platforms (which I think is important for a learners tool) I'm not sure this will be practical.

Would love some feedback on our high-contrast theme though

ntoll:
@reithi what @ZanderBrown says.

To put this into context... the feedback we got from teachers is that they use single resources over several platforms, ergo the consistency of the look is important for them (beginners, who may be lacking in confidence, may not know what to do as a result of a trivial [to us] platform specific difference in layout).

Does this make sense? Please do let us know your thoughts, ideas and feedback.

tmontes:
Possible idea, dependent on #693:
* At least on Windows (and maybe Linux? not sure) two "start menu" entries could be created: one "default mu" that starts Mu with the default day theme, and a second one "dark mu" which people like @reithi would have probably noticed and used from the start.

dybber:
I like the idea of having two different start-menu entries, one just called "Mu", and another called "Mu (High contrast)" or similar. I guess it's the High contrast theme that would be helpful for users like @reithi.

dybber:
Oh, good find! That should be doable, I can make a PR and we can get people to test it on various systems. I guess if we can get it to work on Mac, Windows and Gnome that will be good enough for a "best effort", everyone else will have to switch manually. As Mu remembers the theme settings, this will only be necessary to do on the first run of Mu.

ZanderBrown:
> or "Gnome"

Oh no

dybber:
> > or "Gnome"
> 
> Oh no

Woops! Edited :-)

PR #1317, opened 2021-02-20
WiP: Auto-detecting dark-theme preference (see issue #328)

dybber:
Work in progress: auto-detecting dark-theme preference for Mac, Windows, GNOME to make Mu more accessible. See issue #328.

Based on the answers in this  Stackoverflow post: https://stackoverflow.com/questions/65294987/detect-os-dark-mode-in-python

Still needs to be tested on the various platforms.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -28,6 +28,7 @@ import webbrowser
 import random
 import locale
 import shutil
+import subprocess
 
 import appdirs
 from PyQt5.QtWidgets import QMessageBox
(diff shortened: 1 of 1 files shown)


mu/logic.py
```diff
@@ -855,6 +856,88 @@ def connect_to_status_bar(self, status_bar):
...
+    def detect_theme_preference(self):
+        """Detect OS dark/light theme preference.
+        Currently implemented for Mac, Windows, GNOME. For any other
+        operating systems, it will always return "day"
+        """
+        # Implementations inspired by the following Stackoverflow submissions:
+        # https://stackoverflow.com/questions/65294987/detect-os-dark-mode-in-python
+
+        def macos_prefers_dark_theme(self):
+            if sys.platform != "darwin":
```
ntoll, line 868: Surely `sys.platform == "darwin"` ..?

mu/logic.py
```diff
@@ -855,6 +856,88 @@ def connect_to_status_bar(self, status_bar):
...
+            for i in range(1024):
+                try:
+                    value_name, value, _ = winreg.EnumValue(reg_key, i)
+                    if value_name == "AppsUseLightTheme" and value == 0:
+                        return True
+                except OSError:
+                    break
+            return False
+
+        def GNOME_prefers_dark_theme(self):
```
ntoll, line 903: `gnome_prefers_dark_theme` ..? Capitalization is not PEP8-y IIRC. ;-)

mu/logic.py
```diff
@@ -855,6 +856,88 @@ def connect_to_status_bar(self, status_bar):
...
+            if sys.environ["XDG_CURRENT_DESKTOP"] != "GNOME":
```
ntoll, line 904: How about a safer: `sys.environ.get("XDG_CURRENT_DESKTOP")` ..?

mu/logic.py
```diff
@@ -855,6 +856,88 @@ def connect_to_status_bar(self, status_bar):
...
+
+            currentTheme = (
+                subprocess.run(getArgs, capture_output=True)
+                .stdout.decode("utf-8")
+                .strip()
+                .strip("'")
+            )
+
+            darkIndicator = "-dark"
+            return currentTheme.endswith(darkIndicator)
```
ntoll, line 922: Not sure if `darkIndicator` is needed. Why not just `return currentTheme.endswith("-dark")` ..?

ntoll (commented):
Looking good so far..!

ntoll:
Will need tests. :-)

dybber:
> Will need tests. :-)

Yes, this is completely untested actually, I hope some one on Windows and Ubuntu (GNOME) will test it, and I will have to make some further tests on Mac.

mu/logic.py
```diff
@@ -855,6 +856,71 @@ def connect_to_status_bar(self, status_bar):
...
+        operating systems, it will always return "day"
+        """
+        # Implementations inspired by the following Stackoverflow submissions:
+        # https://stackoverflow.com/questions/65294987/detect-os-dark-mode-in-python
+
+        def macos_prefers_dark_theme(self):
+            if sys.platform == "darwin":
+                cmd = "defaults read -g AppleInterfaceStyle"
+                result = subprocess.run(cmd.split())
+                return result.return_code == 0
```
carlosperate, line 871: It will work fine without it, but since the function is meant to return a boolean maybe we want to close it with a `return False`? Instead of the implicit `return None`.

mu/logic.py
```diff
@@ -855,6 +856,72 @@ def connect_to_status_bar(self, status_bar):
...
+            return False
+
+        def gnome_prefers_dark_theme(self):
+            if sys.environ.get("XDG_CURRENT_DESKTOP") != "GNOME":
+                return False
+            # Detect dark mode in GNOME
+            cmd = "gsettings get org.gnome.desktop.interface gtk-theme"
+            result = subprocess.run(cmd.split(), capture_output=True)
+            current_theme = result.stdout.decode("utf-8").strip().strip("'")
+            return current_theme.endswith("-dark")
```
ZanderBrown, line 906: There are sooo many ways this can break, the key points being it doesn't consider `gtk-application-prefer-dark-theme` (where the theme name is still Adwaita, though strictly speaking it would be Adwaita:dark) or HighContrastInverse

Side note: *Puts GNOME hat on* We don't currently have dark mode, there is an effort to properly spec one (which would be shared with elementary and hopefully others) but at present everything is unsupported hacks

ntoll:
Any progress on this..?

Issue #1236, opened 2021-01-11
In Linux we may exit Mu with a KeyboardInterrupt if a QProcess fails to launch and we kill PID 0


devdanzin:
In some places, like Web mode, we launch a `QProcess` and later get its PID so we can send it a `SIGINT`:

```python
    def stop_server(self):
        if self.runner:
            try:
                pid = self.runner.process.processId()
                os.kill(pid, signal.SIGINT)
            except Exception as ex:
```

However, if somehow that `QProcess` isn't running (e.g. `flask` isn't available/failed to launch) it returns 0 as PID and that kills Mu on Linux with a `KeyboardInterrupt` (because we guard against `Exception`, but only `BaseException` would catch `SIGINT`).

Same code on Windows doesn't result in Mu closing, instead we get `OSError: [WinError 87] The parameter is incorrect` in the log.

The code below can be used to check what happens in other OSes:
```python
import os, signal
try:
    os.kill(0, signal.SIGINT)
except Exception as e:
    print("All is well, we caught:", e)
```

This isn't something likely to happen, but if it ever does we never get the chance to log an error. I'd like to add a check that `pid` isn't zero nor the same as the current process (`os.getpid()`) so we don't quit Mu and maybe raise an informative exception which would get logged. Thoughts?

dybber:
I guess this is linked to #1236, I think it would be better to use QProcess.kill or QProcess.terminate to terminate process. Perhaps first attempt with QProcess.terminate (sends SIGTERM), and if the process doesn't close gracefully we force it with QProcess.kill (sends SIGKILL)

dybber:
We could add a function like this to the PythonProcessPane:

```python
def stop_process(self):
    if self.process:
        self.process.terminate()
        terminated = self.process.waitForFinished(10)
        if not terminated:
            self.process.kill()
            self.process.waitForFinished()
        self.running = False
```

and then use that in both Python-mode, Web-mode and Pygame-mode? 

Then it will try to terminate, wait up to 10 seconds, if it doesn't close within 10 seconds, it will kill the process instead.

dybber:
Unless there are any specific reasons why we use:
```python
pid = self.runner.process.processId()
os.kill(pid, signal.SIGINT)
```
for the Web-mode?

ntoll:
Memory is hazy, but yes, there is/was a reason I did it via `os.kill`. I'll have to revisit the code to remember why.

dybber:
[...]

Issue #1004, opened 2020-02-07
Flask server not shut down on quit


mlibby:
* What you were trying to do?
Quit Mu. :)

* What steps you took to make this happen:
Start the simple example "Hello, Mu" Flask example, verify that it loads in the browser. It does. Quit Mu. Refresh browser on Hello, Mu page. Start a Flask web site up from another IDE (VS Code).

* What you expected to happen:
I expected the server to respond with the site launched by VS Code

* What actually happened:
First I got a 404, then on paring the URL down to http://localhost:5000/, I got the Hello Mu page. So it looks like the Flask process kicked off by Mu is still running even though Mu is not.

* Why this difference is problematic (it may not be a bug!):
Normally I'd expect a process started within my IDE to get killed when the IDE quits.

* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 
Running Windows 10 with Mu 1.1.0.alpha.2

Please remember to attach a **copy of the full log files for Mu**.
```
2020-02-07 17:14:58,745 - root:113(run) INFO: 

-----------------

Starting Mu 1.1.0.alpha.2
2020-02-07 17:14:58,745 - root:114(run) INFO: uname_result(system='Windows', node='whippet', release='10', version='10.0.18362', machine='AMD64', processor='Intel64 Family 6 Model 142 Stepping 9, GenuineIntel')
2020-02-07 17:14:58,745 - root:115(run) INFO: Python path: ['C:\\Program Files\\Mu\\pkgs', 'C:\\Program Files\\Mu', 'C:\\Program Files\\Mu\\Python\\python36.zip', 'C:\\Program Files\\Mu\\Python\\DLLs', 'C:\\Program Files\\Mu\\Python\\lib', 'C:\\Program Files\\Mu\\Python', 'C:\\Users\\m\\AppData\\Local\\python\\mu\\site-packages', 'C:\\Program Files\\Mu\\pkgs\\IPython\\extensions']
2020-02-07 17:14:58,745 - root:116(run) INFO: Language code: en_US
2020-02-07 17:14:59,262 - mu.logic:587(__init__) INFO: Setting up editor.
2020-02-07 17:14:59,262 - mu.logic:609(__init__) INFO: Settings path: C:\Users\m\AppData\Local\python\mu\settings.json
2020-02-07 17:14:59,262 - mu.logic:610(__init__) INFO: Session path: C:\Users\m\AppData\Local\python\mu\session.json
[...]

PR #1319, opened 2021-02-22
WiP: Move child process shutdown from modes to PythonProcessPane

dybber:
This is work in progress and needs to be tested across platforms. The goal is to solve both #1004 and #1236

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -936,6 +936,18 @@ class PythonProcessPane(QTextEdit):
             self.process.start(interpreter, args)
             self.running = True
 
+    def stop_process(self):
+        if self.process:
+            self.process.terminate()
+            terminated = self.process.waitForFinished(10)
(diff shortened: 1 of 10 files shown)


tjguk:
I've only looked cursorily at the code, but I'm +1 on the general idea here.

ntoll:
@dybber this is still marked as WiP. Anything need doing to un-WiP it into a ready-to-review-and-test PR..? :+1:

ntoll:
Hi @dybber and @tjguk 

I've taken a quick look at the code and I'm also :+1: on this with the following caveat: The orphaned child processes from #1004 no longer appear on Windows.

@dybber I'm also painfully aware that you've marked this as WiP. What more were you hoping to do..?

ntoll:
OK, I've tested this on both Windows and Linux and it appears to do exactly what it says. Merging.

Issue #1321, opened 2021-02-22
ESP-mode does not always close FileManager threads properly


dybber:
When opening the file-system pane in ESP-mode, a thread is launched, but we don't close it explicitly and leave it up to the garbage collector, so errors like these can happen:

`QThread: Destroyed while thread is still running` if the user closes Mu before the thread is finished executing its task (e.g. listing a files can take a long time)

dybber:
From reading the code it seems this is also the case for the micro:bit mode

PR #1323, opened 2021-02-22
Ensure modes terminate the threads they use for flashing/file system operations before Mu quits

dybber:
This is a fix for issue #1321, where both micro:bit and ESP-mode didn't properly close the threads that they opened.


ntoll:
:wave: @dybber any chance the failing tests can be fixed on this PR..? Thank you..!

Issue #1322, opened 2021-02-22
Qt doesn't terminate nicely when trying to close it with a KeyboardInterrupt (CTRL-C)


dybber:
When trying to close Mu using CTRL-C in a terminal, you will get all kinds of errors, depending on the current code executing. Instead we want Mu to shutdown in a more controlled maner.

There are some hints here to how we might go about this: https://coldfix.eu/2016/11/08/pyqt-boilerplate/#keyboardinterrupt-ctrl-c

PR #1324, opened 2021-02-22
WiP: Handle CTRL-C interrupts from terminal, to shutdown Mu gracefully

dybber:
Installs a SIGINT interrupt handler, to catch if a user types CTRL-C in a shell to close Mu, when the interrupt occurs close Mu by calling `editor.quit()`

This is a draft, as I'm not sure if this makes any sense on Windows, and I still need to add tests.

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -24,6 +24,7 @@ from logging.handlers import TimedRotatingFileHandler
 import os
 import platform
 import sys
+import signal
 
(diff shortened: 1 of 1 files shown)


dybber:
Forgot to link this with the issue #1322

ntoll:
@dybber :wave: this is marked as a WiP.

Any progress..? Thanks..! :-)

PR #1316, opened 2021-02-20
Prevent creation of multiple venvs during testing

tjguk:
Sort out a problem creating venvs (clashing version of pyserial)
Plus ensure that no real venvs are created in Mu's standard data dir

PR #1329, opened 2021-02-24
Unpin pyserial version for the modes dependencies

carlosperate:
Cherry picking the commits from https://github.com/mu-editor/mu/pull/1316 into master instead of the splash branch.


carlosperate:
Thanks for the reviews!
The current CI hangups are known issues, so we can merge this.

PR #1341, opened 2021-03-02
Add Tiny 2040 support for Pi 2040 mode.

theLMGN:


theLMGN:
@carlosperate I was using CircuitPython, however this works with MicroPython too. Not sure where the VID & PID come from, I just checked whatr it was identifying as on my local machine.

DeviceHunt says it's from 'MCS' https://devicehunt.com/search/type/usb/vendor/16D0/device/08C7 so probably the latter.

carlosperate:
Right, so the odd thing is that the "normal" MicroPython build should configure the USB VID/PID to the values configured in the mode (0x2E8A/0x0005), and CircuitPython should be using the Adafruit VID 0x239A.
This is the first time I am seeing 0x16d0/0x08c7, are you 100% sure that were the details from your device, and not from a different USB device plugged in at the same time (or part of your PC)?

If so, where did you get the MicroPython/CircuitPython builds?

theLMGN:
This is what the Tiny 2040 shows up as when running CircuitPython

```
Tiny 2040:

  Product ID:	0x08c7
  Vendor ID:	0x16d0
  Version:	1.00
  Serial Number:	FD964D61FC47F353
  Speed:	Up to 12 Mb/s
  Manufacturer:	Pimoroni
  Location ID:	0x00200000 / 1
  Current Available (mA):	500
  Current Required (mA):	100
  Extra Operating Current (mA):	0
  Media:
Tiny 2040:
  Capacity:	7.3 MB (7,340,544 bytes)
  Removable Media:	Yes
  BSD Name:	disk7
  Logical Unit:	0
  Partition Map Type:	MBR (Master Boot Record)
  SMART status:	Verified
  USB Interface:	4
  Volumes:
CIRCUITPY:
  Capacity:	7.3 MB (7,340,032 bytes)
  Free:	7.3 MB (7,304,192 bytes)
  Writable:	Yes
  File System:	MS-DOS FAT16
  BSD Name:	disk7s1
  Mount Point:	/Volumes/CIRCUITPY
  Content:	DOS_FAT_12
  Volume UUID:	A60B15F9-A688-3D3D-8024-B8D0223D1D43`
```

carlosperate:
Thanks for USB information!

Looks like Pimoroni are using the `0x16d0` VID for CircuitPython:
- https://github.com/adafruit/circuitpython/blob/cd48c5ee8317fb94fed1fd6c430ab83d1b487ac9/ports/raspberrypi/boards/pimoroni_keybow2040/mpconfigboard.mk
- https://github.com/adafruit/circuitpython/blob/cd48c5ee8317fb94fed1fd6c430ab83d1b487ac9/ports/raspberrypi/boards/pimoroni_picosystem/mpconfigboard.mk
- https://github.com/adafruit/circuitpython/blob/cd48c5ee8317fb94fed1fd6c430ab83d1b487ac9/ports/raspberrypi/boards/pimoroni_tiny2040/mpconfigboard.mk

@theLMGN I assume the board still shows with VID `0x2E8A` and PID `0x0005` when running MicroPython?
As far as I know it's the only VID/PID configured for the rpi2040 port, unless there are other ports for these Pimoroni boards in a different repo?
https://github.com/micropython/micropython/blob/b8f4c623f9002099b0910ac873ef9e3acbcf428c/ports/rp2/tusb_port.c#L29
Where do you download the MicroPython uf2 file this board?

If you like to update the PR to add these values for the CircuitPython mode we can definitely merge that.
As far as the raspberry pi mode goes, if the MicroPython VID/PID is the same we'll leave it untouched, but if it has different ones I'd like to check where they come from before merging it.

theLMGN:
Yes. There is no special MP build for the Tiny 2040, you use regular Pi Pico buiilds

```
  Product ID:	0x0005
  Vendor ID:	0x2e8a
  Version:	1.00
  Serial Number:	000000000000
  Speed:	Up to 12 Mb/s
  Manufacturer:	MicroPython
```

mu/modes/pico.py
```diff
@@ -40,6 +40,7 @@ class PicoMode(ESPMode):
     valid_boards = [
         # VID  , PID,    Manufacturer string, Device name
         (0x2E8A, 0x0005, None, "Raspberry Pi Pico"),
+        (0x16d0, 0x08c7, None, "Tiny 2040")
```
carlosperate, line 43: As far as I understand this VID/PID only appears with CircuiPython, so we should remove this entry from this mode.

mu/modes/circuitpython.py
```diff
@@ -72,6 +72,7 @@ class CircuitPythonMode(MicroPythonMode):
         (0x239A, 0x80AC, None, "Unexpected Maker FeatherS2"),
         (0x303A, 0x8002, None, "Unexpected Maker TinyS2"),
         (0x054C, 0x0BC2, None, "Spresense"),
+        (0x16d0, 0x08c7, None, "Tiny 2040")
```
carlosperate, line 75: I believe the tests are failing because the black formatter wants a comma here.
You can run the test locally with `make check` (although if you are using Python 3.5 black won't run as it is not available)

carlosperate:
Thanks for updating the PR @theLMGN! Added a couple of minor comments.

carlosperate:
@theLMGN we'll be merging CircuitPython PRs soon, as you happy to update this PR? Thanks again for your contribution!

carlosperate:
Thanks for the update @theLMGN!
At the moment the CI is failing due to the formatting. If you run `make tidy` (or `python make.py tidy`) it will apply the `black` formatter and pass the tests.

PR #1354, opened 2021-03-06
Fix device naming for Adafruit-made CircuitPython boards

ajs256:
Currently, all CircuitPython boards using USB VID `0x239A` show up in Mu as an `Adafruit CircuitPlayground`. This is no longer true, so I have genericized the name to "Adafruit CircuitPython board".

Note: They occasionally assign their VID to community boards, so this isn't always right.


ntoll:
Will close when #1371 lands.

ajs256:
Shall I just close this now and remove it from the backlog?

ntoll:
Fixed by #1371

ntoll:
Thank you @ajs256 for the contribution. :+1:

PR #1356, opened 2021-03-10
uFlash: Add micro:bit MicroPython v2.0.0-beta.4.

carlosperate:
This PR only changes the hex embedded inside uFlash.
This hasn't been pushed to the uFlash repo yet as the tests for the latest changes are still to be finished up (sorry, it's in my todo list!), but will be pushed after that is done.

Tested it with a micro:bit V1 and V2 on macOS, but since the change is in the hex itself there shouldn't be any difference in other OS. As part of the release checklist it'll also be tested in the other platforms when testing the installers.


carlosperate:
@ntoll if possible it'd be great to get this in for the next release.

ntoll:
Thank you..!

Issue #1358, opened 2021-03-13
pip doesn't work on first run.


ntoll:
FYI @tjguk - I'm going to look into this *right now* but may drop it if it's a rabbit hole. Reporting for visibility, but I hope the fix is easy (see below).

* What you were trying to do

Start Mu for the first time and then install a new package into the clean venv.

* What steps you took to make this happen,

From a clean start, run Mu and allow it to set up a new venv and other assets. Then, without restarting Mu, try to install a package from the admin screen in Python3 mode. Screen-grab of this happening shown below.

![mu_venv_oddness](https://user-images.githubusercontent.com/37602/111033186-df3ebb80-8407-11eb-8030-8b2c5c1cb1c5.gif)

* What you expected to happen,

I should be able to see the usual installation log and pip should work as expected.

* What actually happened,

The installation log is just blank. If I restart Mu, it works as expected.

* Why this difference is problematic (it may not be a bug!),

It should work first time... ;-)

* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 

Latest master. I think something like the path to Python or `pip` in the venv is not being set properly on first run such that the admin based install packages widget cannot work.

ntoll:
OK... it looks like it's because the `settings.json` and `venv.json` files are not created until the editor is closed for the first time. I'll do a quick fix and push to master so this bug isn't in beta.2. :+1:

ntoll:
OK... that was a red herring. Still trying to figure out what's going on.

ntoll:
Parking for the moment.

TL;DR I don't think the slots are created in the right way here: 

https://github.com/mu-editor/mu/blob/master/mu/interface/dialogs.py#L647

I suspect some state in the virtualenvironment.py module isn't set on first run. I.e. the `venv` singleton is in a pre-ensure-and-created state, when it should really be re-configured post-creation. I did try calling `relocate` but that didn't fix things.

@tjguk any ideas..?

ntoll:
I'll mention this as a known bug in the release notes.

tjguk:
Just a couple more datapoints: even if I implement a "blocking" version of the unblocked run, by basically running the code to completion and then calling the output` slot, it *still* doesn't produce the output, all it does complete and the "FINISHED" text appears on the text display. 

[...]

PR #1368, opened 2021-03-16
Pip fail2

tjguk:
Force the `Pip` object to recreate its internal QProcess object on every run. This sidesteps the issue where it's first invoked as part of a worker thread during the Splash screen, causing every later invocation to run cross-thread

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -224,7 +224,7 @@ def setup_logging():
     log = logging.getLogger()
     log.setLevel(logging.DEBUG)
     log.addHandler(handler)
-    # ~ log.addHandler(stdout_handler)
+    log.addHandler(stdout_handler)
     sys.excepthook = excepthook
 
 
--- a/mu/app.py
+++ b/mu/app.py
@@ -122,7 +122,7 @@ class AnimatedSplash(QSplashScreen):
         lines = text.split("\n")
         lines.append(
             "This screen will close in a few seconds. "
-            + "Then a crash report tool will open in your browser."
+            "Then a crash report tool will open in your browser."
         )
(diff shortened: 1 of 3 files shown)


ntoll:
This looks great. I just tested it and I can confirm that no QThread warnings appear when Mu is setting up a virtual environment on first run. Furthermore, it appears that I can install packages *on first run* (fixing #1358).

However, I have a request... the current logging in the splash screen on first run is rather opaque (see below). Could we re-write the logging so they're more informative, confidence building, friendly..? :-)

mu_logging_opaque

tjguk:
Thanks @ntoll  -- yes, that was all interim debug logging while I was trying to narrow stuff down. I've removed most of it now. In fact, I was going to make a suggestion back in the other direction that we limit the splash logging handler to show only INFO and higher. (ie to ignore DEBUG). At that point we can certainly refine the INFO output so it tells a clearer story.

tjguk:
So while we've definitely got the culprit, finding the eventual solution will take a little working through. 

The long and short is that we need to ensure that, when Pip runs *after* `create_and_ensure` it uses a different QProcess object from the one used *inside* `create_and_ensure`. We can definitely do that crudely -- basically extending the quick fix I did earlier to prove the concept. But I'd like to do it more elegantly.

However... that's interacting awkwardly with the usual Qt problem: Python object lifetimes. IOW the ideal solution is a long-lived `Pip.QProcess` object, but one that distinguishes between being in a thread or not.

tjguk:
Pushed a working solution. I'll add a test later, but let's see if the tests pass ok.

Issue #1127, opened 2020-10-04
Drop Python 3.5 support.


tmontes:
(leaving this here as a reminder for future consideration, after having read that 3.5 is no longe supported by the core dev team - not that it needs to be done “now”)

IIRC, supporting 3.5 is (was, hopefully) fundamental in order to allow Mu to run on the previous version of Raspbian. Current Raspbian ships with 3.6, again, IIRC.

If that’s the case, I wonder when we could do this. IOW, how many Mu users are still running an older Raspbian version?

Dropping 3.5 would allow:
* Softening CI load.
* Progressively move to f-strings.

Thoughts? :-)

ZanderBrown:
```
$ cat /etc/os-release 
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
$ python3
Python 3.7.3 (default, Jul 25 2020, 13:03:44) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 

```

The final stretch (so 3.5) release was 2019-04-08, and didn't support the Pi 4, is 18 months (or is it 18 years....) long enough? Not sure but given 3.5 stopped getting security fixes last month I would suggest it's fairly reasonable to jump to 3.7

ZanderBrown:
> Progressively move to f-strings.

Just on this though it's worth noting we can't use f-strings with gettext

Issue #281, opened 2017-08-15
numu + circuitpython : REPL doesnt get input?


ladyada:
i got the repl working and its printing out all nicely, but if i focus on it and press ^C or ^D it doesn't cancel/restart the python program

![image](https://user-images.githubusercontent.com/1214161/29326360-3d2a9764-81b9-11e7-9e06-0e18f3b6093e.png)

i also can't get this to take input either :/
```
name = input("What is your name? ")
print("hello "+name)
```
doesn't ever print

![image](https://user-images.githubusercontent.com/1214161/29326467-98e43ab0-81b9-11e7-9c3e-80e49bbce2fa.png)

running py3.5.2, installed requirements.txt running windows 7 x64 pro with 'python3 run.py'

ladyada:
eh? i restarted twice and now it works? is there a preferences or state folder i can delete to refresh my install? 

cause i swear the REPL didnt appear the first time at all, and second time didnt have input. (same thing happened on two win7 computers)

ntoll:
If you click on the cog you'll see the logs. The first thing Mu writes to logs are things like details of your system and locations of various things such as a `settings.json` file (usually in the default location for application settings for the given OS). The log window may also reveal what the REPL was up to - I added it because we got lots of feedback from teachers who wanted to work out what their students had done to get themselves into a certain state. ;-) 

I've used the `input` builtin function on other MicroPython boards and "it just works" for me (I realise this isn't helpful to you, but it may indicate it's something to do with your local set-up).

Keep this feedback coming! It's gold-dust for stuff to fix, improve, smooth over when I get back from holiday. Until then, assume alpha quality code. ;-)

ladyada:
ok i did more work and i narrowed this down to CircuitPython expecting **DTR** to be set (e.g. terminal open) but qtserial does not actually set DTR, even if you tell it to. which is...weird. but reproducable. here's a simple script that tests DTR setting and data reads. in windows 7, at least, DTR is always 'false' 

> import sys, time
> from PyQt5.QtWidgets import QWidget, QApplication, QPushButton
> from PyQt5.QtCore import QObject, pyqtSlot, QIODevice, pyqtSignal
> from PyQt5.QtSerialPort import QSerialPort
> 
> PORT = "COM119"
> 
> class DVIFlasher(QObject):
>     ser = None
>     flash_done = pyqtSignal(int, int)
> 
>     def __init__(self):
>         super(DVIFlasher, self).__init__()
[...]

PR #1371, opened 2021-03-17
CircuitPython mode: support for multiple serial ports per board; use serial port to identify

dhalbert:
Addresses #1352.

CircuitPython can now be built with a second serial port to send data back and forth to the host, in addition to the REPL port. See #1352 for details. This PR identifies the REPL ports correctly.

- Identify REPL ports by using the `adafruit-board-toolkit` library (brand new) from pypi.org. The library uses pyserial's `serial.tools.list_ports` functionality to find ports with particular interface name strings. The REPL ports start with `"CircuitPython CDC "`. On Linux it uses only the `pyserial` library. For Windows and MacOS, the scripts for those platforms provided by `pyserial` need fixing and enhancement, and substitute versions are included in `adafruit-board-toolkit`. We are PR'ing these fixes back to `pyserial`.
- Using the REPL port interface strings means we can identify any CircuitPython board by its REPL interface name. So the code to identify by USB VID/PID can be removed, and does not need to be updated when new boards are added.
- Dropping USB VID/PID means that all CircuitPython boards are now just labeled as "CircuitPython board". Previously, boards with specific VID/PIDs were named explicitly, and all Adafruit boards were erroneously named "Adafruit CircuitPlayground".
- The list of CircuitPython modules that should not be filenames has been updated.
- `pyserial>=3.5` must be true for the `adafruit-board-toolkit` library to work properly. Previously `pyserial` was pinned at `3.4`.
- `adafruit-board-toolkit` is now an additional dependency.
- `.gitignore` now ignores `*~` and `TAGS`, which are common when using emacs.

Tested on Ubuntu 20.04, Windows 10 Pro 20H2, and MacOS 11.2.2, with two different CircuitPython 6.2.0-beta.3 boards, both with secondary USB CDC channels.


dhalbert:
Just checking, are you planning to continue Python 3.5 support? I can try to roll back `adafruit-board-toolkit`s requirement for 3.6. I see #1127 which proposes to drop 3.5.

dhalbert:
I am puzzled by the one failing test above, which is a segementation fault. If someone with privs on this repo could trigger a re-run, that would be great.

ntoll:
Hi @dhalbert - I'm sorry I've not been able to get to this sooner, but "real work" has got in the way, so currently blasting through PRs at weekends. A quick heads up, for the duration of beta, we'll be cutting a release every two weeks (with the next release, `beta.3` arriving this coming Monday). A bit of new Mu context too: the way we package Mu has changed significantly so we're in a better place for more nimble releases. TL;DR - Mu has a core set of dependencies that MUST be installed for Mu to run. Mu also has a further core set of dependencies that are installed in a virtualenv so that the various modes can run.

This work should definitely be in the beta, but I don't think it'll make Monday's. It would be great to get this into `beta.4` in just over a fortnight.

For me there are two things that need resolving, both of which are relatively trivial:

* PySerial version decision. I'm comfortable with >=3.5, but I'm less sure why PySerial is needed to be "declared" in both the Mu dependencies and virtualenv dependencies. @carlosperate what am I missing here..?
* The `adafruit_board_toolkit` module needs to be included in our packaging in some way. I suspect, since it's not a core part of Mu but is definitely a part of CircuitPython mode, that it should be in the virtualenv dependencies. @carlosperate would you agree.

Since the `adafruit_board_toolkit` isn't referenced as a dependency anywhere, the tests currently fail for me. I suspect the best place to fix this is by adding the module to the `mode_packages` list in `/mu/wheels/__init__.py`.

I hope this is all clear. As always, it's such a pleasure to collaborate and I hope you're well given such strange times. :hugs:

dhalbert:
> * The `adafruit_board_toolkit` module needs to be included in our packaging in some way. I suspect, since it's not a core part of Mu but is definitely a part of CircuitPython mode, that it should be in the virtualenv dependencies. @carlosperate would you agree.
> 
> Since the `adafruit_board_toolkit` isn't referenced as a dependency anywhere, the tests currently fail for me. I suspect the best place to fix this is by adding the module to the `mode_packages` list in `/mu/wheels/__init__.py`.

Thanks, @ntoll! I've pushed a new commit that updates `/mu/wheels/__init__.py` to include `adafruit-board-toolkit` and bumps `pyserial` to `>=3.5`. I didn't realize that was another place for dependencies. But does that cover the packaging dependencies? I looked for yet another place where `pyserial` was a dependency for packaging, and didn't find it.

mu/modes/circuitpython.py
```diff
@@ -236,6 +270,34 @@ def get_volume_name(disk_name):
...
+        # and see if any of their device names match the one passed in.
+        for comport in circuitpython_serial.repl_comports():
+            if comport.device == port_name:
+                return Device(
+                    vid,
+                    pid,
+                    port_name,
+                    serial_number,
+                    manufacturer,
+                    self.name,
```
carlosperate, line 294: I guess in this case the device will always be identified as "CircuitPython"?
I was hoping to go in the other direction and include more Adafruit board names in this mode to be able to see that in the UI (of course, always falling back to a default name for any device with the right VID, but unknown PID).

dhalbert: That's right, so it's a bit of a functionality regression. But he current list is woefully incomplete, so most boards were identifying incorrectly as "CircuitPlayground".

I already thought about adding some board-name lookup capability to the `adafruit_board_toolkit`. Maybe it would be better to do that as a separate PR in order not to pile too much on this PR? I can work on that soon, I just didn't want to hold up this PR.

carlosperate:
> PySerial version decision. I'm comfortable with >=3.5, but I'm less sure why PySerial is needed to be "declared" in both the Mu dependencies and virtualenv dependencies. @carlosperate what am I missing here..?

Not 100% sure, but I assume it's because "core" files still use pyserial:

https://github.com/mu-editor/mu/blob/7f66a3d81c42c810e29b5ea575b755c6e40f1d33/mu/modes/base.py#L26

I think there was a reason not to move that one to QtSerial, but I cannot remember. Maybe @dybber knows, but he will likely be offline for a little while (if you are reading this, no need to worry, just a tag for future reference when you are back).
Hopefully there is something in a git commit or PR, thinking about it, it might be simply because microFs uses pyserial.

> The adafruit_board_toolkit module needs to be included in our packaging in some way. I suspect, since it's not a core part of Mu but is definitely a part of CircuitPython mode, that it should be in the virtualenv dependencies. @carlosperate would you agree.

Yeah, a dependency in mu.wheels is the right approach, which I think it has been added in the latest commit, thanks @dhalbert! We should be able to remove it from setup.py now.

dhalbert:
> Yeah, a dependency in mu.wheels is the right approach, which I think it has been added in the latest commit, thanks @dhalbert! We should be able to remove it from setup.py now.

Just to be clear, so I should push another commit to remove `adafruit-board-toolkit` from `setup.py`?

mu/modes/circuitpython.py
carlosperate: > the current list is woefully incomplete, so most boards were identifying incorrectly as "CircuitPlayground".

Oh yeah, that's the main thing I wanted to do, populate that list with more board names, and ensure the `compatible_board()` method was able to first check for existing VID/PID pairs and then fallback if there is a VID/`None` entry.
(Sounds like a great use case for the new PEP 634 pattern matcher, too bad we need to support 3.5+ :)

I think we can try to match the name here with the PIDs in the original `valid_boards` list, that way we can keep it and that reduces the diff too.

dhalbert: The current list is actually very out of date, and all Adafruit boards are already wildcarded to `"Adafruit CircuitPlayground"`. We actually support about 200 boards now.
https://github.com/mu-editor/mu/blob/f56bf64ff570d6268862b0396e79295031e6aedf/mu/modes/circuitpython.py#L43-L75

dhalbert: Our answers crossed here. I'm suggesting instead that we remove the board names from the Mu codebase altogether. Instead, you'd call `adafruit_board_toolkit.circuitpython_board.name(vid, pid)` and it would give you the name, if it knew the name. Then Mu does not have to be updated constantly as new boards are added.

We will probably do some automation to keep `adafruit_board_toolkit` up to date in terms of board names. We can get the board names from circuitpython.org.

dhalbert: > remove the board names
and the pid/vids

carlosperate:
Yes, although I suspect some test might need to be updated, it's possible we've mocked out the usuage of all the other modules in `mu/wheels/__init__.py`.

dhalbert:
@carlosperate I tried removing `adafruit-board-toolkit` from `setup.py`, and then uninstalled it as well, to see whether a `pip3 install .` of mu would work. But it doesn't. It fails immediately with `ModuleNotFoundError: No module named 'adafruit_board_toolkit'`. So I think we need to leave the requirement in `setup.py` if someone is doing a pip3 install of mu instead of a packaged install.

dybber:
The dependency on pySerial comes from these lines:
https://github.com/mu-editor/mu/blob/7f66a3d81c42c810e29b5ea575b755c6e40f1d33/mu/modes/base.py#L120-L126

As far as I know, it's the only place pySerial is used. The mentioned issues predates my involvement, but reading through issue #281, it seems likely that we can actually avoid pySerial as a dependency after all. It seems back then the following was attempted:
```
self.ser.dataTerminalReady = True
```
To set the Data Terminal Ready flag with QSerialPort, but I think that is not what you're supposed to do. Instead you have to use the setter:
```
self.ser.setDataTerminalReady(True)
```
https://doc.qt.io/qt-5/qserialport.html#dataTerminalReady-prop

dhalbert:
adafruit-board-toolkit does depend on pyserial, specifically its `tools.list_ports` functionality, so it will need to be included for that reason. If it were just in `setup.py`, and mu itself did not use pyserial, then it would be brought in by dependency checking. But in the packaging you do, does that also pull in dependencies automatically?

ntoll:
Hi @dhalbert - apologies for the slowness on my part.

I just rebased and "massaged" where the dependencies should be declared. But it all looks good to me and it'll be in the upcoming beta release. All the CI is green so merging.

Thank you so much for your contribution and I hope all is well with you and yours..! :wave:

Issue #1373, opened 2021-03-17
Fix TravisCI failures


carlosperate:
Now that the test flakiness is not as bad the TravisCI tests are not usually getting stuck and we can get the test results.
We should look into what the problem is and fix it.

PR #1375, opened 2021-03-17
Tests: Skip test_Window_remember_plotter_position if QtChart not available.

carlosperate:
This was an easier fix than I though, this should make all the TravisCI checks green again.
I thought I saw some issues related to the venvs a while ago, but was likely fixed with some of the other changes in that area.
In the end the only thing left standing was a test on the plotter pane that didn't have a "skip" for platforms without QtCharts (Arm Debian with python 3.5)

Fixes https://github.com/mu-editor/mu/issues/1373.


carlosperate:
I'll merge now as this is fairly innocuous.

PR #1376, opened 2021-03-17
User the version of PyGame Zero hosted on PyPI.

ntoll:


mu/wheels/__init__.py
```diff
@@ -19,7 +19,7 @@
 mode_packages = [
     (
         "pgzero",
-        "git+https://github.com/mu-editor/pgzero.git@mu-wheel",
+        "pgzero>=1.2.1",
```
carlosperate, line 22: Minor nitpick, but we can collapse this entry in a single line now :)

ntoll: Heh.... just about to step into my penultimate meeting. Lemme see how quick I can do this... ;-)

ntoll: Turns out that's just the way black formats things..! ;-) (So no change)

PR #1388, opened 2021-03-20
Add 30s timeout to pytest run

tjguk:
Add a 30s timeout to the pytest run in the hope that -- if tests hang, they'll fail faster!

### DIFF
--- a/tests/interface/__init__.py
+++ b/tests/interface/__init__.py
@@ -1,0 +1,3 @@
+import pytest
+
+pytestmark = pytest.mark.skip("Skipping for hangs")
--- a/tests/interface/__init__.py
+++ b/tests/interface/__init__.py
@@ -1,3 +1,3 @@
-import pytest
+# ~ import pytest
 
-pytestmark = pytest.mark.skip("Skipping for hangs")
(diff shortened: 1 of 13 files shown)


tjguk:
Well that doesn't seem to have helped at all :) I can see the 30s timeout noted on each of the test runs, but 3.7 on Ubuntu latest has been running for 20+ minutes with no results and no timeout!

tjguk:
I think this is about as good as it's going to get for the moment. In sum, it looks like something in test_app is causing the hangs. For now, I've left `test_app:test_run` skipped as that seems to give the best outcome

tests/interface/__init__.py
```diff
@@ -0,0 +1,3 @@
+# ~ import pytest
+
+# ~ pytestmark = pytest.mark.skip("Skipping for hangs")
```
carlosperate, line 3: Should we delete all these commented out skips or do you still need them for more testing?

tjguk: Those can definitely go -- I'll raise a small PR. (Sorry; I've gone backwards and forwards so much over the code, trying to narrow down the test failures that I forgot about them..)

PR #1390, opened 2021-03-21
Space in username 

### LINKED ISSUE #1372: Virtual Environment fails on Windows when user path contains a space
(First reported via https://github.com/mu-editor/mu/issues/1367)

On the surface it appears that the check for the Python interpreter in `VirtualEnvironment.ensure` fails if the user's installation path contains a space -- typically because their username contains a space. Not verified yet, but the OP in that report confirmed that re-installing under a different username solved the problem.

tjguk:
(Fixes #1372)

Attempt to workaround the non-reproducible bug where a username containing a space causes the venv interpreter to fail


carlosperate:
Not sure how, but it looks like this has introduced an issue in TravisCI:

- https://travis-ci.org/github/mu-editor/mu/builds
- https://travis-ci.org/github/mu-editor/mu/jobs/763813148#L2028

PR #1400, opened 2021-03-23
Restore tests with spaces in venv with TravisCI Ubuntu Arm fix

carlosperate:
Older versions of virtualenv create a pip file (a python file without extension) in the venv bin folder that uses a "normal"
shebang pointing to the python executable inside the venv. Something like
```
#!"/path/to/venv with spaces/bin/python"
```

If the path contains a space the shell shebang parsing can break and stop processing the path after it (even if the space is escaped with `\ `). When this happens it means that `mu.virtualenv` fails trying to call the `venv/bin/pip` executable with a `bad interpreter: No such file or directory` error.

From version 16.* of virtualenv the shebang has been changed to use a format that should work fine with spaces:

```
#!/bin/sh
'''exec' "/path/to/venv with spaces/bin/python3" "$0" "$@"
' '''
```

So this PR adds #1390 again and the min version in setup.py.

### DIFF
--- a/mu/virtual_environment.py
+++ b/mu/virtual_environment.py
@@ -521,10 +521,11 @@ class VirtualEnvironment(object):
         #
         process = subprocess.run(
             [
-                self.interpreter,
+                os.path.basename(self.interpreter),
                 "-c",
                 'import sys; print("%s%s" % sys.version_info[:2])',
             ],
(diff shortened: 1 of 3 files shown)


tjguk (approved):

tjguk:
Thanks for doing the spadework there, @carlosperate

Issue #1422, opened 2021-03-30
Right click menu in the Files panel crashes Mu


marc0janssen:
macOS: Big Sur
Latest Mu-editor
No logfile available

Traceback (most recent call last):
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/interface/panes.py", line 607, in contextMenuEvent
    local_filename = self.currentItem().text()
AttributeError: 'NoneType' object has no attribute 'text'

ntoll:
Hi @marc0janssen,

Could you please attach your log file..? Instructions here: https://codewith.mu/en/howto/1.1/read_logs

Thank you..!

marc0janssen:
> Hi @marc0janssen,
> 
> Could you please attach your log file..? Instructions here: https://codewith.mu/en/howto/1.1/read_logs
> 
> Thank you..!

Here is the log... but I can only select the log after I restart MU-editor... I don't think it is very useful



2021-03-31 09:44:58,494 - mu.interface.main:942(show_message) DEBUG: Could not find an attached BBC micro:bit.
2021-03-31 09:44:58,508 - mu.interface.main:943(show_message) DEBUG: Please make sure the device is plugged into this computer.

The device must have MicroPython flashed onto it before the file system will work.

Finally, press the device's reset button and wait a few seconds before trying again.
2021-03-31 09:45:01,968 - mu.interface.main:942(show_message) DEBUG: Could not find an attached device.
2021-03-31 09:45:01,969 - mu.interface.main:943(show_message) DEBUG: Please make sure the device is plugged into this computer.

It must have a version of MicroPython (or CircuitPython) flashed onto it before the REPL will work.

Finally, press the device's reset button and wait a few seconds before trying again.
2021-03-31 09:45:03,913 - mu.modes.base:474(toggle_repl) INFO: Toggle REPL on.
2021-03-31 09:45:42,924 - mu.logic:737(check_usb) INFO: microbit device connected on port: /dev/cu.usbmodem14302(VID: 0x0D28, PID: 0x0204, manufacturer: 'ARM')
2021-03-31 09:45:44,914 - mu.modes.base:61(get_default_workspace) INFO: Using workspace /Users/marco/mu_code from settings file
2021-03-31 09:45:44,940 - mu.modes.microbit:570(toggle_files) INFO: Toggle filesystem on.
2021-03-31 09:45:48,665 - root:174(excepthook) ERROR: Unrecoverable error
[...]

PR #1432, opened 2021-04-01
Fix crash when Files pane has empty lists and right click menu is used

carlosperate:
To fix it stop showing any right click menu if no entry in the list is selected, which also applies when the list is empty.

I've also added more tests for this area as there was a couple of missing cases.

And replaced `micro:bit` with `device` in the Files pane texts, as this applies to all MicroPython modes enabling this feature, not just the micro:bit.

Fixes https://github.com/mu-editor/mu/issues/1422.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -534,12 +534,15 @@ class MicroPythonDeviceFileList(MuFileList):
         self.list_files.emit()
 
     def contextMenuEvent(self, event):
+        menu_current_item = self.currentItem()
+        if menu_current_item is None:
+            return
         menu = QMenu(self)
         delete_action = menu.addAction(_("Delete (cannot be undone)"))
         action = menu.exec_(self.mapToGlobal(event.pos()))
         if action == delete_action:
             self.disable.emit()
-            microbit_filename = self.currentItem().text()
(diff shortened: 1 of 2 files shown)


mu/interface/panes.py
```diff
@@ -597,16 +598,19 @@ def on_get(self, microbit_file):
...
         ).format(microbit_file)
         self.set_message.emit(msg)
         self.list_files.emit()
 
     def contextMenuEvent(self, event):
-        menu = QMenu(self)
+        menu_current_item = self.currentItem()
+        if menu_current_item is None:
+            return
         local_filename = self.currentItem().text()
```
ntoll, line 610: Could we re-use `menu_current_item` here..?

ntoll (approved):
LGTM :+1: (see minor nitpick)

mu/interface/panes.py
carlosperate: Yes, thanks for the catch! I removed the fix temporarily to check the new tests failed and forgot to re-add that here.

carlosperate:
Applied the review comment (thanks @ntoll!), and all tests pass. I'll merge :shipit:

Issue #1444, opened 2021-04-07
Issues with the latest jupyter-client version 6.1.13


carlosperate:
Released this mornig:
- https://github.com/jupyter/jupyter_client/releases/tag/6.1.13
- Diff: https://github.com/jupyter/jupyter_client/compare/6.1.12...6.1.13

These issues are reverted when downgrading to 6.1.12, but need to investigate further.

This is causing issues running the tests in Python 3.5:

```
$ python make.py check
============================= test session starts ==============================
platform darwin -- Python 3.5.10, pytest-6.1.2, py-1.10.0, pluggy-0.13.1 -- /Users/runner/hostedtoolcache/Python/3.5.10/x64/bin/python
cachedir: .pytest_cache
Using --random-order-bucket=module
Using --random-order-seed=978133

rootdir: /Users/runner/work/mu/mu, configfile: pytest.ini
plugins: random-order-1.0.4, cov-2.11.1, timeout-1.4.2
Unable to write settings to /Users/runner/Library/Application Support/mu/venv.json:
collecting ... collected 675 items / 13 errors / 662 selected
{

  "dirpath": "/Users/runner/Library/Application Support/mu/mu_venv-35-20210407-101553"
==================================== ERRORS ====================================
}
______________________ ERROR collecting tests/test_app.py ______________________
Traceback (most recent call last):
../../../hostedtoolcache/Python/3.5.10/x64/lib/python3.5/site-packages/_pytest/python.py:571: in _importtestmodule
  File "/Users/runner/work/mu/mu/mu/settings.py", line 181, in save
    mod = import_path(self.fspath, mode=importmode)
    with open(saving_to_filepath, "w", encoding="utf-8") as f:
../../../hostedtoolcache/Python/3.5.10/x64/lib/python3.5/site-packages/_pytest/pathlib.py:517: in import_path
FileNotFoundError: [Errno 2] No such file or directory: '/Users/runner/Library/Application Support/mu/venv.json'
    importlib.import_module(module_name)
../../../hostedtoolcache/Python/3.5.10/x64/lib/python3.5/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:985: in _gcd_import
    ???
<frozen importlib._bootstrap>:968: in _find_and_load
    ???
<frozen importlib._bootstrap>:957: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:673: in _load_unlocked
[...]

PR #1445, opened 2021-04-07
dependencies: Temporarily pin old version of jupyter-client.

carlosperate:
As a workaround for https://github.com/mu-editor/mu/issues/1444.


tjguk:
LGTM; we obviously need to dive further into the issue, but this works for now

carlosperate:
I've rebased to fix the conflict in setup.py from the latest merge, and refined the comment. I'm not totally sure we'll need this PR, I guess that'll depend on the next jupyter-client and qtconsole release.

ntoll:
I had to resolve a conflict introduced by the recent Adafruit update... but that was simple to do. Merging.

PR #1462, opened 2021-04-09
docs: Add PyQtChars package for Raspberry Pi install instructions

carlosperate:
The package is only available from Buster, so this is noted as well in the docs.


carlosperate:
@ntoll if you enable GitHub integration in redathedocs we can get temp PR builds of the docs for preview before merging.

ntoll:
Hmm.... I thought I *had* configured GitHub integrations in ReadTheDocs. :-/ Will check/change now. Merging in any case. :-)

ntoll:
Hmm.... just checked... I do have GitHub integration switched on. What exactly am I looking for..?

carlosperate:
You might have to remove and re-add it again. I think the last time I configured it the hooks didn't "install" correctly in an old project.

Issue #930, opened 2019-10-08
Unify black configuration for cross platform use of make command


ntoll:
The configuration of the behaviour of Black should be consistent across all platforms and expressed in a single location.

See the context from @tmontes here:

https://github.com/mu-editor/mu/pull/929#issuecomment-539467092

carlosperate:
PR: https://github.com/mu-editor/mu/pull/1464

PR #1464, opened 2021-04-09
Unify black config to a single place.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/930.

>The configuration of the behaviour of Black should be consistent across all platforms and expressed in a single location.
>
>See the context from @tmontes here:
>
>https://github.com/mu-editor/mu/pull/929#issuecomment-539467092

The make logic has been slowly been shifting to make.py, so this follows that pattern and the Makefile `tidy` and `black` commands now simply call make.py, this way we have a single source for the black configuration.

Also adds the `clean` step to the `tidy` and `black` commands  in make.py, as this was done originally in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,16 +56,10 @@ coverage: clean
 	export LANG=en_GB.utf8
 	pytest -v --random-order --cov-config .coveragerc --cov-report term-missing --cov=mu tests/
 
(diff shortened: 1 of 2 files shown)


ntoll:
LGTM :shipit:

PR #1465, opened 2021-04-09
CI: Increase timeout for the make check step

carlosperate:
In some cases `make check` doesn't have enough time to finish. I noticed this once a few days ago (didn't capture the log) and then again today:
https://github.com/mu-editor/mu/runs/2307637871?check_suite_focus=true#logs

So this PR simply increases the timeout a couple of minutes.


carlosperate:
Again: https://github.com/mu-editor/mu/runs/2308125984?check_suite_focus=true

carlosperate:
Fairly simple PR, so I'll merge.

PR #1470, opened 2021-04-11
Update what dependencies are included in setup.py and wheels/__init__.py

carlosperate:
Each commit explains the individual changes:

- nudatus is moved back as a Mu basic dependency
    - It is imported directly by Mu in the micro:bit mode, and not via the user venv, so it needs to be an `install_requires` dependency and can be removed from wheels/__init__.py.
- PySerial is removed from `wheels/__init__.py` as it not used by Mu from the user venv, it is needed to run Mu itself and so it's a `install_requires` dependency.
    - Technically it is needed for esptool, but it is already listed as a dependency of that package 
- As the `qtconsole` and `pyserial` dependencies in setup.py are needed to run the Mu editor and cannot be moved to the venv wheels, I've removed the comments suggesting that.
- `black` is removed as a testing requirement, because it is already a Mu dependency in `install_requires`.

### DIFF
--- a/mu/wheels/__init__.py
+++ b/mu/wheels/__init__.py
@@ -21,7 +21,6 @@ mode_packages = [
     ("flask", "flask==1.1.2"),
     ("serial", "pyserial>=3.0"),
     ("qtconsole", "qtconsole==4.7.4"),
-    ("nudatus", "nudatus>=0.0.3"),
     ("esptool", "esptool==3.*"),
 ]
 WHEELS_DIRPATH = os.path.dirname(__file__)
--- a/mu/wheels/__init__.py
+++ b/mu/wheels/__init__.py
@@ -19,7 +19,6 @@ logger = logging.getLogger(__name__)
 mode_packages = [
     ("pgzero", "pgzero>=1.2.1"),
     ("flask", "flask==1.1.2"),
-    ("serial", "pyserial>=3.0"),
     ("qtconsole", "qtconsole==4.7.4"),
     ("esptool", "esptool==3.*"),
 ]
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ install_requires = [
     #
     "qtconsole==4.7.4",
     "pyserial==3.4",
+    "nudatus>=0.0.3",
     # `flake8` is actually a testing/packaging dependency that, among other
     # packages, brings in `pycodestyle` and `pyflakes` which are runtime
     # dependencies. For the sake of "locality", it is being declared here,
@@ -71,7 +72,6 @@ extras_require = {
         # under the runtime venv?
         #
         'black>=19.10b0;python_version > "3.5"',
-        "nudatus",
     ],
     "docs": [
         "sphinx",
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -31,12 +31,9 @@
     + ';"arm" not in platform_machine and "aarch" not in platform_machine',
     "PyQtChart==5.13.1"
     + ';"arm" not in platform_machine and "aarch" not in platform_machine',
-    #
-    # FIXME: Maybe should be in a mode?
-    # qtconsole, pyserial
-    #
     "qtconsole==4.7.4",
     "pyserial==3.4",
+    "nudatus>=0.0.3",
```
ZanderBrown, line 36: May or may not be relevant: 0.0.5 is the first to provide a wheel on PyPI

carlosperate: Since this was the original pinned version when the upstream Debian package was created I thought it'd be best to leave it the same in case it is the only version available over there.

mu/wheels/__init__.py
```diff
@@ -19,9 +19,7 @@
 mode_packages = [
     ("pgzero", "pgzero>=1.2.1"),
     ("flask", "flask==1.1.2"),
-    ("serial", "pyserial>=3.0"),
     ("qtconsole", "qtconsole==4.7.4"),
```
carlosperate, line 22: Tried to remove this one as well, but get the following error when launching the Mu REPL:

```
  File "/Users/microbit-carlos/workspace/mine/mu-upstream/mu/modes/python3.py", line 109, in start_kernel
    self.repl_kernel_manager.start_kernel()
  File "/Users/microbit-carlos/workspace/mine/mu-upstream/mu/modes/python3.py", line 42, in start_kernel
    kernel_cmd, kw = self.pre_start_kernel(**kw)
  File "/Users/microbit-carlos/workspace/mine/mu-upstream/.venv/lib/python3.6/site-packages/jupyter_client/manager.py", line 292, in pre_start_kernel
    kernel_cmd = self.format_kernel_cmd(extra_arguments=extra_arguments)
  File "/Users/microbit-carlos/workspace/mine/mu-upstream/.venv/lib/python3.6/site-packages/jupyter_client/manager.py", line 210, in format_kernel_cmd
    cmd = self.kernel_spec.argv + extra_arguments
  File "/Users/microbit-carlos/workspace/mine/mu-upstream/.venv/lib/python3.6/site-packages/jupyter_client/manager.py", line 114, in kernel_spec
    self._kernel_spec = self.kernel_spec_manager.get_kernel_spec(self.kernel_name)
  File "/Users/microbit-carlos/workspace/mine/mu-upstream/.venv/lib/python3.6/site-packages/jupyter_client/kernelspec.py", line 230, in get_kernel_spec
    raise NoSuchKernel(kernel_name)
jupyter_client.kernelspec.NoSuchKernel: No such kernel named mu_venv-36-20210412-001637
```

QtConsole has more dependencies than needed for the venv, so maybe we can replace it with something else, but this is simpler for now.

tjguk:
For me, please go ahead and merge this @carlosperate  -- we'll clearly have to loop back round again later, especially with the QtConsole stuff, but let's move it forward for now

PR #1487, opened 2021-04-18
Add logging and smooth out flow for virtual environment creation

tjguk:


ntoll:
Tests appear to (mostly) pass (and work for me), so merging.

PR #1491, opened 2021-04-19
Remove adafruit_board_toolkit from user venv and lock major version of PySerial in setup.py

carlosperate:
This will likely fix the pip upgrades issues.


ntoll:
The new update to the CircuitPython mode requires `adafruit_board_toolkit`. Not sure if this change is problematic.... I don't think so, because `adafruit_board_toolkit` is referenced in setup.py, right..?

carlosperate:
Yep, is only imported in the CircuitPython mode, so only needed in setup.py (when I wrote my original PR comment I wasn't too familiar with how the dependencies related to the different environments). I've added a small note about that in the commit message as well as I always go back through the git history to see why we changed/locked/add/removed dependencies.

Issue #1518, opened 2021-04-28
Black breaking changes in latest releases (CI will fail until fixed)


carlosperate:
I think we might need to lock black to an older version:
- New/tweaked rules means some things that passed before, now fails
    - This is fine if the formatted output from the newest version still pass in the old one
- Running the tests it looks like `black.PY36_VERSIONS` is no longer available

carlosperate:
Okay, so `black.PY36_VERSIONS` was removed in https://github.com/psf/black/pull/1717 in v21.4b0 released 3 days ago.

I also notice that depending the Black version it formats some things slightly differently.

I'll address both these things in a PR.

carlosperate:
PR https://github.com/mu-editor/mu/pull/1519 created.

PR #1519, opened 2021-04-28
Update Tidy button usage of black to be compatible with latest version 

carlosperate:
This PR contains two things:
- The fix to the black deprecation of `PY36_VERSIONS` in https://github.com/mu-editor/mu/commit/5587168ca89cac632c487541d504dac8be4de788
- Reformatting of the source code so that it works in the minimum version of black supported and the latest
    - Had to manually massage the formatting a bit (mostly removing trailing commas in single line argument lists) to ensure it passes with both versions.

Fixes https://github.com/mu-editor/mu/issues/1518.

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -1139,7 +1139,9 @@ class Editor(QObject):
             self._load(path)
 
     def direct_load(self, path):
-        """ for loading files passed from command line or the OS launch"""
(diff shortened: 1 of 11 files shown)


ntoll:
Thank you @carlosperate - as always..! :star:

Issue #1472, opened 2021-04-13
venv subprocesses do not log their stdout/stderr when the exit is non-zero


carlosperate:
For example this case from https://github.com/mu-editor/mu/issues/1471, but there is many more examples in other crash reports ([other examples](https://github.com/mu-editor/mu/issues?q=returned+non-zero+exit+status+1)):

```
2021-04-12 00:15:47,708 - mu.virtual_environment:579(create) INFO: Creating virtualenv: C:\Users\bnystrom\AppData\Local\python\mu\mu_venv-38-20210412-001547
2021-04-12 00:15:47,709 - mu.virtual_environment:580(create) INFO: Virtualenv name: mu_venv-38-20210412-001547
2021-04-12 00:15:54,871 - root:174(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 442, in ensure_and_create
    self.ensure()
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 467, in ensure
    self.ensure_path()
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 480, in ensure_path
    raise VirtualEnvironmentError(message)
mu.virtual_environment.VirtualEnvironmentError: C:\Users\bnystrom\AppData\Local\python\mu\mu_venv-38-20210412-001122 does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\app.py", line 162, in run
    raise ex
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\app.py", line 149, in run
    venv.ensure_and_create(self.display_text)
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 449, in ensure_and_create
    self.create()
  File "C:\Users\bnystrom\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 583, in create
    subprocess.run(
[...]

Issue #1421, opened 2021-03-30
Mu doesn't verify if all required (or unexpected) wheels are present in mu/wheels/


carlosperate:
It basically it (kind of) assumes that if any wheel is present, all required wheels are present.

https://github.com/mu-editor/mu/blob/ad16f421b091f82b50124a176df429398482011f/mu/virtual_environment.py#L644-L649

Of course, if a wheels is missing, it will then fail the `ensure_key_modules()` step, so it will fail (correctly) at that point during the venv creation. But other than that it will not try to download the wheels at any stage. This might be good, if we don't want to have the Mu from installers trying to download wheels during the start up process. But it also means it misses a chance to recover if an internet connection is present.

This will initially mostly affect developer environments, as that is the easiest way to have differences between  `mu.wheels.__init__` and the present wheels, in this case we "just" need to remember to manually remove all the wheels from the `mu/wheels/` directory.

However, I think it could also be a problem for users pip upgrading Mu from PyPI. Does pip remove the old mu folder before installing the new? would the old wheels stay and cause this issue?

Perhaps we should verify in `install_baseline_packages` (the code snippet above) that all the wheels required in `mu.wheels.__init__` are present?

carlosperate:
> However, I think it could also be a problem for users pip upgrading Mu from PyPI. Does pip remove the old mu folder before installing the new? would the old wheels stay and cause this issue?

Based on https://github.com/mu-editor/mu/issues/1488 I think this is an issue for pip upgrading Mu.

carlosperate:
Just to add a note here as well, based on https://github.com/mu-editor/mu/issues/1496, we can encounter the same issue when users update Mu via the Windows installer. It looks like the update overwrites existing files, but also leaves behind any files from the previous installation.

tjguk:
My current thinking -- which I don't have time to do tonight -- is to give each mu-version its own directory under `wheels/`. (Could be a zip file, but doesn't seem to add much).

That way, the download process for a given version writes the files it needs, and the venv create process reads the right ones, too.

carlosperate:
PR https://github.com/mu-editor/mu/pull/1650

carlosperate:
@SteveChoatie @prefontdon @parsko41 @Elikyals @dcarmean @soarek @mitag @webbbn @donxw

Hi everyone! 👋

The issues you were having with Mu should have been resolved with the beta.5 release: https://github.com/mu-editor/mu/releases/tag/1.1.0-beta.5

[...]

PR #1530, opened 2021-05-03
SUPERSEDED - Detect version changes in Mu for venv

tjguk:


mu/virtual_environment.py
```diff
@@ -464,6 +474,34 @@ def quarantine_failed_venv(self):
...
+        # Now reinstall the original user packages
+        #
+        logger.debug("About to reinstall user packages: %s", user_packages)
+        self.install_user_packages(user_packages)
+
+        #
+        # Finally, quarantine the previous virtual environment with
+        # a marker to say it's been superseded
+        #
+        self.quarantine_venv(reason="SUPERSEDED")
```
carlosperate, line 503: Hasn't `self.relocate(self._generate_dirpath())` a few lines up changed this instance to be a new virtualenv? If that's the case, isn't this quarantining the newly created venv?

mu/virtual_environment.py
```diff
@@ -622,6 +692,7 @@ def create(self):
         self.install_baseline_packages()
         self.register_baseline_packages()
         self.install_jupyter_kernel()
+        self.settings["mu_version"] = mu_version
```
carlosperate, line 695: Just to confirm, will this end up in `settings.json` or `venv.json`? If it's `settings.json` then I think this will be the only setting in that file (right now), everything else is in `session.json`.

mu/virtual_environment.py
```diff
@@ -478,24 +516,52 @@ def ensure_and_create(self, emitter=None):
...
+                    venv_mu_version = self.settings.get(
+                        "mu_version", "-no-version-"
+                    )
+                    if venv_mu_version != mu_version:
+                        logger.warning(
+                            "Venv created by Mu version %s; Current Mu is version %s",
+                            venv_mu_version,
+                            mu_version,
+                        )
+                        self.recreate()
```
carlosperate, line 546: I guess this means every time we use a different Mu version it will recreate the venv? Even if venv was "compatible".
Would it be good enough to `self.ensure()` first before recreating if the version is different?

tjguk: Well that's basically what we're doing right now, and we then run into the problems of having two slightly different versions of whatever. I started off trying pick things apart -- uninstalling older versions of baseline packages etc. -- but I came to the conclusion that creating a new venv and porting across the user packages was a safe option.

mu/virtual_environment.py
ntoll: Playing devils' advocate for the "other half" of the "convinced" stakes.... having everything in a single json file means you only have one place to look for such configuration-y stuff.

mu/virtual_environment.py
carlosperate: Right, but looking a few lines up, from `self.relocate(self._generate_dirpath())`:
- The `self._generate_dirpath()` generates the string for the new venv path (let's call it `/path/to/my_newly_crete_venv`)
- `self.relocate()` does `self.path = "/path/to/my_newly_crete_venv"` to this new value
- `self.create()` creates the venv stored in `self.path`
- `self.install_user_packages(user_packages)` installs all the user packages
- And then this line quarantines `self.path`? So it will go from `/path/to/my_newly_crete_venv` to `/path/to/my_newly_crete_venv.SUPERSEDED`. How does it quarantine the previous venv?

mu/virtual_environment.py
carlosperate: > Well that's basically what we're doing right now, and we then run into the problems of having two slightly different versions of whatever. 

Right, yes, but that's only for the cases where upgrading has incompatibilities. We probably had quite a few people upgrading Mu from beta.2 to beta.3 without having to recreate the venv, and without issues.
The thing that makes me a bit nervous is that sooner or latter at the point where we "port" the user packages to  a new venv, then something will fail due to the pip resolver finding incompatibilities. So the longer we can reuse a venv the better, the idea being that if something is not compatible `self.ensure()` should fail anyway? 

> I started off trying pick things apart -- uninstalling older versions of baseline packages etc. -- but I came to the conclusion that creating a new venv and porting across the user packages was a safe option.

Oh yeah, 100% agree.

mu/virtual_environment.py
```diff
@@ -376,9 +381,13 @@ def run_subprocess(self, *args, **kwargs):
         logger.debug(
             "Process returned %d; output: %s",
             process.returncode,
-            compact(process.stdout.decode("utf-8")),
+            compact(
+                process.stdout.decode(sys.stdout.encoding, errors="replace")
+            ),
```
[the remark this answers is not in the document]
carlosperate: There are more cases in the crash reports where I *think* some pip stdout/stderr might be missing. I'll start capturing new ones here: https://github.com/mu-editor/mu/issues/1472#issuecomment-839654636

mu/virtual_environment.py
tjguk: Damn. I think you're right. I'll quarantine first, but any failure to quarantine should simply log and move on, so even if there's some kind of virus-checking issue, we should still create the new venv.

mu/virtual_environment.py
```diff
@@ -370,15 +375,18 @@ def run_subprocess(self, *args, **kwargs):
         process = subprocess.run(
             list(args),
             stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT,
+            stderr=subprocess.PIPE,
             **kwargs
         )
+        output = process.stdout.decode(sys.stdout.encoding, errors="replace") + process.stderr.decode(sys.stderr.encoding, errors="replace")
```
carlosperate, line 381: Maybe we can also add a short string in front of stderr that makes searching in logs easier?

mu/virtual_environment.py
```diff
@@ -487,15 +530,53 @@ def ensure_and_create(self, emitter=None):
...
+                #
+                # If any of the ensure/create steps cause a VirtualEnvironmentError
+                # retry up to a maximum number in case we've just been unlucky
+                # with network, file contention etc.
+                #
+                logger.error(exc.message)
+                self.quarantine_venv()
                 if n < n_tries:
                     self.relocate(self._generate_dirpath())
                     try_to_create = True
```
carlosperate, line 582: Any reason not to use `self.recreate()` here?

mu/virtual_environment.py
```diff
@@ -618,10 +703,10 @@ def create(self):
         before re-raising the error.
         """
         self.create_venv()
-        self.upgrade_pip()
```
carlosperate, line 621: If `upgrade_pip()` is now dead code we might as well remove it as well.

tjguk: Done

mu/virtual_environment.py
tjguk: Done

mu/virtual_environment.py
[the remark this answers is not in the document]
tjguk: Unless I've missed something (certainly possible) I'm not sure what I can do differently here. It might be that my breaking out stdout/err into two streams will help. I don't think it will hinder, so let's go for that and hope it improves things.

mu/virtual_environment.py
tjguk: Ok; I'm going to stick with the separate `venv.json` for now simply because I don't want to change too much. But I'm definitely open to a later merge of the various files.

mu/virtual_environment.py
tjguk: I think for this release I'm going to stay with the over-cautious approach; otherwise we'll never get the release out. I entirely agree that there's a risk in the other direction: that we might "damage" a perfectly good venv. That said, we have a fallback position (albeit a manual one) because we don't remove the existing venv and we can always update `venv.json` to point to it.

mu/virtual_environment.py
tjguk: Changes

mu/virtual_environment.py
tjguk: I think I was going for a broad symmetry around the loop, but I will take a look and see if a `recreate` makes more sense.

carlosperate:
@tjguk the tests fail on my system (and I believe also in the CI) because I didn't have the `mu_version` key in the venv.json file. 
If I run the application (which adds a value) then they pass.
If I remove the file, they fail again.

tjguk:
> 
> 
> @tjguk the tests fail on my system (and I believe also in the CI) because I didn't have the `mu_version` key in the venv.json file.
> If I run the application (which adds a value) then they pass.
> If I remove the file, they fail again.

Thanks; that should be an easy fix then.

tjguk:
Ok; so there was a bug in the logic which detected a failure via a QProcess. We were using the exitStatus which is not the same as the exitCode. We're now detecting a pip install failure and dumping the output to log and -- via an Exception -- to the crash-handler.

So that's definitely a step forward.

Of course this only takes us one step forward; we'll still have to deal with the pip-conflict/whatever situation is actually causing the underlying user's failure.

tjguk:
... which, going by #1421, will often be a pip conflict because we have old and new wheels both present.

tjguk:
I realised that this PR was sprawling not a little. I'm breaking it out in smaller chunks which build on each other and which, hopefully, easier to review separately:

#1622 -- improves the logging from subprocesses, the handling of exit conditions from subprocesses, and test coverage in general for the `virtual_environment` module

#1624 -- adds to #1622 the detection of a change of Mu version when we pre-emptively recreate the user's venv

The remaining PR, not yet ported from this (#1530) branch will add functionality to handle changes of downloaded wheels across Mu versions, almost certainly by using version-tagged zip files.

PR #1563, opened 2021-05-12
Add paste to REPL

ntoll:
This PR implements (and includes tests for) a mechanism for quickly pasting content from the editor pane to an active REPL. The context menu (accessed by right/alt clicking in the editor widget) is only active if two conditions are met:

1. There is some text selected in the editor pane to paste into the REPL.
2. There is an active REPL open.

In addition, single line fragments of code can be pasted as well as multi line code. If the highlighted code is indented in a way that wouldn't work with the REPL, Mu tries to correct this by looking at the indent level of the first selected line. I've also added unit tests to exercise all code paths.

You can see this in action in the following screenie:

paste_demo_full

As the demo shows, I've ensured that pasting text into a MicroPython-ish REPL is enclosed in the `CTRL-E` and `CTRL-D` commands to switch off MicroPython's native "intelligent indent" feature of the REPL.

@tjguk, @carlosperate I've tested this on Linux. Would be helpful to know if this works on Windows and OSX (thank you). I expect it to work across all MicroPython/CircuitPython devices (I've checked micro:bit and CircuitPython) and in Python3 REPL.

Any thoughts on pasting into the Python "REPL" after running a script in Python3 mode..? This currently doesn't work, but I think we probably need it, and I've not had the time over my lunch break to implement it yet (but wanted to give you a heads up of what this looks like).

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -85,8 +85,10 @@ class EditorPane(QsciScintilla):
     Represents the text editor.
     """
 
-    # Signal fired when a script or hex is droped on this editor
+    # Signal fired when a script or hex is droped on this editor.
     open_file = pyqtSignal(str)
+    # Signal fired when a context menu is requested.
+    context_menu = pyqtSignal()
 
     def __init__(self, path, text, newline=NEWLINE):
         super().__init__()
@@ -122,6 +124,12 @@ class EditorPane(QsciScintilla):
         self.setModified(False)
         self.breakpoint_handles = set()
         self.configure()
+
+    def contextMenuEvent(self, event):
+        """
+        A context menu (right click) has been actioned.
+        """
+        self.context_menu.emit()
 
     def wheelEvent(self, event):
         """
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -38,8 +38,9 @@ from PyQt5.QtWidgets import (
     QTabBar,
     QPushButton,
     QHBoxLayout,
+    QMenu,
 )
-from PyQt5.QtGui import QKeySequence, QStandardItemModel
+from PyQt5.QtGui import QKeySequence, QStandardItemModel, QCursor
 from mu import __version__
 from mu.interface.dialogs import (
     ModeSelector,
@@ -461,6 +462,8 @@ class Window(QMainWindow):
             # Bubble the signal up
             self.open_file.emit(file)
 
+        new_tab.context_menu.connect(self.on_context_menu)
+
         self.tabs.setCurrentIndex(new_tab_index)
         self.connect_zoom(new_tab)
         self.set_theme(self.theme)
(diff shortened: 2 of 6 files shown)


ntoll:
OK... added support for the `PythonProcessRunner` (used for the "REPL" when Python3 scripts run).

carlosperate:
Based on the GIF it looks like this replaces the current context menu?
image

Would it be better for this to be an additional entry?

ntoll:
@carlosperate wait, what..? We already have a context menu..?!?!?!? :-) #doh

Lemme arrange things so the new entry appears in the existing menu (or not, depending on the context). I'm sure I checked this (but obviously didn't). My bad, and I feel rather silly.

ntoll:
OK... it looks like this is just the default context menu (hence why I couldn't find it in the code when I checked - I just assumed we didn't have one as a result). Obviously we need to "inherit" / modify the default context menu. I'll figure out how over lunch.

ntoll:
OK.... Took a bit of working out how best to do this, but I think, overall, the code is simpler as a result. In any case, the default context menu always displayed and I insert a new "paste to repl" option at the top. See screenie to see it in action in various contexts:

revised_paste


Thanks @carlosperate for the eagle eyes spotting of the missing default context menu. This PR is **vastly** improved by the resulting revision. :-)

carlosperate:
>Any thoughts on pasting into the Python "REPL" after running a script in Python3 mode..? This currently doesn't work, but I think we probably need it, and I've not had the time over my lunch break to implement it yet (but wanted to give you a heads up of what this looks like).

Has this been already implemented? If the script has ended it drops into the REPL, and that kind of work? no? Is not the same as the iPython REPL, there is likely some issues to resolved, but I'm not 100% sure if this is what you meant.


Is it meant to drop the code into the debugger as well?
image

Same with the pgzero "play" mode, Web mode "run", etc

carlosperate:
Also, in the Micropython mode it enters the paste mode, and exists it after pasting the content, which automatically executes it.
In the Python 3 mode it only pastes the code, without executing it.

Was there a reason not to execute the code as well in Python mode?

carlosperate:
Right, so in a single line it copy only the selected text, as expected:
image

But in a multiline it copies the entire lines:
image

ntoll:
Hi @carlosperate - thanks for all this feedback, exactly what I was hoping for. Here's my plan:

* Yes, I did update things so the Python process pane (as well as the IPython pane) can accept copy/paste.
* The difference between MicroPython and not-MicroPython is interesting. As you know, the MicroPython REPL tries to "intelligently" indent your code as you type. When pasting raw code into the REPL such "intelligence" actually results in misformatted code that won't run. Hence my reason for switching to paste mode in the REPL. It's just how it works that the `CTRL-D` that gets you out of paste mode will also evaluate it. Not sure what to do - that's MicroPython rather than Mu.
* Actually, the drop into paste mode should only happen if we're pasting in multi-line fragments of text. **TODO** Fix this.
* I don't want to execute the code I paste into regular Python mode, because I might be pasting something I want to complete or add fragments to. Executing stops me from doing this.
* Yes, multi-line copy will always copy the whole line because that's how we a) work out indentation fixing, b) with multi-line you generally want ALL the lines, and c) Should you only want fragments, you can paste each one individually (and not executing pasted content in Python mode, means you can continue to paste until you're finished and press return).
* Yes, I need a story for debug, PyGameZero, web mode etc....

Is this clear..? I feel we're getting close, but we're not there yet. Your feedback ensures we release something that'll work well from the get go (and which we can refine given feedback).

ntoll:
@carlosperate I've fixed the MicroPython fragment problem I found and I've ensured that paste-to-repl only appears in a menu for the Python process pane if the Python code is being run with interactive mode (i.e. you drop into the Python shell after the process has completed). For all other (non-interactive) uses of the Python process pane, being able to paste into it doesn't make sense (i.e. web, pgzero and debugger modes).

I think we're nearly there..!

ZanderBrown:
I wonder if we want something like "Send to REPL"? "Copy selected text to REPL" feels a bit clunky

ntoll:
Well, it's a copy (not a send), and remember, we're dealing with folks who might be 11 years old, so it might not be obvious what's being sent or copied. ;-) (But I agree, it is clunky.... but I think clarity is also important too)

ntoll:
Thanks everyone for the feedback on this branch. I think it's in a state that could do with user testing, so will merge and we can refine from feedback on future branches. :+1:

PR #1572, opened 2021-05-17
Ensure, if no file extension provided, a .py one is used. 

### LINKED ISSUE #1571: Mu Editor 1.1.0b4 does not add the file extension .py when saving a new file for the first time on Linux
**Steps to recreate:**

1. Install the latest beta via pip in a virtual environment (Kubuntu 21.04 using a Python 3.8.10 virtual environment via pyenv).
2. Start the Mu Editor as normal.
3. Click on New to create a new tab.
4. Click on Save and in the resulting dialog box, enter a file name for the file, but do not add the .py extension and hit enter.

**Expected behaviour**

File saves with the .py extension added automatically.

**Actual behaviour**

File saves as a plain text file with no file extension.

This behaviour is different to that experienced on the Windows version which does add the file extension by default.

Thanks

ntoll:
Fixes #1571.

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -420,6 +420,12 @@ class Window(QMainWindow):
             "Python (*.py)",
         )
         self.previous_folder = os.path.dirname(path)
+        # Ensure there's a .py extension if none is provided by the user.
+        # See issue #1571.
+        name, ext = os.path.splitext(os.path.basename(path))
+        if not ext:
+            # No extension, so add .py as default.
+            path += ".py"
         logger.debug("Getting save path: {}".format(path))
         return path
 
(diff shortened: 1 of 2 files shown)


mu/interface/main.py
```diff
@@ -420,6 +420,12 @@ def get_save_path(self, folder):
             "Python (*.py)",
         )
         self.previous_folder = os.path.dirname(path)
+        # Ensure there's a .py extension if none is provided by the user.
+        # See issue #1571.
+        name, ext = os.path.splitext(os.path.basename(path))
```
carlosperate, line 425: Just a warning that this does not work with dot files, so something like `.gitignore` will be renamed to `.gitignore.py`. We can probably consider this out of scope, but I thought it was worth pointing out.

ntoll: Good point. For safety we should probably wrap the code in a conditional like `if not name.startswith("."):`

carlosperate: Yeah, sounds good 👍

ntoll: OK... fixed and tested in latest commit on this branch. Will merge once the CI turns green. :+1: Thanks for the heads-up @carlosperate

Issue #1556, opened 2021-05-10
Collection of FileExistsError and/or PermissionError crashes during venv creation


carlosperate:
Creating this issue to collect the cases where we encounter an error like this:

```
Traceback (most recent call last):
  File "C:\Users\idea\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 462, in quarantine_failed_venv
    os.rename(self.path, error_dirpath)
PermissionError: [WinError 5] Access is denied: 'C:\\Users\\idea\\AppData\\Local\\python\\mu\\mu_venv-38-20210509-222823' -> 'C:\\Users\\idea\\AppData\\Local\\python\\mu\\mu_venv-38-20210509-222823.FAILED'
```

In many cases this seems to be a "secondary" error, where something else has failed beforehand, and then quen trying to rename the venv we get a different error.

For this reason is better to keep these individual issues open, as the first problem might be different. There is a chance that they all have similar problems manifesting in different ways, but as that is not clear yet let's look at them individually.

carlosperate:
From https://github.com/mu-editor/mu/issues/1500:

> [mu.log](https://github.com/mu-editor/mu/files/6360982/mu.log)
> 
> Version: 1.1.0.beta.4
> Locale: en_GB
> Platform: Windows 10 10.0.19041 AMD64

First it errors when trying to upgrade pip, and then we a `FileExistsError` trying to rename the `venv` to `venv.FAILED'`:
```
2021-04-22 21:09:00,315 - mu.virtual_environment:379(run_subprocess) DEBUG: Process returned 101; output: Unable to create process using 'C:\Users\jason hart\AppData\Local\Programs\Mu Editor\Python\python.exe -m pip install --upgrade pip'
2021-04-22 21:09:00,315 - mu.virtual_environment:464(quarantine_failed_venv) ERROR: Unable to quarantine C:\Users\jason hart\AppData\Local\python\mu\mu_venv-38-20210422-210256 as C:\Users\jason hart\AppData\Local\python\mu\mu_venv-38-20210422-210256.FAILED
Traceback (most recent call last):
  File "C:\Users\jason hart\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 494, in ensure_and_create
    self.create()
  File "C:\Users\jason hart\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\virtual_environment.py", line 624, in create
[...]

Issue #1441, opened 2021-04-06
runninng mu-editor with pip version < 19 doesn't work in Linux


carlosperate:
Not 100% sure what feature that was added to pip v10 is required, but this minimum version is needed to be able to ~~install~~ download the `cryptography` package  (which has recently moved to Rust) on Linux, which is needed for `esptool`:

With pip v9, which is the default in Ubuntu 18.04:
```
$ /home/carlos/workspace/mu/venv/bin/python -m pip download --destination-directory '/home/carlos/workspace/mu/mu/wheels esptool==3.*
Collecting esptool==3.*
  File was already downloaded /home/carlos/workspace/mu/mu/wheels/esptool-3.0.tar.gz
Collecting bitstring>=3.1.6 (from esptool==3.*)
  File was already downloaded /home/carlos/workspace/mu/mu/wheels/bitstring-3.1.7.tar.gz
Collecting cryptography>=2.1.4 (from esptool==3.*)
  File was already downloaded /home/carlos/workspace/mu/mu/wheels/cryptography-3.4.7.tar.gz
    Complete output from command python setup.py egg_info:
    
            =============================DEBUG ASSISTANCE==========================

            If you are seeing an error here please try the following to
            successfully install cryptography:
    
            Upgrade to the latest pip and try again. This will fix errors for most
            users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
            =============================DEBUG ASSISTANCE==========================
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-bjn709ho/cryptography/setup.py", line 14, in <module>
        from setuptools_rust import RustExtension
    ModuleNotFoundError: No module named 'setuptools_rust'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-bjn709ho/cryptography/
```

With pip v10:

```
$/home/carlos/workspace/mu/venv/bin/python -m pip download --destination-directory '/home/carlos/workspace/mu/mu/wheels esptool==3.*
Collecting esptool==3.*
  File was already downloaded /home/carlos/workspace/mu/mu/wheels/esptool-3.0.tar.gz
Collecting bitstring>=3.1.6 (from esptool==3.*)
  File was already downloaded /home/carlos/workspace/mu/mu/wheels/bitstring-3.1.7.tar.gz
Collecting cryptography>=2.1.4 (from esptool==3.*)
[...]

PR #1573, opened 2021-05-17
Update virtualenv dependency to a version packing pip v19.3

carlosperate:
virtualenv vendors its own pip, and we need to ensure users have pip v19.3 or higher so that it's capable to install all dependencies in the user venv correctly.

The `cryptography` package used by `esptool` now uses Rust, so it needs pip v19.0. The next PyQt5 wheels requires `manylinux2010` support so that needs pip v19.3.

`virtualenv` v16.7.6 is the earliest version to pack v19.3 (the v19.1 included is only required for Python 3.4, which Mu does not support): https://github.com/gaborbernat/virtualenv/tree/16.7.6/virtualenv_support

More info in these GH issues:
https://github.com/mu-editor/mu/issues/1441#issuecomment-814790143
https://github.com/mu-editor/mu/issues/1556#issuecomment-842503883

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ install_requires = [
     #
     # Needed for creating the runtime virtual environment
     #
-    "virtualenv>=16.0.0",
+    "virtualenv>=16.7.6",
     #
     # Needed for packaging
     #


tjguk:
Thanks for the analysis and the PR @carlosperate -- please go ahead and merge. I've already removed the "upgrade pip" process from the virtualenv in #1530 so hopefully the combination of this and that should avoid those "can't upgrade pip" issues we've been getting.

PR #1578, opened 2021-05-19
CI: Add Arm Raspbian test runner

carlosperate:
Originally looked into this way back during the Christmas break, finally got around to put it together for Mu.

It basically utilises the awesome [dockerpi](https://github.com/lukechilds/dockerpi) project to create a docker image with Qemu emulating armv6 and running Rasbian/RaspberryPi-OS. Kudos to all the hard work done over there to get this running.

I simply had to create some custom Raspbian images to enable SSH access and install the Debian packages needed (PyQt5 et all): https://github.com/carlosperate/rpi-os-custom-image

And then created the new docker images to wrap it all up: https://github.com/carlosperate/docker-qemu-rpi-os

This PR updates the GitHub Actions workflow to create a new job with these docker images.
Unfortunately because of the way the docker is run in the GH Actions platform, we cannot bash directly into Raspbian as one would do when running these images locally, so I had to set it up as a "service" and SSH into it on each test step.

To start this adds 2 jobs, both on the same Raspbian Stretch image. One installs Mu in a virtual environment, the other on the system Python. The majority of the crash logs we get from Raspbian have been pip installed in the system Python, so it'll be good to test that as well (mostly how it might or might not clash with the Debian system Python packages).

Once I have pushed more docker images with Buster I'll add those here as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -109,8 +109,8 @@ jobs:
         uses: appleboy/ssh-action@master
         with:
           host: rpios
-          username: ${{ secrets.PI_USERNAME }}
-          password: ${{ secrets.PI_PASSWORD }}
+          username: pi
+          password: raspberry
           port:  ${{ job.services.rpios.ports[5022] }}
           script: |
             mkdir ~/mu && cd ~/mu
@@ -124,8 +124,8 @@ jobs:
         uses: appleboy/ssh-action@master
         with:
           host: rpios
-          username: ${{ secrets.PI_USERNAME }}
-          password: ${{ secrets.PI_PASSWORD }}
+          username: pi
+          password: raspberry
           port:  ${{ job.services.rpios.ports[5022] }}
           # Purposely skip apt update, install version from image stale index
           script: sudo apt-get install -y python3-virtualenv
@@ -133,8 +133,8 @@ jobs:
         uses: appleboy/ssh-action@master
         with:
           host: rpios
-          username: ${{ secrets.PI_USERNAME }}
-          password: ${{ secrets.PI_PASSWORD }}
+          username: pi
+          password: raspberry
           port:  ${{ job.services.rpios.ports[5022] }}
           command_timeout: 20m
           script: |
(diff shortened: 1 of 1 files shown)


carlosperate:
Separated the "install in system Python" test environment as that one has some issues and I don't want to block this PR, so I'll add those changes in a different one.

Once the tests pass this should be ready to be reviewed and merged.

carlosperate:
Added the a test runner with the latest Buster image. Should be good to go once all the test finish.

.github/workflows/test.yml
```diff
@@ -88,3 +88,81 @@ jobs:
...
+      fail-fast: false
+    services:
+      rpios:
+        # Custom made images for Mu: https://github.com/carlosperate/docker-qemu-rpi-os
+        image: ghcr.io/carlosperate/qemu-rpi-os-lite:${{ matrix.docker-tag }}-mu
+        ports: ["5022:5022"]
+    steps:
+      # This delay is a bit hacky, but can't find a way to signal when OS is ready
+      - name: Wait 2m30s for the docker image to start up QEMU and Raspberry Pi OS
+        run: sleep 150
```
carlosperate, line 107: This step is a bit hacky, but will look into it in the next revision (when adding tests for `pip install mu` in the system Python).

.github/workflows/test.yml
```diff
@@ -88,3 +88,81 @@ jobs:
...
+          host: rpios
+          username: ${{ secrets.PI_USERNAME }}
+          password: ${{ secrets.PI_PASSWORD }}
+          port:  ${{ job.services.rpios.ports[5022] }}
+          script: |
+            mkdir ~/mu && cd ~/mu
+            git init
+            git remote add origin ${{ github.server_url }}/${{ github.repository }}.git
+            git fetch --progress --depth=1 origin ${{ github.sha }}
+            git checkout --progress FETCH_HEAD
```
carlosperate, line 120: We cannot use the standard "git checkout" GH Action step, so we have to figure out what to checkout. I've based this on the official Action step source code, but I'm not 100% sure if it'll work, so I've created this PR from my fork to check it: https://github.com/mu-editor/mu/pull/1593/

carlosperate: PR #1593 failed correctly, so this part is correct.

Basically GitHub creates a detached merge commit in this repository and it passes this hash to the CI, so these steps don't need to add the fork as a remote.

carlosperate:
This is good to go, I'll merge it by the end of the day if nobody has any comments (or let me know if anybody wants a bit more time to review it).

PR #1580, opened 2021-05-19
Update adafruit-board-toolkit to ~=1.1 to fix M1 MacOS issues

### LINKED ISSUE #1543: M1 Mac issue with adafruit-board-toolkit
We are seeing an issue that MacOS on M1 Macs is not returning the USB CDC interface names properly to `adafruit_board_toolkit`, so it can't detect that a serial port is on a CircuitPython.

It works fine on non-M1 Macs up through MacOS 11.3.1, but is not working on M1 Macs on 11.3.1 and earlier ones. We're attempting to do more testing, but I don't personally have an M1 Mac and neither does @Neradoc, who wrote the code to properly get the interface names.

This would explain some issues being reported around CircuitPython boards not being detected on MacOS in the beta.,

dhalbert:
Updates the version specifier for `adafruit-board-toolkit` so the minimum version is `1.1.0.` Any 1.x version will be upward compatible and be OK.

`adafruit-board-toolkit` 1.1.0 has two important fixes:
- MacOS interface name fetching is improved so it works back to Sierra and forward to MacOS Big Sur on M1. There are differences between M1 MacOS and non-M1. See https://github.com/adafruit/Adafruit_Board_Toolkit/pull/4.
- The interface names don't have to start with "CircuitPython"; there are some boards that uses other prefixes. See https://github.com/adafruit/Adafruit_Board_Toolkit/pull/2.

Fixes #1543.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ install_requires = [
     #
     # adafruit-board-toolkit is used to find serial ports and help identify
     # CircuitPython boards in the CircuitPython mode.
-    "adafruit-board-toolkit>=1.0.1",
+    "adafruit-board-toolkit~=1.1",
     "pyserial~=3.5",
     "nudatus>=0.0.3",
     # `flake8` is actually a testing/packaging dependency that, among other


dhalbert:
Tested with a Trinket M0 running CircuitPython 7.0.0, and set up to present two serial interfaces (REPL and data). Tested on High Sierra, Intel Big Sur, and M1 Big Sur (all up to date as of today). In all cases, the board is recognized and the Serial button goes to the correct `/dev/cu.*`.

ntoll:
Bravo @dhalbert and many thanks for the contribution - it was on my todo to check the status of this effort and you've pre-empted me and saved me some time. So thank you..! :+1:

PR #1347, opened 2021-03-04
Update bug report template to make log collapsible

ajs256:
The log file can get quite long. I edited the issue template to have the reporter put the log file in a `<details>` tag, which makes it collapsible.

Example of the details tag in action below:

<details>
<summary>Summary of hidden content</summary>
Hidden text
</details>

Here's how it's used:
```html
<details>
<summary>Summary of hidden content</summary>
Hidden text
</details>
```

PR #1582, opened 2021-05-22
Make the issue template more instructive

ajs256:
I added some instructions to the issue template. These:
* Make it seem more like a form, and easier for someone to fill out
* Put the logs in a `<details>` tag, making it collapsible. (as done in #1347)


ntoll:
Thanks..!

PR #1585, opened 2021-05-22
Support TinyS2 ESP32-S2 Board

rgreer4:
Adds support for the new TinyS2 ESP32-S2 board.  Appears to have a different set of VID/PID values than what is already in Mu.

### DIFF
--- a/mu/modes/esp.py
+++ b/mu/modes/esp.py
@@ -52,6 +52,7 @@ class ESPMode(MicroPythonMode):
         (0x1A86, 0x7523, None, "HL-340"),
         (0x10C4, 0xEA60, None, "CP210x"),
         (0x0403, 0x6001, "M5STACK Inc.", "M5Stack ESP32 device"),
+        (0x303A, 0x8002, "UnexpectedMaker", None),  # TinyS2 (ESP32-S2)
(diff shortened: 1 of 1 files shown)


rgreer4:
(Note I should have mentioned this is supporting a direct CP install without the UF2 bootloader, hence the difference in VID/PID I think.  In case ya'll dont want to bother with a variant like this.)

Neradoc:
You mentioned CP, but you modified the file for the Micropython mode on ESP, is that what you wanted to do ?
Circuitpython mode no longer needs a list of vid/pid in the current beta, using the Serial connection interface name instead.

FYI this vid/pid is from https://github.com/espressif/usb-pids where UnexpectedMaker registered a pid for Circuitpython `0x8002` and one for the UF2 bootloader `0x8005`, as is usually preferred. The presence of the UF2 bootloader does not influence the PID of Circuitpython.

ntoll:
:wave: OK.... if this is a change for CP, it should probably be done in the context of the CP mode rather than ESP. Happy to follow steer from folks more "in the know" about where this should happen. :+1:

Thanks for the contribution.

ntoll:
Any movement on this..?

Issue #1312, opened 2021-02-20
SVG-versions of Mu-buttons and icons


dybber:
Currently Mu's buttons and icons are .PNG images, and thus not very easy to change and customize. It has been requested in #970 that we make new versions of the buttons for the "dark mode" and the "high contrast" modes, but I believe the best way to create those buttons, would be to first make SVG-versions of the existing buttons, and then experiment with the colors to create the dark-mode buttons.

The current Mu-buttons that will need to be recreated are available at: https://github.com/mu-editor/mu/tree/master/mu/resources/images

When creating the icons it can be useful to use these vector icons as a starting point:
https://github.com/mu-editor/Mu-tate/tree/master/mu/icons

I think we want to go for an almost exact identical look and feel as the current Mu icons and buttons.

ntoll:
OK... so the markup/CSS in "Mu-tate" should help with working out how to display the buttons as SVG. I may take a look today if I find the time.

ntoll:
OK... Just made a start by changing the images to SVG. Here they are pre-styling... ;-)

![mu_svg](https://user-images.githubusercontent.com/37602/108593629-b1b4a400-736c-11eb-8aae-45472b9142e6.png)

ntoll:
So, this is FAR more complicated than I at first thought.

The buttons in the `QToolBar` area are defined as `QAction` instances (as per idiomatic Qt). However, `QAction` instances require a `QIcon` instance to define the image. It turns out `QIcon` instances cannot be styled by CSS - hence no progress further than the black SVGs.

Unless we completely re-write how we do the button bar and move away from `QToolBar` / `QAction` (which feels like a LOT of work), I don't think using SVG for buttons will be possible.

I'm going to close this as a "won't fix" due to the problems outlined above. :-/

Please feel free to re-open if you're able to think of a quick/simple fix.

dybber:
I wasn't thinking that we should do that. No CSS is necessary for what I propose. I was thinking that for every PNG file, we should get someone to create a matching SVG file.

For example, we have the Run button styled in this way: https://github.com/mu-editor/mu/blob/master/mu/resources/images/run.png
I'd like to have that very same image, including the white background and orange border in SVG-format. 

That should be doable. Then when we have those SVG icons/buttons, it will be easy to make other SVG-versions for other color schemes.

dybber:
[...]

PR #1603, opened 2021-05-28
Added SVG versions of all circular button images in mu/resources/images

Rybec:
These are the SVG icons requested in issue #1312.

First note that I am not a graphic designer.  I merely have significant experience with Inkscape, and this seemed like something fairly easy to do.  (I mention this, because I don't have the time, motivation, or skill to work on complex graphics.)

Second, there are some potential stylistic issues with these.  Specifically, on some of the icons, I have left parts transparent that others might prefer to be white.  Things that should obviously not be transparent, like cuff buttons on the check buttons, are white, but places like the center holes in the gears of the snippets button and the middle of the sun on the theme_day button I've left transparent.  If there are places where the transparency is a problem, please let me know, and I can add background when I have time.  Beyond this, I did my best to remain faithful to the original images.  If minor adjustments are needed, like adding background to transparent parts, let me know, and I will work on it as I have time.  Changing colors on existing elements should be trivial for anyone to do.  If the SVGs need to be rendered to PNGs, also let know.  It has been a while, but I've done batch rendering with Inkscape before, and it's way faster than trying to render 37 images manually, one-by-one.  I can also make new SVG buttons, given the png files for them to use as a template.

I don't know if this counts as a contribution big enough to add to the AUTHORS file, but if it is, I can be included as "Ben Williams" or "Rybec Arethdar".

### DIFF
--- a/mu/resources/images/svg/browse.svg
+++ b/mu/resources/images/svg/browse.svg
@@ -81,14 +81,7 @@
      inkscape:label="Layer 1"
      inkscape:groupmode="layer"
      id="layer1"
-     transform="translate(0,-237.38055)">
-    <ellipse
-       style="fill:none;stroke:#ffcc33;stroke-width:4.34999;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
-       id="path871"
-       cx="29.809498"
-       cy="266.46942"
-       rx="2.4063713e-13"
-       ry="0.72084588" />
+     transform="translate(0,-237.38055)"> -->
     <g
        id="g895">
       <path
--- a/mu/resources/images/svg/browse.svg
+++ b/mu/resources/images/svg/browse.svg
@@ -81,7 +81,7 @@
      inkscape:label="Layer 1"
      inkscape:groupmode="layer"
      id="layer1"
-     transform="translate(0,-237.38055)"> -->
+     transform="translate(0,-237.38055)">
     <g
        id="g895">
       <path
--- a/mu/resources/images/svg/browse.svg
+++ b/mu/resources/images/svg/browse.svg
@@ -1,0 +1,121 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="59.619438mm"
+   height="59.619438mm"
+   viewBox="0 0 59.619438 59.619438"
+   version="1.1"
+   id="svg8"
+   inkscape:version="1.0.2-2 (e86c870879, 2021-01-15)"
+   sodipodi:docname="browse.svg">
+  <defs
+     id="defs2" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="5.0271529"
+     inkscape:cx="125.9087"
+     inkscape:cy="124.07886"
+     inkscape:document-units="mm"
+     inkscape:current-layer="layer1"
+     inkscape:document-rotation="0"
+     showgrid="false"
+     inkscape:snap-page="true"
+     fit-margin-top="0"
+     fit-margin-left="0"
+     fit-margin-right="0"
+     fit-margin-bottom="0"
+     showguides="true"
+     inkscape:guide-bbox="true"
+     inkscape:window-width="1920"
+     inkscape:window-height="1001"
+     inkscape:window-x="-9"
+     inkscape:window-y="-9"
+     inkscape:window-maximized="1"
+     inkscape:snap-to-guides="true">
+    <sodipodi:guide
+       position="29.809499,69.619441"
+       orientation="-1,0"
+       id="guide24"
+       inkscape:label=""
+       inkscape:locked="false"
+       inkscape:color="rgb(0,0,255)" />
+    <sodipodi:guide
+       position="-9.9999998,29.809721"
+       orientation="0,1"
+       id="guide26"
+       inkscape:label=""
+       inkscape:locked="false"
+       inkscape:color="rgb(0,0,255)" />
+    <sodipodi:guide
+       position="-19.894461,69.619437"
+       orientation="0,1"
+       id="guide879"
+       inkscape:label=""
+       inkscape:locked="false"
+       inkscape:color="rgb(0,0,255)" />
+  </sodipodi:namedview>
+  <metadata
+     id="metadata5">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title />
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <g
+     inkscape:label="Layer 1"
+     inkscape:groupmode="layer"
+     id="layer1"
+     transform="translate(0,-237.38055)">
+    <ellipse
+       style="fill:none;stroke:#ffcc33;stroke-width:4.34999;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+       id="path871"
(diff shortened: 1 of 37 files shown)


ZanderBrown:
Nice! I can see some of these still need some work (i.e some sizing feels off, random yellow blobs) but a really great start

Rybec:
I wasn't sure about the random yellow blobs, because they don't show in in Inkscape.  They only showed up when I installed something to give me thumbnails in Windows Explorer, so I wasn't sure what the source was.

I'll take a look when I have some time and see what I can do.  If necessary, I can edit the XML directly to get rid of things that don't belong.  And honestly, maybe I should do that, because I can clean up some of the cruft Inkscape leaves behind while I am at it.  (Nothing against Inkscape, but I do wish it had an option for cleaning out the unnecessary metadata that it keeps in the files.)

The sizing is based on the sizes of the original buttons.  I basically just imported them at their normal scale and used them as templates.  Note that if you are viewing them with any Adobe products, the sizing _will_ be off, because Adobe is not compliant with the SVG standard.  One of the compliance issues is with scaling.

Rybec:
Ok, so I just looked at what all of the files look like in LaserGRBL (a laser cutting application that can process SVG paths into laser cutting paths, which gives me a view I can't get with Inkscape).  At some point, all of the images have a path that goes to the center, even though nothing is there (or at least, in Inkscape, nothing appears to be there).

I am thinking maybe the best solution would be to make the main yellow/orange circle path into a full object.  Right now, it is just a circle object with a wide line width.  Since this is the only object that is the same color as the blob in the middle, I suspect it is the source.  I'll see if this fixes that!

Rybec:
Alright, the yellow line in the middle is gone.  I did end up having to edit the XML directly.  There was an infinitesimally small ellipse right there, but for some reason, Inkscape wasn't even picking it up.  (It was also causing some strange background effects in some viewers that went away when I removed the stray ellipse.)  It was in every image except button.svg (which is just the circle with no icon).  There was also an issue with two of the arcs rendering in sounds.svg, which I fixed.  I did not convert the outside circle stroke to paths, because if the buttons need an opaque background, it is trivial to add this by adding a fill to the outer circle.  Once converted to paths though, this will no longer work.  If it becomes necessary to convert them, there are other options for adding background, but so long as it isn't necessary, it is better the way it currently is.

Another note on the size issue: SVGs are highly scalable.  Any rendering library should be able to scale them.  So if the rendering code displaying the current buttons doesn't display the SVGs at the same scale, it should be trivial to scale them to the right size during rendering.

ZanderBrown:
Ah yes, out of context the ? looked overly small but your right - it matches the original

Rybec:
There are a few where the size doesn't quite look right, because the edges of the original are fuzzy, so there is no definite place to start the edges.  I had to work out where the hard edges of the SVG elements needed to go.  Before moving on to the next one though, for each one I zoomed out to approximately the scale used in the application and compared.  If I could see any noticeable differences at that scale, then I made adjustments to remove them.  At significantly larger scales, some differences can be seen, but that's mainly because of going from the soft, pixelated edges of the original images to the hard, smooth edges of the SVGs.

Anyhow, thanks for checking my work.  Now I know that if I am doing work like this for art, I should double check it in a browser, since browsers will display tiny artifacts that are not visible in Inkscape.  (The vast majority of my experience with Inkscape is designing things for laser cutting or for importing into 3D modeling software to extrude for 3D printing.)

ntoll:
@Rybec wow.... this is wonderful work. Many thanks for your contribution. I've just had a look at the svg files and notice only one slight fly-in-the-ointment: some of the SVG files have whitespace as white instead of transparent (see, for example, `static.svg` or the eyes in the snake in `modes.svg`). But I'm really scraping the bottom of the barrel here - completely blown away that we have SVG files now. Please do let me know when you feel they're in a good state and I'll make the necessary code changes to switch the buttons to using them instead of the png files we currently use.

Thank you..!

One final question, while I remember: what process would you suggest for when we need to create new buttons. Are there some simple instructions we can add to our documentation so folks know what to do in the future if they want to contribute something that needs a new button..?

Rybec:
I left some white space in intentionally, where it made sense to me.  The full list of buttons that have white left in are:

The cuff button in check.svg, check-good.svg, and check-bad.svg are white.
The text ".css" in css.svg is white.
The Arduino symbol in files.svg is white.
The two dots in load.svg and save.svg are white.
The snake eye in modes.svg is white.
The controller buttons in play.svg are white.
static.png had a slightly off white background square, which I retained in pure white in static.svg.
The text in templates.svg is white.

I can make any of those white sections transparent, if you would like.  I think most of those are more appropriate as white rather than transparent, but static.svg is one that I wasn't sure about.  As mentioned above, the original also has that white square background in a slightly off white color.  My assumption was that it was intended to represent a photo or image background.  Honestly, I don't actually like how that one looks.  So let me present two options.  The first is, it would take me only a few seconds to completely remove the white background box.  The second is, I _could_ give the two image shapes within the button fitted white backgrounds and remove the large white background.

As far as modes.svg goes, having the eye in white makes more sense to me, but if you want it transparent instead, it shouldn't take more than a minute or so for me to change that.

Let me know what you want with those, and I'll do it!



I'll put the new button guidelines in a separate comment.

Rybec:
Now, for new buttons.  (Note my comment on fonts in the third paragraph below.  It might be a good idea to set an official (open source) font for buttons using the folder base (and maybe for all buttons aside from fonts.svg).  If you don't know what the original font was, I can help find a similar one, and I can change the existing buttons to use that font.)  Anyhow, let me know what you think of these guidelines.  I tried to keep them simple, though some of the stylistic effects can be challenging to achieve.  I didn't want to turn this into an Inkscape tutorial!


Use Inkscape if you can, because Adobe doesn't follow the SVG standard when it comes to scale.  (Adobe will _probably_ still work though, since this isn't a physical application...)

Start with button.svg.  That will provide the correct "page" size, the correct positioning on the page, and the circular border.

Buttons related to files can start with a copy of files.svg.  Remove the Arduino symbol and put whatever you need on top of the folder icon in white.  This will maintain consistency in shape, position, and scale between all buttons that use the folder icon.  (Note that I don't know what font was used for the original .png buttons, so instead of text, I reproduced the text in css.svg and templates.svg using vector curves.  This is not ideal, and it would be very difficult to produce the same font style using this technique, without reference text in the original font.  If anyone knows the original font, please let me know, and I can change those to use real text in that font.  That would allow mean we can add that font to the guidelines for new buttons.  Note that fonts do have copyrights.  If the original font isn't under an open source license, we might want to choose a new font to use for these, that is open source.  Let me know if you want to do this, as I have some experience finding open source fonts that meet particular needs.)

If you need an element that already exists in another button, either start with that button and remove unnecessary elements, or open that button, copy that element from there, and paste it into your new button.  This will maintain consistency between buttons that use related elements, and if you need to scale an element, it will maintain the same proportions as that element in other buttons.

Colors can be picked from any of the other buttons.  The primary dark teal-ish color is #336699.  If red or green are needed for "good" or "bad" or "success" or "failure" type buttons/icons, use the color picker tool to get the color you need from check-good.svg or check-bad.svg.

Try to maintain stylistic choices when making new buttons:

If you have overlapping elements that are not part of one composite element, take a look at browse.svg and load.svg, and note how the arrow elements have a transparent border around them that cuts into the element beneath.  (This effect can be challenging to achieve, but it is possible by grouping the element that needs cut, and then using Path->Difference to make the cut.  To make this work correctly, the element needing cut needs to be below the element doing the cutting.  So for browse.svg, you would select the world element (which is a circle and some curved lines grouped into a single element), and press End (or Object->Lower to Bottom, in the menus).  Then you would select it and an arrow where the stroke covers everything that needs cut out.  Then use Path->Difference.  Note that this will delete the cutting element, the arrow, in this case.  If you need to retain it, use Edit->Duplicate to create an in-place copy before doing the Path-Difference operation.)

Most elements with corners in the original button set have those corners rounded.  The rounding mostly ranges from around 1mm to 3mm corner radius.  To maintain consistent style, compare your button elements to similarly sized elements in existing buttons and use a similar rounding radius on corners on that element.


If you don't know how to get the effect you want in Inkscape, a quick internet search will generally turn up good information.  Inkscape also has a fairly good community.  Ask questions in [Inkscape's forums](https://inkscape.org/forums/), which has a beginners section that might be worth searching before asking, or the [Graphics Design StackExchange](https://graphicdesign.stackexchange.com/) looks like a good place to ask questions as well (but make sure to tag your post [inkscape] if your question is Inkscape specific).

ntoll:
Hi @Rybec @ZanderBrown - just finding some time for GitHub gardening of PRs etc, and wondering on the status of this wonderful work..?

Rybec:
@ntoll: See my second to last comment.  The last comment I made above is a process for making new buttons, which you asked for.  The comment before it is the most recent status report as well as some explanation of why I did things the way I did.

Personally, I think they look good, but if you want any changes, let me know, and I will see what I can do.

keith-packard:
These look awesome! I've got an add-on pull-request that gets these working in mu, #1814

Rybec:
That's awesome!  Thank you.  It feels good to see my work here going somewhere.

keith-packard:
This merge request should be closed -- it was merged as a part of #1814

carlosperate:
Awesome, thank you so much @Rybec and @keith-packard!

Issue #1600, opened 2021-05-27
venv pip install step does not install all wheels


carlosperate:
## Issue description
We've had a series of Crash reports where it looks like `pip install ...` step for wheels finishes successfully, but then the `pip list` only shows a partial list of packages.

These problems are usually manifested with the `ModuleNotFoundError: No module named 'pgzero'` (or a different module), as it might have installed some of the wheels, but not reached `pgzero`.

## Example log
For example in https://github.com/mu-editor/mu/issues/1591 we find:

```
2021-05-24 14:48:24,370 - mu.virtual_environment:193(run) DEBUG: Process output: Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\adafruit_board_toolkit-1.0.1-py3-none-any.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\backcall-0.2.0-py2.py3-none-any.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\bitstring-3.1.7-py3-none-any.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\cffi-1.14.5-cp38-cp38-win_amd64.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\click-7.1.2-py2.py3-none-any.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\colorama-0.4.4-py2.py3-none-any.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\cryptography-3.4.7-cp36-abi3-win_amd64.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\decorator-5.0.7-py3-none-any.whl
Processing c:\users\zachary.desktop-96jpn9t\appdata\local\programs\mu editor\python\lib\site-packages\mu\wheels\ecdsa-0.16.1-py2.py3-none-any.whl
[...]

PR #1610, opened 2021-05-31
venv: Install each wheel individually. 

carlosperate:
This also provides more granular logs and decreases the chance of having a timeout of the pip install process. 

Before this the 3 minutes configured for the timeout might not have been long enough to install all the wheels together, so with this PR we have 3 min per wheel which should be enough for individual install.

Should hopefully fix https://github.com/mu-editor/mu/issues/1600

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -264,6 +264,7 @@ def run():
     setup_logging()
     logging.info("\n\n-----------------\n\nStarting Mu {}".format(__version__))
     logging.info(platform.uname())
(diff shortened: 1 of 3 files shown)


carlosperate:
Looks like mock.call_args_list has changed behaviour in python 3.8 (the version I am using) so tests in earlier versions of Python are failing. I'll look into it.

carlosperate:
Reviewed by @tjguk in https://github.com/mu-editor/mu/issues/1600#issuecomment-851641660, tests are now passing, so merging.

PR #1622, opened 2021-06-06
Improve logging & test coverage for virtual_environment

tjguk:
I realised that #1530 was sprawling somewhat so I've started to pull it apart to a series of more focused PRs. This is the first, which attempts to improve the logging from subprocesses created by the `virtual_environment` module and its objects.

I've also filled many of the gaps in the test coverage for this module -- the ones which remain are all to do with Qt slots, and I haven't yet decided on the best way to approach those.

The only change I've made outside the virtual_environment module and its tests, is to centrally limit the output to 1800 chars which will form part of the crash handler URL. I was having to do that anyway in several places and it made sense to put it at the final point where it's needed.

Although we usually debug-log the full output from subprocesses, I've added the slightly compact version into a number of exception messages so we should see it much sooner via a crash report than having to wait until logs arrive. (Or guess, when they don't).

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -175,7 +175,7 @@ def excepthook(*exc_args):
         log_file = base64.standard_b64encode(LOG_FILE.encode("utf-8"))
         error = base64.standard_b64encode(
             "".join(traceback.format_exception(*exc_args)).encode("utf-8")
-        )[:1800]
+        )[-1800:]
         p = platform.uname()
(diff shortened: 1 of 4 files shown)


tjguk:
Assuming that this PR doesn't crash and burn in the test phase, or as a result of anyone's review comments, the further PRs will branch off this rather than from master.

PR #1624, opened 2021-06-06
Detect change of mu version

tjguk:
Detect a change of mu and recreate the user's virtual environment


tjguk:
NB At present two tests are consistently failing on my dev setup after removing the existing venv. Haven't had a chance to find out why yet, but they're obviously relying on the presence of something in the venv setup being already in place:

`test_venv_folder_already_exists`
`test_venv_second_try`

tjguk:
Ok; I propose to merge this PR (which will effectively also merge #1622 which it's built on top of). There may be some rework needed in the light of any comments, but I'd rather not let it grow stale.

PR #1644, opened 2021-06-15
Update deprecated `logging.warn()` to `warning()`.

carlosperate:
`logging.warn()` is deprecated, so this is just a simple PR to update all instances to the equivalent `warning()`  version: https://docs.python.org/3.9/library/logging.html#logging.warning


ntoll:
Thank you @carlosperate :+1:

Issue #1683, opened 2021-07-05
Debugpy (new ipykerner v6 dep) is not macOS 10.13 High Sierra compatible


paco218:
Version: 1.1.0.beta.5
Locale: es_MX
Platform: Darwin 17.7.0 Darwin Kernel Version 17.7.0: Mon Aug 31 22:11:23 PDT 2020; root:xnu-4570.71.82.6~1/RELEASE_X86_64 x86_64

```
ate
    self.install_baseline_packages()
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 880, in install_baseline_packages
    self.install_from_zipped_wheels(zipped_wheels_filepath)
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 855, in install_from_zipped_wheels
    self.pip.install(wheel, deps=False, index=False)
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 270, in install
    return self.run(
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 246, in run
    result = self.process.run_blocking(
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 122, in run_blocking
    return self.wait(wait_for_s=wait_for_s)
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 186, in wait
    raise VirtualEnvironmentError(
mu.virtual_environment.VirtualEnvironmentError: Process finished but with error code 1:
ERROR: debugpy-1.3.0-cp38-cp38-macosx_10_14_x86_64.whl is not a supported wheel on this platform.

```

carlosperate:
Thanks for the report @paco218!
Would you be able to attach the Mu log to this GitHub Issue as well? It'll be really useful to us for debugging. File should be located in: `~/Library/Logs/mu`

What version of macOS are you using?

carlosperate:
Okay, so based on the Darwin version it looks like you are in 10.13 High Sierra.

I believe the problem is from the recent debug capabilities added to `ipykerner` v6 released less than a week ago on June 30th:
[...]

PR #1686, opened 2021-07-06
Set ipykernel version to be < 6.0 for macOS 10.13 compatibility

carlosperate:
This is a `qtconsole` dependency, as `qtconsole` only declares the min version, Mu gets the latest available during packaging, downloading v6 released less than a week ago.
`ipykernel` v6 adds `debugpy` as a dependency, which is not compatible with macOS 10.13 High Sierra.
This dependency is only needed for the debugging feature added to iPython in v6, so it's fine go back and use an older version.

As the iPython dependencies are also installed in the user venv  we have to add the same to `mu/wheels/__init__.py`.

After the Mu v1.1 release we'll have to update PyQt5 to a newer version that will drop support for macOS 10.13, so at that point we can unpin `ipykernel`.

Fixes https://github.com/mu-editor/mu/issues/1683.

### DIFF
--- a/mu/wheels/__init__.py
+++ b/mu/wheels/__init__.py
@@ -40,6 +40,10 @@ ZIP_FILEPATH = os.path.join(WHEELS_DIRPATH, mu_version + ".zip")
 mode_packages = [
     ("pgzero", "pgzero>=1.2.1"),
     ("flask", "flask==1.1.2"),
+    # FIXME: ipykernel max ver added for macOS 10.13 compatibility, min taken
+    # from qtconsole 4.7.7. Full line can be removed after Mu v1.1 release.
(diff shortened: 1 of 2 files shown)


ntoll:
LGTM

carlosperate:
Turns out this is also required for compatibility with Python 3.5 and 3.6, since `ipython>=7.21.0` needs Python 3.7+.

`ipython>=7.21.0` is required by `ipykernel>=6.0.0a0`, fetched by `qtconsole==4.7.7` not having a max version:
-  https://github.com/ipython/ipykernel/blob/6.0.0a0/setup.py#L93
- https://github.com/jupyter/qtconsole/blob/4.7.7/setup.py#L78

PR #1747, opened 2021-08-10
Overload insertFromMimeData instead of paste in REPL window

keith-packard:
Qt defines a lower-level function, insertFromMimeData, that handles
both pasting from the clipboard and Linux middle button pasting. Use
that instead of just overloading 'paste' so that whatever the user
does gets handled in the right way.

### DIFF
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -44,7 +44,6 @@ from PyQt5.QtWidgets import (
     QGridLayout,
     QLabel,
     QMenu,
-    QApplication,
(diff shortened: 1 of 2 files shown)


ntoll:
@keith-packard :wave: Apologies for the delay - we're all volunteers and I'm only just finding some time to do GitHub "gardening". As a Linux user who regularly uses the middle-button copy/paste feature, this works a treat. Many thanks for taking the time to fix this - I wasn't aware of the Qt `insertFromMimeData` function.

PR #1761, opened 2021-08-20
Added logic to function under Crostini on ChromeOS and return correct CIRCUITPY path if it exists

### LINKED ISSUE #1758: Can't use `mount` to find path under Crostini (Linux Apps) on ChromeOS
The `mount` command inside this VM/container doesn't show any external storage devices, but there are other ways to see those files.

```
9p on /mnt/chromeos type 9p (rw,nosuid,nodev,relatime,sync,dirsync,access=any,trans=fd,rfd=26,wfd=26)
tmpfs on /mnt/external type tmpfs (rw,nosuid,nodev,noexec,relatime,mode=755)
```

On ChromeOS you can "Share with Linux" from a directory (whether on the Chromebook or a USB or SD card) and this uses a special 9p client/server model to access the files in a secure way, which is why it doesn't show a specific external device above even though I have mapped the CIRCUITPY USB drive using the "Share with Linux".

The lovely thing about the way the ChromeOS sharing is designed is that it uses a predictable path, so we can say with strong confidence that the device will show up at `/mnt/chromeos/removable/CIRCUITPY` because it uses the volume label as the last segment of the path.
image
image

```
$ ls /mnt/chromeos/removable/
CIRCUITPY
```

https://github.com/mu-editor/mu/blob/2b20da24e5219002fe19781509fa8665a9842464/mu/modes/circuitpython.py#L170

The potential fix (I'll pull request shortly for review).

```
            if os.path.exists('/mnt/chromeos'):
                # We're on ChromeOS
                if os.path.exists('/mnt/chromeos/removable/CIRCUITPY/'):
                    device_dir = '/mnt/chromeos/removable/CIRCUITPY/'
                else:
                    self.view.show_message('If your Circuit Python device is plugged in, you need to "Share with Linux" on the CIRCUITPY drive in the "Files" app then restart Mu.')
```

espoelstra:
Fixes #1758 

May add similar functionality for micro:bit since it uses a flash drive based deployment mechanism as well.

### DIFF
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -200,6 +200,12 @@ class CircuitPythonMode(MicroPythonMode):
                             repr(e), mount_command
                         )
                     )
+            if os.path.exists('/mnt/chromeos'):
+                # We're on ChromeOS
+                if os.path.exists('/mnt/chromeos/removable/CIRCUITPY/'):
+                    device_dir = '/mnt/chromeos/removable/CIRCUITPY/'
+                else:
+                    self.view.show_message('If your Circuit Python device is plugged in, you need to "Share with Linux" on the CIRCUITPY drive in the "Files" app then restart Mu.')
 
         elif os.name == "nt":
             # We're on Windows.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -205,9 +205,12 @@ class CircuitPythonMode(MicroPythonMode):
                 if os.path.exists("/mnt/chromeos/removable/CIRCUITPY/"):
                     device_dir = "/mnt/chromeos/removable/CIRCUITPY/"
                 else:
-                    self.view.show_message(
-                        'If your Circuit Python device is plugged in, you need to "Share with Linux" on the CIRCUITPY drive in the "Files" app then restart Mu.'
+                    m = _(
+                        "If your Circuit Python device is plugged in,"
+                        + ' you need to "Share with Linux" on the CIRCUITPY drive'
+                        + ' in the "Files" app then restart Mu.'
                     )
+                    self.view.show_message(m)
 
         elif os.name == "nt":
             # We're on Windows.
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -200,12 +200,14 @@ class CircuitPythonMode(MicroPythonMode):
                             repr(e), mount_command
                         )
                     )
-            if os.path.exists('/mnt/chromeos'):
+            if os.path.exists("/mnt/chromeos"):
                 # We're on ChromeOS
-                if os.path.exists('/mnt/chromeos/removable/CIRCUITPY/'):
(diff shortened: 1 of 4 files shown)


ntoll:
OK. This is a welcome addition. Thank you so much..! :slightly_smiling_face: 

I have a few comments suggestions:

* Please `make tidy` before committing so our CI doesn't complain about code formatting. (Hint, running `make check` will run exactly the same suite of tests and checks used in our CI).
* I wonder if the changes could be moved up to the base class for modes that target MicroPython - then all devices could benefit from these changes (micro:bit, ESP etc...)? Happy to help make this happen if the code isn't obvious (tl;dr - there's a `MicroPythonMode` base class in the `mu.modes.base` module. Perhaps common code could go in there?).
* The unit tests are :+1: - thank you for ensuring your changes are exercised. :rocket:

espoelstra:
> * I wonder if the changes could be moved up to the base class for modes that target MicroPython - then all devices could benefit from these changes (micro:bit, ESP etc...)? Happy to help make this happen if the code isn't obvious (tl;dr - there's a `MicroPythonMode` base class in the `mu.modes.base` module. Perhaps common code could go in there?).

It appears the `MicroPythonMode(BaseMode)` and the `MicroBitMode(MicroPythonMode)` attempt to use serial communication as the primary means for putting/pulling files from the micro:bit and CircuitPython/pyboard both overrides to use the `device_dir`/`workspace_dir`.

I'll have to test a micro:bit from Crostini but I'll bet that without adding an additional fallback to the `mount`||`/mnt/chromeos/removable` logic it might require mapping the USB device and fixing the permissions.

carlosperate:
Right, so the micro:bit mode uses uFlash (in the contrib folder) to find the MICROBIT drive. But because python code cannot be saved directly there it has a different process to programme the device, compared with the CircuitPython mode or the MicroPython (pyboard) mode.

We can probably refactor this code into a generic `MicroPythonMode.find_device_path()` method that only does the path discovery, and then each mode can call it wherever it might need it.

mu/modes/circuitpython.py
```diff
@@ -200,6 +200,14 @@ def workspace_dir(self):
                             repr(e), mount_command
                         )
                     )
+            if os.path.exists("/mnt/chromeos"):
+                # We're on ChromeOS
+                if os.path.exists("/mnt/chromeos/removable/CIRCUITPY/"):
+                    device_dir = "/mnt/chromeos/removable/CIRCUITPY/"
+                else:
+                    self.view.show_message(
+                        'If your Circuit Python device is plugged in, you need to "Share with Linux" on the CIRCUITPY drive in the "Files" app then restart Mu.'
```
carlosperate, line 209: If you wrap this string like this it will become translatable: `_("string goes here")`

espoelstra:
@carlosperate I also fixed the "line too long" warning that failed the builds by wrapping the message across a few lines.

ntoll:
Merging.... thanks for the contribution..! :+1:

Issue #1457, opened 2021-04-09
Windows: The venv verifying modules step launches command prompt windows


carlosperate:
When Mu verifies the modules in the venv it opens background command prompt windows, which can be a bit "scary looking". 

I haven't experienced this in macOS nor Linux, so it looks like a Windows-only thing.

Would it be possible to hide these?
Just a guess, but maybe it's possible the modules are checked with pythonw.exe and we could avoid the extra windows with python.exe?

https://user-images.githubusercontent.com/4189262/114197719-856bdb80-994a-11eb-9ba1-3c82b448146a.mp4

tjguk:
@carlosperate  that's very odd. As you might imagine, I've run this process a few times myself on Windows, and I've *never* seen it launch background console windows. I note that you're running Win8.1 which *might* have something do to with it -- but I'm surprised.

Running `pythonw.exe` is probably possible, but it tends to interfere with stdout|err channels, so there might be complications.

carlosperate:
I'm seeing the same thing in Windows 10:

https://user-images.githubusercontent.com/4189262/114200707-5efb6f80-994d-11eb-95c4-9f5a718db16c.mp4

tjguk:
@carlosperate is that running from the installer? Or from a dev venv? (Just so I can reproduce...)

carlosperate:
In both cases it was from the installer.

carlosperate:
I assume this is because when running from the installer Mu is launched via `pythonw.exe`, and so the `virtual_environment.py` module launches a new process with `sys.executable`, which also uses `pythonw.exe` and we see all the terminal windows open and close.

We can update `virtual_environment.py` to:
- Check if the `sys.executable` ends in  `w.exe`,
    - If it does, check if there is a version of the exe without the `w` (so `python.exe` instead of `pythonw.exe`)
        - If it does, launch with that executable instead

PR #1774, opened 2021-08-27
venv: Hide startup pop-up command prompts on Windows.

carlosperate:
On Windows, when Popen is configured with `shell=True` it uses the wShowWindow parameter, which by default is set to SW_HIDE: https://docs.python.org/3/library/subprocess.html#subprocess.STARTUPINFO.wShowWindow

This hides the command prompt windows that appear when an installable Mu start ups: https://github.com/mu-editor/mu/issues/1457

This setting can have some unintended side effects, so sharing for feedback first as a draft PR.

### DIFF
--- a/mu/virtual_environment.py
+++ b/mu/virtual_environment.py
@@ -717,6 +717,7 @@ class VirtualEnvironment(object):
             self.interpreter,
             "-c",
             'import sys; print("%s%s" % sys.version_info[:2])',
+            shell=True if self._is_windows else False,
         )
         if not ok:
             raise VirtualEnvironmentEnsureError(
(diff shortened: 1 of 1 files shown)


ntoll:
@tjguk any thoughts on this..?

ntoll:
Will wait on feedback from @tjguk before merging/releasing.

tjguk:
> Looks like hiding these windows when not using pythonw.exe is an open task in CPython: https://bugs.python.org/issue30082

Yes, by chance I was following the discussion there just recently for other reasons.

I'm happy to merge the PR here and see if it takes us a step forward.

carlosperate:
@ntoll we can merge this PR as well.

PR #1790, opened 2021-09-06
Allow a user to change manually change their locale via a new tab in the admin dialog.

ntoll:
This "spike" does exactly what it says on the tin. Once folks are happy with the approach I'll "solidify" the code with better and more specific test coverage.

It looks like this:

mu_select_translate

Feedback welcome.

### DIFF
--- a/mu/i18n.py
+++ b/mu/i18n.py
@@ -13,3 +13,12 @@ localedir = os.path.abspath(os.path.join(os.path.dirname(__file__), "locale"))
 gettext.translation(
     "mu", localedir=localedir, languages=[language_code], fallback=True
 ).install()
+
+
+def set_language(language_code, localedir=localedir):
+    """
+    Utility function to allow admin pane updates to change the locale.
+    """
+    gettext.translation(
+        "mu", localedir=localedir, languages=[language_code], fallback=True
+    ).install()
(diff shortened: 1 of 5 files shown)


carlosperate:
Awesome!

Would a unicode emoji work in the settings tab? If Mu is configured in a language the user doesn't understand a globe icon (or flag/s?) usually helps identify the language menu.

ntoll:
Hmm.... that's a great idea and I've just tried it, but the emoji looks very (too) small. I've tried, 🌍 💬 🗣️ and 🔤 and none of them look particularly good (IMHO).

carlosperate:
Right, I guess the text in the tab is not very large.
Maybe @ZanderBrown knows how we could enlarge a unicode or svg emoji in a tab?

ntoll:
Tēnā koe. :slightly_smiling_face: :new_zealand: :rocket: 

Of course. We even have docs (that need revising) that explain how to do this: https://mu.readthedocs.io/en/latest/translations.html

ntoll:
And to be clear... we'd love a Māori translation of the UI @was-dtech - if you're offering... :+1:

was-dtech:
Tēnā koe 😀   Will get onto it this weekend.  May need some help re the technical side of things.

PR #1812, opened 2021-09-21
Restore QTextEdit-based REPL font sizes when theme is set

keith-packard:
Setting the theme resets the font size in QTextEdit panes to the
default, resetting any zoom level selected by the user. You can see
this by changing the theme or opening a new editor tab while the
MicroPython REPL or Python3 runner is open

To work around this, record the last font size set and use that to
restore the font when the theme changes.


ntoll:
Hi @keith-packard - well spotted and thank you for the fix. Merging. :shipit:

PR #1814, opened 2021-09-22
Svg icons

keith-packard:
These patches get the svg-icons created by @Rybec working with mu, along with adding a couple more simple SVG icons for the scroll bar arrows and the 'mu' logo.


ntoll:
@keith-packard thank you for these changes..! The failing tests in CI were due to network errors at GitHub's end. I've tested locally, so merging. :+1: :shipit:

PR #1815, opened 2021-09-22
.github: Switch test matrix from Ubuntu 16.04 to Ubuntu 18.04

keith-packard:
Github is deprecating Ubuntu 16.04 runners, which makes tests using
them fail. Switch to a later version.


carlosperate (approved):
Thanks @keith-packard!

Issue #1813, opened 2021-09-22
Error decoding user Python script


carlosperate:
So far this has only been reported on beta.5, although there is nothing around the code dealing with this that was modified for that release.

```
"C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\__main__.py", line 9, in <module>
    main()
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\__main__.py", line 5, in main
    run()
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\app.py", line 365, in run
    editor.restore_session(sys.argv[1:])
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\logic.py", line 888, in restore_session
    self.direct_load(old_path)
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\logic.py", line 1145, in direct_load
    self._load(path)
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\logic.py", line 1017, in _load
    text, newline = read_and_decode(path)
  File "C:\Users\usuario\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\logic.py", line 318, in read_and_decode
    text = btext.decode(encoding)
LookupError: unknown encoding: uft-8
```
Other forms:
```
  File "C:\Users\PC\AppData\Local\Programs\Mu Editor\Python\lib\site-packages\mu\logic.py", line 318, in read_and_decode
    text = btext.decode(encoding)
LookupError: unknown encoding: 2
```

Affected issues/crash reports:
- https://github.com/mu-editor/mu/issues/1781
- https://github.com/mu-editor/mu/issues/1770
- https://github.com/mu-editor/mu/issues/1771
- https://github.com/mu-editor/mu/issues/1772
- https://github.com/mu-editor/mu/issues/1780
- https://github.com/mu-editor/mu/issues/1837

[...]

PR #1817, opened 2021-09-23
Fix crash when user script encoding cookie is invalid.

carlosperate:
If the user script has a encoding shebang/cookie (`#-*- coding: utf-8 -*-`) it will try to use that to decode the text. If the codec name is invalid Python throws a `LookupError` exception.

Fixes https://github.com/mu-editor/mu/issues/1813.


ntoll:
Aha... that's a fun find. Thanks for the fix @carlosperate :+1: :shipit:

PR #1818, opened 2021-09-23
Avoid re-scaling SVG icon pixmaps

keith-packard:
When an SVG file is loaded to a QPixmap, it is rasterized at the default size specified in the SVG file. If it is then used at a different size, the temporary pixmap image is rescaled. It's better to generate the pixmap at the target size directly as it generates cleaner results without rescaling artifacts and saves time and memory associated with the temporary image.

This is done by adding a named 'size' parameter to the load_pixmap function and then having that function use the existing
load_icon function and using QIcon's ability to generate a pixmap of the desired size.


ntoll:
@keith-packard you're on fire :fire: - thanks for this contribution. :slightly_smiling_face:

keith-packard:
> @keith-packard you're on fire fire - thanks for this contribution. slightly_smiling_face

Thanks! Now my patch queue is down to just the changes needed to run snek. I got the word that we might be starting classes again in January, so I'm cleaning up my classroom tools, and mu is a big part of that! Thanks for the awesome system!

PR #1828, opened 2021-09-28
Avoid TypeErrors when moving windows to float positions

hroncok:
On Python 3.10+, we cannot use floats for int arguments:

TypeError: arguments did not match any overloaded call:
  move(self, QPoint): argument 1 has unexpected type 'float'
  move(self, int, int): argument 1 has unexpected type 'float'

### DIFF
--- a/mu/interface/main.py
+++ b/mu/interface/main.py
@@ -1098,8 +1098,8 @@ class Window(QMainWindow):
             x = None
         if y and (y <= 0 or y > screen_height):
             y = None
-        x = (screen_width - size.width()) / 2 if x is None else x
-        y = (screen_height - size.height()) / 2 if y is None else y
+        x = (screen_width - size.width()) // 2 if x is None else x
+        y = (screen_height - size.height()) // 2 if y is None else y
         self.move(x, y)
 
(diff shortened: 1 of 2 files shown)


ZanderBrown:
> On Python 3.10+, we cannot use floats for int arguments:

I don't remember seeing that in the 3.10 release notes? Makes some sense but quite the gotcha

hroncok:
Will also adapt test.

hroncok:
> Builtin and extension functions that take integer arguments no longer accept Decimals, Fractions and other objects that can be converted to integers only with a loss (e.g. that have the `__int__() `method but do not have the `__index__()` method). (Contributed by Serhiy Storchaka in [bpo-37999](https://bugs.python.org/issue37999).)

https://docs.python.org/3.10/whatsnew/3.10.html#other-language-changes

I know that PyQt and WxPython apps are affected by this.

ntoll:
Hi @hroncok - thanks for bringing this to our attention and for submitting a patch.

I get a unit test failure in `test_Window_autosize_window_off_screen`. Will revise the test and push/merge.

Thank you again..! :+1:

PR #1844, opened 2021-10-04
Depend on `ipython_genutils` to fix issue with missing `ipykernel` dependency

microbit-carlos:
We keep `ipykernel` below version 6 to keep compatibility with macOS 10.13 (and possibly 10.14 as well), since v6 includes a dependency on `debugpy` and that only has wheels for newer macOS versions. This also maintains compatibility with Python 3.6, as v6 is Python 3.7+.

Turns out v5 of `ipykernel` depends on `ipython_genutils`, but this isn't explicitly declared as a dependency in the package. It also depends on `traitlets`, which incidentally brought `ipython_genutils`, so it's worked fine so far. However `traitlets` dropped this dependency in v5.1, so `ipython_genutils` isn't picked anymore and as a workaround we need to manually specify it in the Mu dependencies and the user venv dependencies.


carlosperate:
@tmontes I think you mentioned it might be enough to add this dependency only to the user virtualenv, but I've added it to setup.py just in case as well.

carlosperate (approved):
Merging quickly to be able to create working builds again.

tmontes:
> @tmontes I think you mentioned it might be enough to add this dependency only to the user virtualenv, but I've added it to setup.py just in case as well.

...ok, should be harmless (crossing fingers!). Things are looking good on my side:

* Cleaned up everything (dev + user venvs).
* Updated `master`.
* Created dev venv from scratch + `pip` installed Mu onto it.
* Mu launched and ran fine from source.
* Packaged with `pup`, installed, Mu launched ran fine.
* Both on macOS and Windows.

Go @ntoll! :)

PR #1795, opened 2021-09-09
Updating pt-PT translation

xbecas:
Contributing to #1138.

Issue #1138, opened 2020-10-26
Update translations for the v1.1 release


carlosperate:
This is to be done **only** when we are hitting the beta releases and the strings are stable. This is to minimise changing text and hopefully simplify the work of the translators.

We might need to reach out to our wonderful community to have a look at the translations at that point.

To help the process when the strings are ready we could also update all the pot files to include the updated strings as that could save everybody a step and reduce the room for error.

carlosperate:
Related issues:
- Check Chinese translation. #534
- Turkish Translation #598
- Swiss german translation (de_CH) #785
- A french spelling mistake in mu.mo "Ecris ton programme" #967

Related PRs:
- Locale SK #322
- Create mu.po #962 (Ukrainian)
- update of slovak translation #990
- fr translation typo #1041
- Fixes for French localization #1043

So if we do this:

> We might need to reach out to our wonderful community to have a look at the translations at that point.

It might be better to merge some of these PRs first as that would have less conflicts at this stage, then update the translation files, and then ask translators to review them?

xbecas:
Hi all. This is my first contribution attempt. @tmontes has been assisting me at setting up a development enviroment so I could start contributing with the Portuguese translation update.

I'm facing the issue that @carlosperate mentioned above, namely trying to come up with an updated .po file.

When I run the `make translate` I get an empty .po file. When I run the `make translateall` I get the following error:
```
(.venv) C:\Docs\GIT\mu\mu.git>python make.py translateall
*** mu\modes\web.py:176: Seen unexpected token "+"
*** mu/modes\web.py:176: Seen unexpected token "+"
Traceback (most recent call last):
  File "c:\users\vasco.mora\appdata\Local\Programs\Python\Python38\tools\i18n\pygettext.py", line 626, in <module>
    main()
  File "c:\users\vasco.mora\appdata\Local\Programs\Python\Python38\tools\i18n\pygettext.py", line 619, in main
    eater.write(fp)
  File "c:\users\vasco.mora\appdata\Local\Programs\Python\Python38\tools\i18n\pygettext.py", line 460, in write
    print('msgid', normalize(k, encoding), file=fp)
  File "c:\users\vasco.mora\appdata\Local\Programs\Python\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u03c0' in position 480: character maps to <undefined>

[...]

PR #1851, opened 2021-10-10
Revamped i18n process using Babel.

tmontes:
All,

Have been working with @xbecas in improving the general tranlation process in the context of #1138 and #1795.
We have come up with this, now Babel based.

We propose a two-step process, going forward, (optionally) free from tools such as PoEdit:

* Translators run `make translate-prep LOCALE=xx_XX`:
  * This either creates a new `mu.po` file if working with a new translation...
  * ...or updates the existing one taking into accoung new/removed message strings/ids.
* Once the `mu.po` file is edited/translated, translators can then run `make translate-compile LOCALE=xx_XX` to compile the `mu.po` to `mu.mo` file, ready to be commited.

MORE:
* The changes to the `Makefile` are still very rough and could use improvements (do we have skills here?). :)

NEXT:
* Trying to gather preliminary feedback from you.
* Update the docs at https://mu.readthedocs.io/en/latest/translations.html.
* Update `make.py` to provide the same capabilities (side-thought: having `Makefile` and `make.py` is really a bad idea... should we go `make.py` all the way?)

THOUGHTS?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,9 @@ all:
 	@echo "make publish-test - publish the project to PyPI test instance."
 	@echo "make publish-live - publish the project to PyPI production."
 	@echo "make docs - run sphinx to create project documentation."
-	@echo "make translate_begin LOCALE=xx_XX - create/update a mu.po file for translation."
-	@echo "make translate_done LOCALE=xx_XX - compile translation strings in mu.po to mu.mo file."
-	@echo "make translate_test LOCALE=xx_XX - run translate_done and launch Mu in the given LOCALE."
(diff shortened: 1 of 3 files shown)


tmontes:
I feel this is now in a merge-able state.

SUMMARY

* New translation procedure:
  1. `make translate_begin LANG=xx_XX` -- creates/updates the source `.po` file.
  2. Edit `.po` file with tool of choice: Poedit, text editor, you name it.
  3. `make translate_test LANG=xx_XX` -- compiles `.po` to `.mo` and launches Mu in that LANG.
  4. Create PR.

* Same procedure for updating existing translations or creating new ones.
* Same procedure on macOS, Linux, and Windows.
* Docs updated.

PR #1863, opened 2021-10-23
Virtual env runtime fix for running as Linux AppImage.

tmontes:
FACT:

* While exploring Linux AppImage packaging [in this `pup` PR](https://github.com/mu-editor/pup/pull/183) I found a serious limitation.
* Each time Mu runs (as an AppImage thingie) it is mounted under a different mount point.
* This invalidates existing virtual envs because their interpreter -- a symlink to the `python` executable in Mu's installation -- no longer exists (it does, but each time under a different path).

THIS PR:

* Detects that we're runing as an AppImage.
* Re-symlinks the existing virtual env (if any) interpreter to the currently valid system interpreter.

PS: Isn't packaging fun? :)


mu/virtual_environment.py
```diff
@@ -696,6 +696,21 @@ def ensure_interpreter(self):
...
+        elif os.environ.get("APPDIR") and os.environ.get("APPIMAGE"):
+            # When packaged as a Linux AppImage, Mu Editor is mounted
+            # on a random(ish) path each time it runs. This breaks the
+            # existing venv: its interpreter symlinks to a path that
+            # likely no longer exists. This horrible hack fixes that
+            # by re-symlinking to the now valid interpreter path.
+            # PS: This is a horrible hack and it seems to work! :)
+            logger.info("No interpreter found at: %s", self.interpreter)
+            os.unlink(self.interpreter)
+            os.symlink(sys.executable, self.interpreter)
```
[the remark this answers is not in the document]
tmontes: Generally agreed: the `--upgrade` `venv` option migh serve us, *but* we're using `virtualenv` which, from what I've seen, does not expose such a "feature".

Maybe for "minimal guarantees" we could use the `--symlinks` (which defaults to `True`) when calling `python -m virtualenv` in `mu/virtual_environment.py:773`. Other than that, not really sure... :)

tmontes:
Updated the PR:

* Use the `--symlink` option when creating the virtual env on non-Windows platforms.
  (wondering about the `filter` approach I took: can it be made more readable?)
* Ignore (but log) failures in removing what is expected to be a stale symlink.

@tjguk, @carlosperate, @ntoll: your input is welcome. :)

carlosperate:
I'm very low on bandwidth to be able to test this until the end of the month, but the code looks good to me and I'd be happy for the PR to be merged it if you've already tested it 👍

carlosperate:
I'd suggest to wait for the v1.1 final release before we merge AppImage support, we can then start testing this on Linux on with the first 1.2-beta.

ntoll:
Ack on post 1.1.

tmontes:
NOTE:

* Python 3.7 on macOS had failed for a non-obvious motive: retriggered the job run.

mu/virtual_environment.py
```diff
@@ -782,17 +804,21 @@ def create_venv(self):
...
+            None,
+            (
+                sys.executable,
+                "-I",
+                "-m",
+                "virtualenv",
+                "-p",
+                sys.executable,
+                "-q",
+                "" if self._is_windows else "--symlinks",
```
carlosperate, line 817: Turns out the `--symlinks` flag was only added to the `virtualenv` pacakge in version 20.0.0.
We currently support v16 and Debian/Raspbian Buster ships with v15, so in those environments this fails the virtualenv creation.

carlosperate: https://github.com/mu-editor/mu/issues/2256

Issue #1454, opened 2021-04-09
CI tests segmentation fault


carlosperate:
It looks like it is happening more often. In the past 10 runs or so it seems like there is always 1 job (out of 21) that gets a seg fault:

```
tests/interface/test_main.py::test_Window_setup Fatal Python error: Segmentation fault

Thread 0x00007f141db02700 (most recent call first):
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/unittest/mock.py", line 1857 in _mock_set_magics
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/unittest/mock.py", line 1833 in __init__
  File "/home/runner/work/mu/mu/tests/interface/test_main.py", line 1531 in test_Window_setup
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/_pytest/python.py", line 183 in pytest_pyfunc_call
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/_pytest/python.py", line 1641 in runtest
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/_pytest/runner.py", line 162 in pytest_runtest_call
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/manager.py", line 87 in <lambda>
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pluggy/manager.py", line 93 in _hookexec
[...]

PR #1874, opened 2021-11-02
Preliminary Linux AppImage packaging.

tmontes:
As discussed:

* This updates the `Makefile` with a new `linux` target: brings in the latest `pup` and packages Mu as a Linux AppImage thing -- currently 3.8 x86_64 only!

Goes along with:

* PR #1863 that makes running Mu as an AppImage usable.
* The latest `pup` release 1.0.0a14.

Outstanding:

* Updating GitHub Actions to automate this somehow -- @carlosperate, can you assist, please?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,20 @@ macos: check
 	ls -la ./build/pup/
 	ls -la ./dist/
 
+linux: check
+	@echo "\nFetching wheels."
+	python -m mu.wheels
+	@echo "\nPackaging Mu into a Linux AppImage."
+	python -m virtualenv venv-pup
+	# Don't activate venv-pup because:
+	# 1. Not really needed.
+	# 2. Previously active venv would be "gone" on venv-pup deactivation.
+	./venv-pup/bin/pip install pup
+	./venv-pup/bin/pup package --launch-module=mu --nice-name="Mu Editor" --icon-path=./mu/resources/images/icon.png --license-path=./LICENSE .
+	rm -r venv-pup
+	ls -la ./build/pup/
+	ls -la ./dist/
+
 video: clean
 	@echo "\nFetching contributor avatars."
 	python utils/avatar.py


carlosperate:
I think the only issue in the CI was this: https://github.com/mu-editor/mu/issues/1454
So I've restarted the job, I'd expect it to pass this time.

carlosperate:
I was trying to get the CI to build the AppImage, but for some reason it **always** seg faults during the test, not sure I understand why yet.

tmontes:
> I was trying to get the CI to build the AppImage, but for some reason it **always** seg faults during the test, not sure I understand why yet.

...hmmm, interesting. What exactly is segfaulting? Maybe `pup` is doing something very wrong... Can you link to a log or some such?

carlosperate:
You can apply this patch to your fork and see it on the CI:
- patch: https://github.com/carlosperate/mu/commit/c3d2ef38d9419bdc6f7c528210223284f821072b.patch
- commit: https://github.com/carlosperate/mu/commit/c3d2ef38d9419bdc6f7c528210223284f821072b
- comparison: https://github.com/tmontes/mu/compare/linux-appimage-packaging...carlosperate:linux-appimage-packaging

tmontes:
> You can apply this patch to your fork and see it on the CI:
> 
> * patch: https://github.com/carlosperate/mu/commit/c3d2ef38d9419bdc6f7c528210223284f821072b.patch
> * commit: [carlosperate@c3d2ef3](https://github.com/carlosperate/mu/commit/c3d2ef38d9419bdc6f7c528210223284f821072b)
> * comparison: [tmontes/mu@linux-appimage-packaging...carlosperate:linux-appimage-packaging](https://github.com/tmontes/mu/compare/linux-appimage-packaging...carlosperate:linux-appimage-packaging)

Thanks for the patch. Tried it out and understood the failure:

* `make linux` depends on `make check` which runs the tests.
* Tests need to be run with `xvfb-run make check`.
* Thus, packaging needs to be run with `xvfb-run make linux`.
* That, in turn, needs a few OS level packages installed beforehand, per the "Run tests" workflow:
```
sudo apt-get update
sudo apt-get install -y libxkbcommon-x11-0 xvfb
```

PS: Also, the final artifact extension is `.AppImage`, capitalized.

With these changes hacked in a private branch I managed to have CI produce the artifact. :)

carlosperate:
Of course, I can't believe I forgot about xvfb after having to spend so much time fighting with it and PyQt 😅

From my point of view I'd remove the `check` step in the builds and replace it with `clean` instead. It's always good to make sure the tests pass before we make a release, but there can me some merit in building test installers with failing tests.

tmontes:
> From my point of view I'd remove the `check` step in the builds and replace it with `clean` instead. It's always good to make sure the tests pass before we make a release, but there can me some merit in building test installers with failing tests.

I can see both pros and cons in removing the `check` dependency here: wondering if it isn't better to just leave it as is and manually remove it on the odd occasion when we want to do that. Assuming, of course, we're hand-packaging with `make <platform>`; if it's CI, the reasoning might differ.

Regardless of such a decision, I'd argue that it should have its own PR. :)

tmontes:
For future fun:

* Played with this for a while with three VMs: Ubuntu 18.04, Debian 10, CentOS 7.
* GNOME Desktop Environment in all.
* Packaged on CentOS 7 works ok on Ubuntu 18.04 and Debian 10.
* Packaged on any of the others does not work on CentOS 7.
* More things... Coming down to which platform tag is "seleted by `pip`" when grabbing linux binary wheels...

I say we go forward with this, along with #1863 and let users know this is VERY EXPERIMENTAL. :)

carlosperate:
Do you want to add the CI builds in this PR as well? Or will that go in a new one?
I think for now having it build on the default GitHub `ubuntu-18.04` image is fine, but in the future we probably want to use a docker container with an old OS version so that it links to a older version of glibc (assuming that is still an issue with AppImages). For initial testing a CI build in `ubuntu-18.04` is still much much better than nothing.

ZanderBrown:
> GNOME Desktop Environment in all.

?

carlosperate:
I think he means the 3 distros he tried (Ubuntu, Debian and Centos) were all running GNOME. So we haven't tested other desktops like KDE, etc.

tmontes:
> Do you want to add the CI builds in this PR as well? Or will that go in a new one?

I'd go fully "modular" and add the CI builds in a subsequent PR, yes. :)

> I think for now having it build on the default GitHub `ubuntu-18.04` image is fine, but in the future we probably want to use a docker container with an old OS version so that it links to a older version of glibc (assuming that is still an issue with AppImages). For initial testing a CI build in `ubuntu-18.04` is still much much better than nothing.

Agreed. Hadn't thought of using docker to build "more compatible" AppImages that, in our case, depend on the system's libc. Sounds like a way forward!

carlosperate:
I'd suggest to wait for the v1.1 final release before we merge AppImage support, we can then start testing this on Linux on with the first 1.2-beta.

ntoll:
Just catching up on all the PRs. Merry Christmas holidays everyone... :santa: :tada: 

This is great work. I think it would be good to get a "packaged" version of Linux out there as a post 1.1 task. Perhaps as part of the (inevitable) 1.1.1 release, for instance... and we can rebase any changes forward to 1.2 (or 2.0?) .... to be confirmed.

PR #1900, opened 2021-12-02
Enable f-strings in dark modes. Fixes #1707.

### LINKED ISSUE #1707: Dark Mode should support f"{x}:{y}"

Use an f-string while in dark mode.

```python
x=20
y=30
print(f"{x}:{y}")
```

f-strings should look like other strings

### What actually happened?

The f-string is purple on white. The variables (x,y) are gray on black.

### Operating System Version

Windows 10

### Mu Version

1.1.0-beta-5

### Other Info

_No response_

### Editor Log

_No response_

devdanzin:
Mu themes lacked styling for the following f-string variants:
- DoubleQuotedFString
- SingleQuotedFString
- TripleDoubleQuotedFString
- TripleSingleQuotedFString

This PR adds them, using the same styling as normal strings. We could bikeshed on a better style to set f-strings apart.

### DIFF
--- a/mu/interface/themes.py
+++ b/mu/interface/themes.py
@@ -129,6 +129,7 @@ class DayTheme(Theme):
     SingleQuotedString = DoubleQuotedString = Font(color="#800000")
     SingleQuotedFString = DoubleQuotedFString = Font(color="#800000")
     TripleSingleQuotedString = TripleDoubleQuotedString = Font(color="#060")
+    TripleSingleQuotedFString = TripleDoubleQuotedFString = Font(color="#060")
     Number = Font(color="#00008B")
     Decorator = Font(color="#cc6600")
     Default = Identifier = Font()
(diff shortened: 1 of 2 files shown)


carlosperate:
Thanks for the PR @devdanzin! 

It looks like the tests running on Arm are failing:

```
out: =========================== short test summary info ============================
out: FAILED tests/interface/test_themes.py::test_theme_apply_to - assert 17 == 21
out: ============ 1 failed, 1038 passed, 21 skipped in 413.87s (0:06:53) ============
``` 

I assume this is because they use the Debian packages, which would be older versions of Qscintilla and related, so Mu and the tests will need to take into account that maybe these older lexers don't have implemented fstrings.
Will this work correctly with older QScintilla versions? Hopefully the values are just ignored and no error are thrown?

If you could also add a small before/after screenshot that'd be great too!

devdanzin:
Yes, the logs show e.g. arm-debian uses `QScintilla-2.10.4` and I was able to reproduce the failing test on Windows using `QScintilla-2.10.8`. Good news: this older version doesn't have the bug, nor did it show any ill effects from having the fix applied, apparently ignoring the values as you hoped.

Here's the behavior on new enough versions, same as in the bug report:
before_fstring_fix

And here is what using the normal string styling looks like:
after_fstring_fix

Not so good news: the only way I figured to have the test work with old and new versions is to check whether the lexer supports f-strings and then add four more calls to the expected value. I'm hoping a more elegant way will appear, but that's where it stands right now.

carlosperate (approved):
Tested locally and works wonderfully, all CI tests pass as well, so it's good to merge!
Thanks @devdanzin!

Issue #1292, opened 2021-02-11
Project versioning mismatch


carlosperate:
The GH releases use semver:
- https://github.com/mu-editor/mu/releases/tag/1.1.0-beta.1
- https://github.com/mu-editor/mu/releases/tag/1.1.0-alpha.2

The Python package uses something that is not semver (`1.1.0-beta.1`), is not a [PEP 440](https://www.python.org/dev/peps/pep-0440/) beta release  (`1.1.0b1`/`1.1.0beta1`), doesn't quite fit PEP 440 post release format (`1.1.0.postRelease1`), and probably matches a "development version of a post release" (`X.Y.postN.devM`):
https://github.com/mu-editor/mu/blob/26abb94b309a48aa550524de832010a07b09ace8/mu/__init__.py#L9

The PyPI version ends up being correctly formatted as PEP 440. Does PyPI automatically correct it?
For example, with PyPI version `1.0.0rc1`:
- https://pypi.org/project/mu-editor/1.0.0rc1/  <-- PEP 440
- https://github.com/mu-editor/mu/releases/tag/v1.0.0.rc.1   <-- Semver
- https://github.com/mu-editor/mu/blob/v1.0.0.rc.1/setup.py#L36  <-- Semver
    - https://github.com/mu-editor/mu/blob/v1.0.0.rc.1/mu/__init__.py

The Windows installer actually ends up using PEP 440 format correctly, although I'm not sure how it does this @tmontes?
The resulting file name is `Mu Editor 1.1.0b1.msi`.

@ntoll do you have a specific format you'd like to consolidate? This can be important if in the future we need to store and compare the Mu version for installation upgrades.

tmontes:
> The Windows installer actually ends up using PEP 440 format correctly, although I'm not sure how it does this @tmontes?
> The resulting file name is `Mu Editor 1.1.0b1.msi`.

There are two different things at play, with Windows MSI packaging:

* One is common to all packaging: getting the PEP 440 version. This is done by extracting the version from the metadata of a wheel that `pup` produces from its "source" (see https://github.com/mu-editor/pup/blob/master/src/pup/plugins/metadata.py#L67). That's what you get in the name of the MSI file as well as in the name of the installed product under Windows Programs and Features.
[...]

PR #1908, opened 2021-12-06
Fix Mu version format to match PEP 440.

carlosperate:
To fix https://github.com/mu-editor/mu/issues/1292.

Basically the current version string is neither [semver](https://semver.org/) nor [PEP 440](https://www.python.org/dev/peps/pep-0440/):
https://github.com/mu-editor/mu/blob/ee049b7941f2606b2dfe4c9bfac328995039ebc4/mu/__init__.py#L9

Semver has been used for the GitHub release tags, and PEP440 for the PyPI releases.

For this PR I've picked PEP440 simply because from my perspective is probably better to continue using the same PyPI version format and change the format used in any future git tags, than the other way around. Since PEP440 is the version users have to type when pip installing, it's probably more "user facing": https://pypi.org/project/mu-editor/1.1.0b6/#history

As @tmontes mentions in https://github.com/mu-editor/mu/issues/1292#issuecomment-778513976 neither version format will work on the Windows installer, so doesn't matter too much which one to pick.

Happy to change it to semver if anybody has got any strong opinion about it.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,7 +58,7 @@ author = "Nicholas H.Tollervey"
 from mu import __version__ as full_version
 
 # The short X.Y version.
-version = ".".join(full_version.split(".")[:3])
+version = ".".join(full_version.split(".")[:2])
 # The full version, including alpha/beta/rc tags.
 release = full_version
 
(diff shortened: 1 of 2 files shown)


ntoll:
Outstanding work @carlosperate :grinning:

LGTM :shipit:

PR #1911, opened 2021-12-08
Fix deleting the first character of next line when toggling comment (Ctrl+K). Fixes #1511

### LINKED ISSUE #1511: Commenting (CTRL-K) deletes first character in the next line
If you are reporting a bug, we would like to know:

* What you were trying to do,
From this code I tried to **comment the first `pixel.show()`**
`pixel[1] = (255, 255, 255)`
`pixel.show()`
`pixel[2] = (255, 255, 255)`
`pixel.show()`

* What steps you took to make this happen,
I placed the cursor at the beginning of the line **without highlighting** the command and I pressed CTRL-K

* What you expected to happen,
The line will be commented, like this:
`# pixel.show()`

* What actually happened,
The line is commented and highlighted, but the cursor placed into the next line and the **next line first character is deleted**:
`pixel.show()`
`ixel[2] = (255, 255, 255)`

**IF the line is highlighted before CTRL-K, there is no any problem.**

* Why this difference is problematic (it may not be a bug!),
Because one character is deleted ...

* Technical details like the version of Mu you're using, your OS version and
  other aspects of the context in which Mu was running. 
MU version: 1.1.0.beta.2
OS: Windows 10 Pro, version 2004

Please remember to attach a **copy of the full log files for Mu**. 
The log is attached.
[log_Commenting (CTRL-K) deletes first character in the next line_Tizedes.txt](https://github.com/mu-editor/mu/files/6382285/log_Commenting.CTRL-K.deletes.first.character.in.the.next.line_Tizedes.txt)

devdanzin:
The `EditorPane.toggle_comments` command deletes the first character of the next line if the toggled line ends in `CRLF`. That method expects `EditorPane.toggle_line` to remove one EOL character, but it removes two in this case. We currently add `CRLF` line endings when editing a file[0] on Windows.

This minimal fix replaces `"\r\n"` with `"\n"` to keep changes simple. Having mixed EOL styles in a file shouldn't be a big concern, because that's what already happens on Windows[1], It also doesn't affect which EOL the file will be saved with, as that is determined on load or creation.

There's another way to fix this: make Windows also generate `LF` EOL by using `EditorPane.setEolMode(QsciScintilla.EolUnix)` (I'll add an issue describing that option). I believe having both fixes is better, but am open to picking either.

 [0] which will mostly have `LF` endings if saved from new tab or load, except if loaded from `CRLF` terminated file:
Loaded with `LF` EOL:
lf

Enter adds `CRLF` as EOL:
crlf

[1] We can see the EOL by setting `EditorPane.setEolVisibility(True)`.

### DIFF
--- a/mu/interface/editor.py
+++ b/mu/interface/editor.py
@@ -632,7 +632,8 @@ class EditorPane(QsciScintilla):
             # Toggle the line currently containing the cursor.
             line_number, column = self.getCursorPosition()
             logger.info("Toggling line {}".format(line_number))
-            line_content = self.text(line_number)
+            # Replace CRLF line endings that we add when run on Windows
+            line_content = self.text(line_number).replace("\r\n", "\n")
             new_line = self.toggle_line(line_content)
             self.setSelection(line_number, 0, line_number, len(line_content))
             self.replaceSelectedText(new_line)
--- a/tests/interface/test_editor.py
+++ b/tests/interface/test_editor.py
@@ -928,6 +928,18 @@ def test_EditorPane_toggle_comments_selection_follows_len_change():
     ep.setSelection.assert_called_once_with(0, 0, 2, 4)
 
 
+def test_EditorPane_toggle_comments_handle_crlf_newline():
+    """
+    Check that stray "\r\n" line endings don't lead to deleting the first
+    character of the following line.
+    """
+    ep = mu.interface.editor.EditorPane(None, "test\r\nline 2\n")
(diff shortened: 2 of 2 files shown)


ntoll:
@devdanzin thank you for the contribution. Will merge.

> I believe having both fixes is better, but am open to picking either.

Care to add an additional PR to add the `EditorPane.setEolMode(QsciScintilla.EolUnix)` fix too..?

PR #1916, opened 2021-12-10
Make Editor.envars a dict, so we avoid duplicated entries. Fixes #1736.

### LINKED ISSUE #1736: Web mode: Every time a web script is run it adds the same Flask env variables

Run a flask script in Web Mode

- Open Mu
- Set Web mode
- Open settings
- Remove all env variables
  image
- Run and stop a flash script
- Open the settings, the env variables have been added:
  image
- Close the settings
- Run a stop a flask script again
- Open the settings, the same variables have been duplicated
  image

No duplications of the same env variables

### What actually happened?

The same env variables are added every time the script is run

### Operating System Version

macOS 11

### Mu Version

1.1.0b5

### Other Info

N/A

### Editor Log

N/A

devdanzin:
This solves the issue of duplicated entries on the "Python3 Environment" tab, like those that get added when running Web mode server (#1736).

To avoid errors from old session files, `Editor.restore_session` still accepts envars as list, but converts them to dict.

### DIFF
--- a/tests/interface/test_panes.py
+++ b/tests/interface/test_panes.py
@@ -1707,7 +1707,7 @@ def test_PythonProcessPane_start_process_user_environment_variables():
     ):
         mock_sys.platform = "darwin"
         ppp = mu.interface.panes.PythonProcessPane()
-        envars = [["name", "value"]]
+        envars = {"name": "value"}
         ppp.start_process(
             interpreter,
             script_filename,
--- a/mu/interface/panes.py
+++ b/mu/interface/panes.py
@@ -876,9 +876,11 @@ class PythonProcessPane(QTextEdit):
         interpreter used to launch the child process.
         """
         self.is_interactive = interactive
-        if not envars:  # Envars must be a list if not passed a value.
-            envars = []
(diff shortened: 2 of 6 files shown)


tests/interface/test_panes.py
```diff
@@ -1707,7 +1707,7 @@ def test_PythonProcessPane_start_process_user_environment_variables():
     ):
         mock_sys.platform = "darwin"
         ppp = mu.interface.panes.PythonProcessPane()
-        envars = [["name", "value"]]
+        envars = {"name": "value"}
```
carlosperate, line 1710: Could we also add a test like this were we load a list with repeated values and assert that we still get a single entry?
I've run a manual tests locally by editing the `session.json` file, but it'd be good to have it in the CI as well.

carlosperate (commented):
This is great, thanks @devdanzin!

Code looks good, and gave this a spin and works wonderfully.
I've just added a request for an extra tests if that's okay.
Great work and thanks again!

tests/interface/test_panes.py
devdanzin: Thanks, done!

carlosperate:
Perfect, thanks @devdanzin!

Issue #1867, opened 2021-10-29
Crash Report


guru8000:
Hey, everything was ok with Mu editor untill I installed python on my computer. Then I couldn't open MuEditor anymore. It wasn't crashing, just wasn't opening. I uninstalled it and installed again - didn't help. Then I uninstalled python, reinstalled MuEditor, didn't help. all of it was yesterday. Today, I reversed my computer to the saved point from before installing python. Reinstalled MuEditor and the crash report appeared. Please help.

**REMEMBER TO ATTACH YOUR LOG FILE**

Version: 1.1.0.beta.5
Locale: pl_PL
Platform: Windows 10 10.0.19041 AMD64

```
ll_wheel
    generated_console_scripts = maker.make_multiple(scripts_to_generate)
  File "C:\Users\dell\AppData\Local\python\mu\mu_venv-38-20211029-085254\lib\site-packages\pip\_vendor\distlib\scripts.py", line 422, in make_multiple
    filenames.extend(self.make(specification, options))
  File "C:\Users\dell\AppData\Local\python\mu\mu_venv-38-20211029-085254\lib\site-packages\pip\_internal\operations\install\wheel.py", line 448, in make
    return super().make(specification, options)
  File "C:\Users\dell\AppData\Local\python\mu\mu_venv-38-20211029-085254\lib\site-packages\pip\_vendor\distlib\scripts.py", line 411, in make
    self._make_script(entry, filenames, options=options)
  File "C:\Users\dell\AppData\Local\python\mu\mu_venv-38-20211029-085254\lib\site-packages\pip\_vendor\distlib\scripts.py", line 312, in _make_script
    self._write_script(scriptnames, shebang, script, filenames, ext)
  File "C:\Users\dell\AppData\Local\python\mu\mu_venv-38-20211029-085254\lib\site-packages\pip\_vendor\distlib\scripts.py", line 242, in _write_script
    launcher = self._get_launcher('t')
  File "C:\Users\dell\AppData\Local\python\mu\mu_venv-38-20211029-085254\lib\site-packages\pip\_vendor\distlib\scripts.py", line 390, in _get_launcher
    raise ValueError(msg)
ValueError: Unable to find resource t64.exe in package pip._vendor.distlib

`
[mu.log](https://github.com/mu-editor/mu/files/7439644/mu.log)
``

tjguk:
[...]

PR #1921, opened 2021-12-13
Issue1831 isolate venv

tjguk:
(Change suggested by user @hmmkay)

We're seeing cross-talk when creating the virtual environment which can wrongly pick up system-owned packages rather than the ones for the local venv.

This change uses the `-I` switch to the python executable which temporarily isolates that Python environment


carlosperate:
We've got the first confirmation that the installers from this PR work 🎉
- https://github.com/mu-editor/mu/issues/1867#issuecomment-994803919

I've asked users to test the installer in all the other related Crash Reports, so we can wait a bit more to hear back, but looks great so far :)

carlosperate (approved):
We got a couple of confirmations that this is working, so we are good to merge, thanks @hmmkay and @tjguk!

PR #1923, opened 2021-12-13
microbit: fix typo

rffontenelle:
"to saved" -> "to save"

### DIFF
--- a/mu/modes/microbit.py
+++ b/mu/modes/microbit.py
@@ -326,7 +326,7 @@ class MicrobitMode(MicroPythonMode):
                     _("Cannot save a custom hex file to a local directory."),
                     _(
                         "When a custom hex file is configured in the settings "
-                        "a local directory cannot be used to saved the final "
+                        "a local directory cannot be used to save the final "
                         "hex file."
                     ),
                 )


carlosperate:
Thanks for the contribution @rffontenelle!

@ntoll @tmontes I assume we are open to merge fixes to the original English strings, would that right? Changing them will have an impact to any open translation PR, so I don't know if we are planning to "lock it" at some point.

@tmontes @xbecas If an English string is updated, would that mean than any translated string is lost? If `make translate_begin` is run I assume the old translations of this string might disappear, correct? And if the translation files are **not** updated I assume this would now count as a new string and therefore it won't be translated on the UI, would that be correct?
Does changing English strings have any other impact?

xbecas:
Hi all. As far as I can tell the pybabel engine presents fuzzy matches, so while a small fix / change *will definitely show as a different string*, there is a good chance that pybabel matches it and suggests the match with the previously translated sentence.

My suggestion is that we do not mess with these fixes right away and build a list of the original English typos to fix them after the translations (maybe going through the locales one-by-one, if necessary).

PS: If the fuzzy match fails the impact is that the sentence is left untranslated, thus being presented in English.

rffontenelle:
I confirmed that the current setup will mark the string as *fuzzy* including the previous approximated translation. It means that all languages will have to be updated, or de-fuzzied, in order to not show this string in English.

Please notice that this typo is almost undetected when translating, chances are that the translators didn't even notice it and contributed with the proper translation. In this case, a simple remove of the *fuzzy* flag would solve it.

carlosperate:
Thanks for the clarifications!

@rffontenelle once we merge the Portugal and Brazilian translations would you mind updating those translations as well as part of this PR? (Or removing the fuzzy flag if they are already correct).

rffontenelle:
@carlosperate sure 🇧🇷👍

ntoll:
Hi folks.... many thanks for the work on this (and other i18n tickets). Just getting my GitHub juices going and merging what I can, pending a release very soon.

PR #1936, opened 2021-12-20
Fixes #1922 - fix string for better translatability.

### LINKED ISSUE #1922: " compatible" string is no good for translation

Translate Mu

I bootstrapped translation by running `make translate_begin LANG=pt_BR` and then started translating with my preferred PO editor.

After some strings, I hit `" compatible"` string. Its source location is in

https://github.com/mu-editor/mu/blob/12d80470add3f790a0a38cc559106a618773f3b3/mu/logic.py#L576

It is used in `def name`, so I believe it returns "SOME_BOARD compatible".

Be able to properly translate it.

### What actually happened?

The string `" compatible"` is impossible to be translated. In Portuguese, when something is FOOBAR-compatible, I have to translate reordering to something like "compatible with FOOBAR" (but in Portuguese, obviously). With the current partial string I can't do this correctly,

### Operating System Version

Ubuntu 20.04 LTS WSL2

### Mu Version

N/A

### Other Info

I suggest to have it rephrased with placeholders. E.g. `" {boardname} compatible"` 

### Editor Log

_No response_

carlosperate:
As discussed in https://github.com/mu-editor/mu/issues/1922, now that the two Portuguese translations are merged we can add this fix.

The PR diff is a bit larger because the string lines in the source code have changed with the latest code updates.

@rffontenelle I've changed the english string to `"{} compatible"` instead of `"{boardname} compatible"` in case some translators could think `boardname` also needs to be translated. Is a bit less context, but hopefully avoids that easy mistake.

@rffontenelle @xbecas it's a very easy string, but my Portuguese is quite rusty, so confirmation that it's correct would be great!


rffontenelle (approved):
LGTM

PR #1937, opened 2021-12-20
PR #1923 follow up, remove pt_BR fuzziness & add pt_PT string.

carlosperate:
@ntoll PR #1923 was merged before we could add the updated portuguese translations for the changed string.
The pt_BR string was marked and fuzzy, so I've updated it to remove that, and the pt_PT didn't have a translation for this string, so I've added the Brazilian one, as it looked about to right to me.


ntoll:
Apologies for getting click-y with the merge button. :scream:

carlosperate:
Haha, no worries, this way was quicker ;)

PR #1942, opened 2021-12-22
Delete zip file of wheels when make clean (avoids uploading the zip to PyPI

ntoll:
What the title says. :-) Ensures the Mu installable via `pip` is **small**.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ clean:
 	find . \( -name '*.tgz' -o -name dropin.cache \) -delete
 	find . | grep -E "(__pycache__)" | xargs rm -rf
 	rm -f ./mu/locale/messages.pot
+	rm -f ./mu/wheels/*.zip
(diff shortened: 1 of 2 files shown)


carlosperate (approved):
LGTM 👍

carlosperate:
Before this is merged, if the goal is to avoid having the wheels zip file in the whee or source distribution, do you also want to add it to the MANIFEST file?

ntoll:
The intention is to NOT have the wheels zip file in the wheel or source distribution, since the files needed will automatically be downloaded.

carlosperate:
> The intention is to NOT have the wheels zip file in the wheel or source distribution, since the files needed will automatically be downloaded.

Exactly, so my comment was about adding an entry into the MANIFEST file to exclude them from the packages in the case they are accidentally present during the build process.

PR #1943, opened 2021-12-22
micro:bit: Update board IDs to support V2.2

microbit-mark:
Add board IDs (9905, 9906) to support [micro:bit interface processor revision V2.2](https://support.microbit.org/support/solutions/articles/19000132336-bbc-micro-bit-v2-interface-processor-change-v2-2-)


carlosperate (approved):
Tested and works great, thanks Mark!

PR #1973, opened 2022-01-05
CI: Update the RaspPi Buster image to latest & add legacy release.

carlosperate:
The Buster "legacy" version (a maintenance-only version released after the Bullseye release) has several changes to the "standard" Raspberry Pi OS Buster, so in case that could affect Mu I've added the "legacy" version as a separate test run.

This PR also updates the "normal" Buster release image.

So "2021-05-28" becomes the last Buster release, and "2021-12-02" the latest buster legacy release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


carlosperate:
Merging this as it is an improvement to the CI.

Issue #1849, opened 2021-10-07
Status bar language not updated after change in user preference


aroberge:
### What were you trying to do?

Change the language for the UI

### What steps did you take to trigger the issue?

* Changed the language
* Closed Mu and restarted
* Observed the message shown very briefly in the status bar

### What did you expect to happen?

That the message would be shown in the newly selected language

### What actually happened?

It was not:
![image](https://user-images.githubusercontent.com/629698/136400232-71969125-c49c-4708-afdc-c54cd308a1d7.png)

### Operating System Version

Windows 10

### Mu Version

1.1.0.beta.6

### Other Info

This was originally mentioned in a comment for issue #976

devdanzin:
Confirmed. I think it comes from the fact the MOTD list is [populated/localized](https://github.com/mu-editor/mu/blob/master/mu/logic.py#L89) before we do a [`i18n.set_language(self.user_locale)`](https://github.com/mu-editor/mu/blob/master/mu/logic.py#L936). Then we pick a MOTD to display: [`self.show_status_message(random.choice(MOTD), 10)`](https://github.com/mu-editor/mu/blob/master/mu/logic.py#L952.)

It seems to be fixable by having a non-localized MOTD list, then localizing the selected entry:
```diff
-        self.show_status_message(random.choice(MOTD), 10)       
+        self.show_status_message(_(random.choice(MOTD)), 10)    
```

Working on a test and PR.

devdanzin:
I have two different simple fixes ready for this issue and can't decide which one is better. Indeed, it comes from building MOTD before figuring out `user_locale`.

Option 1, it can be solved by localizing a single entry of the MOTD list after `user_locale` is set (as described in previous comment). It removes `_()` (gettext) from a lot of messages, I fear it might be detrimental for the localization workflow.

Option 2, we create a  simple `get_MOTD` function so `_()` (gettext) remains in place but is only evaluated after  `user_locale` is set.

Maybe I should just create two PRs and we can decide, with code to reference?

xbecas:
Just to add that there are at least two other "places" where this happens:

The Modes Dialog and the comment on a blank file - both visible on the following image, where Portuguese sentences are shown with "English" explicitly chosen from language list. 

PS: my W10 OS is in Portuguese.

![image](https://user-images.githubusercontent.com/32622192/156883964-7bc32e7b-4c5b-494e-8ae3-27cbae6c99f5.png)

PR #2017, opened 2022-01-31
Provide a hack for proper locale initialization

CSharperMantle:
This PR is submitted in the hope to resolve mu-editor/mu#2008.

**Background.** The localized strings (especially strings wrapped by a function call of `_("some_str")`) in Mu are generally created and referenced in two ways: 1) via static class or global variables, and 2) via Qt widgets. Some examples of static variables is listed below.

https://github.com/mu-editor/mu/blob/5a5d7723405db588f67718a63a0ec0ecabebae33/mu/logic.py#L89-L101

https://github.com/mu-editor/mu/blob/5a5d7723405db588f67718a63a0ec0ecabebae33/mu/modes/base.py#L192-L210

The latter works perfectly when the locale is changed, as a restart of the app with a call to `editor.restore_session()` is enough to refresh the translations.

However the former one does not work in this way. This behavior is due to the fact that class and global variables are evaluated when their containing scripts are executed. This is well ahead of the time when `editor.restore_session()` refreshes the locale configuration in user settings. Thus these static strings are not loaded properly. This process can be summarized as follows.

1. `app.py` is evaluated. Modules are imported.
2. `logic.py`, `modes/*.py` and so on are evaluated.
3. Static variables are evaluated when the interpreter compiles their definitions. At this time, the locale settings are platform default, which is not consistent with user settings.
4. `app.run()` calls `editor.restore_session()`. The correct locale settings are refreshed into runtime environment.
5. The window is painted and other localized strings are loaded with the correct locale.
6. The inconsistency is observed between strings loaded from static variables and any other ways.

Note that the following lines DO NOT load localized strings as expected since `self.modes[self.mode].code_template` is evaluated when at step 3 aforementioned when the locale settings have not been refreshed from the user settings yet.

https://github.com/mu-editor/mu/blob/5a5d7723405db588f67718a63a0ec0ecabebae33/mu/logic.py#L956-L963

**Solutions.** In order to enforce a refresh before any localized strings are evaluated, the session restoring process is carried out before critical modules are loaded.

https://github.com/CSharperMantle/mu/blob/9a907af9164dee71370ffdd0db016a5325e018d1/mu/app.py#L48-L63

Note that this solution might not be the best one due to significant ugliness it causes. Other solutions might include 1) deferring call to `_("some_str")` to some point after the session is restored, or 2) load all session-related data before any UI-related scripts are evaluated.


xbecas:
Related to #1849 also!

PR #2021, opened 2022-02-02
Improve zh-CN translation

CSharperMantle:
This PR improves zh-CN translation.

**Changes.** These changes focus on correcting two mistakes:
1. Abusive uses of term "micro:bit" when it should be a general "device", and
2. Strange expressions that might confuse readers.


ntoll:
@CSharperMantle many many thanks for your work on the translation. We really appreciate it. I've just started the CI runs and notice that things are broken. Could you please fix the branch..? Perhaps `make check` before pushing changes so you can be sure all the tests are run locally on your machine before CI is run..? Thank you so much.

CSharperMantle:
Hi @ntoll ,

To make sure I am not going down the wrong path, may I ask why do CI workflows complain about “would reformat mu/contrib/uflash.py”? Is that the reason? If yes, I think I haven't changed it in any way.

ntoll:
Hmm.... I've just tried the tests on my local machine and everything passes. Could you perhaps rebase with master and I'll re-run the CI again..? Thank you.

ntoll:
I've rebased with master and created a new PR with your changes... see #2033

PR #2028, opened 2022-02-08
Fix CI due to new versions of black and pytest.

carlosperate:
This PR does two things:
- Set black version upper bound to <22.1.0.
    - Something with the default settings has changed and v22.1.0 makes some code changes that are not compatible with the same default settings from v19.10b0, which we still want to maintain compatibility.
- Fixes a pytest fixture
    - The first argument to skip was incorrectly set, this didn't seem to be a problem before, but pytest v7.0.0 marks the test as a failure.


ntoll:
Thank you!

PR #2033, opened 2022-02-12
Improve zh-CN translation

ntoll:
Based on #2021 but rebased with latest `master`


ntoll:
@CSharperMantle OK... that went green. Merging you work via this rebase. Thank you so much..!

PR #2058, opened 2022-02-22
venv: Add Flask subdependency upper bound limit.

carlosperate:
Currently Flask v1 depends on Jinja v2, which doesn't have an upper bound limit for MarkupSafe, and the latest release of MarkupSafe v2.1 is not compatible with Jinja v2. So we need to add the upperbound here.

https://github.com/pallets/jinja/issues/1585


davidism:
Would it be possible to update to Flask 2+ (and the latest versions of the other Pallets projects as well)? I'm not really clear what they're used for in Mu, but it seems like it might be some sort of project starter. If that's the case, I'd really prefer new users were introduced to supported versions of our libraries.

PR #2084, opened 2022-02-28
Cut off BaseMode workspace_dir extra call

xbecas:
Simplification of `mu/modes/base.py` by moving the logic of the `get_default_workspace()` function into the BaseMode.workspace_dir() method, as it was just redirecting to the above mentioned function.

Note: first very simple PR on real code (not just translation files) to test the whole process!

PS: test `test_ButtonBar_add_action` fails, but was failing before the changes.


xbecas:
Test `test_ButtonBar_add_action` failure could be related to SetupTools version, according to #2091.

xbecas:
> Test `test_ButtonBar_add_action` failure could be related to SetupTools version, according to #2091.

Fixed by downgrading setuptools from  60.9.3 to 56.0.0, using `pip install setuptools==56.0.0`. Now all tests, passed!

carlosperate (approved):
The CI failure is only due to an issue with the very old Pi OS stretch configuration, so it can be ignored.

Code LGTM and manually tested locally as well 👍

Thanks @xbecas!

Issue #2091, opened 2022-03-02
Mocking super() in tests breaks with setuptools 60.6+


hroncok:
### What were you trying to do?

Run mu tests with latest setuptools installed

### What steps did you take to trigger the issue?

We are rebuilding an older version of mu in Fedora with setuptools 60.9 as we are rebuilding everything. I suppose the newer version of mu has the same problem as the test has never changed.

I don't have an easy reproducer yet. I assume running the tests with setuptools 60.9 should be enough.

The CI here appears to run with an older setuptools version.

I can submit a PR that should try to update it and we can see if it triggers the same failure: https://github.com/mu-editor/mu/pull/2092

### What did you expect to happen?

The tests should pass

### What actually happened?

The tests fail with:

```
=================================== FAILURES ===================================
__________________________ test_ButtonBar_add_action ___________________________

    def test_ButtonBar_add_action():
        """
        Check the appropriately referenced QAction is created by a call to
        addAction.
        """
        bb = mu.interface.main.ButtonBar(None)
        with mock.patch('builtins.super') as mock_s:
>           bb.addAction('save', 'Save', 'save stuff')

tests/interface/test_main.py:119: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
mu/interface/main.py:121: in addAction
    action = QAction(load_icon(name), display_name, self,
mu/resources/__init__.py:37: in load_icon
    return QIcon(path(name))
mu/resources/__init__.py:32: in path
    return resource_filename(__name__, resource_dir + name)
/usr/lib/python3.10/site-packages/pkg_resources/__init__.py:1141: in resource_filename
    return get_provider(package_or_requirement).get_resource_filename(
/usr/lib/python3.10/site-packages/pkg_resources/__init__.py:360: in get_provider
    return _find_adapter(_provider_factories, loader)(module)
/usr/lib/python3.10/site-packages/pkg_resources/__init__.py:1591: in __init__
    self._setup_prefix()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pkg_resources.DefaultProvider object at 0x7f572f61fcd0>

    def _setup_prefix(self):
        # Assume that metadata may be nested inside a "basket"
[...]

PR #2092, opened 2022-03-02
Update setuptools on CI to verify it still works with the latest

hroncok:
Related to https://github.com/mu-editor/mu/issues/2091


carlosperate:
Thanks for the PR @hroncok! This has been fixed in https://github.com/mu-editor/mu/pull/2238, so I'll close this one.

hroncok:
It was fixed, but mu is still not tested with the latest setuptools on the CI, or is it?

hroncok:
Yes, they do. Nevertheless, this was not a breaking change and it broke the test anyway.

PR #2131, opened 2022-03-17
add pyproject.toml according to PEP 517

FantasqueX:
According to the [tutorial](https://peps.python.org/pep-0517/#source-trees) of setuptools and [PEP 517](https://peps.python.org/pep-0517/#source-trees), We should add pyproject.toml


FantasqueX:
Is there anyone to review this PR?

PR #2141, opened 2022-03-22
GitHub Action update: Linux Appimage packaging

tmontes:
MOTIVATION:

* After #1863 and #1874, this is the natural next step.

FURTHER WORK:

* We should really use an older Linux distribution here!


tmontes:
NOTES:

* Seems to work: https://youtu.be/3lubhcCAbOw.

NEXT:

* Can we package it under CentOS 7?

carlosperate (approved):
Let's merge it! 👍

PR #2216, opened 2022-05-09
WIP: Space in name

tjguk:
Trying to reproduce / test the frequent bug on Windows where it fails when there's a space in the username and a file which shadows the first element of the name (typically left over from a badly-behaved installer),

eg "c:\users\tim golden\...\mu" shadowed by "c:\users\tim" which is actually a log file


tjguk:
The intention is to use the Win32 [GetShortPathName](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getshortpathnamew) API for pip operations which will remove spaces from the name

tjguk:
The current state is that it's trying -- and failing -- to fail a test by injecting a path with a space

tjguk:
Trying very hard to get a reproducible here. One point is that it _only_ fails when the space is in the _username_ (ie the part immediately after `c:\users`), not in some other part of the path. Which makes it hard to reproduce, not least because that directly is admin-only, so not easy/possible to ad hoc create a "blocking" file.

tjguk:
Tried to work around by applying `GetShortPathName` to the pip executable: no joy. Still reports "Unable to run <full long name>" so, presumably Windows is LongName-ing it behind the scenes

tjguk:
Superseded by #2259

PR #2217, opened 2022-05-09
Clamp click version to workaround issue with black & other CI fixes

carlosperate:
Click is a black dependency and the latest releases (>8.0.4) have broken older black releases (<22.3.0).
https://github.com/psf/black/issues/2964

Ideally we should update black to the latest version, but right now we might  want to keep compatibility with older black releases.
Once we update all dependencies, including black, we can remove click.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [macos-10.15, windows-2016, ubuntu-18.04]
+        os: [macos-10.15, windows-2019, ubuntu-18.04]
       fail-fast: false
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


carlosperate:
This PR now also includes other fixes required to get all the CI runners working again:
- Updating the deprecated Windows 2016 images to 2019
- Updating the apt sources in Raspbian before attempting to install dependencies (before it was purposely using stale sources, but now they 404)

carlosperate:
The outstanding ci failures are due to this: https://github.com/mu-editor/mu/issues/2091
So we can merge this and fix that in a different PR (ideally by not mocking `super()`).

ntoll:
Great stuff. I'll let this run on `master` then use the resulting AppImage on our downloads page. Thank you! :tada:

PR #2237, opened 2022-05-20
venv: Use -I flag to isolate Jupyter kernel installation in venv.

carlosperate:
Otherwsie it might pick up a Jupyter package from a different Python installation and throw errors due to dependency versions missmatching.


carlosperate:
Tested as working in https://github.com/mu-editor/crash-reports/issues/13 🎉

PR #2238, opened 2022-05-20
Fix tests by reducing builtin mock from global scope to module scope.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/2091.

This should also get our CI back to passing, as it been red for a while.


carlosperate:
@ntoll If all tests pass we should merge this relatively soon get the CI back to green.

carlosperate:
CI passes so mergin this to unblock other PRs.

PR #2255, opened 2022-06-02
venv: Log pip version on venv creation, clarify empty output logging

carlosperate:
I find myself asking a lot of users with Crash Reports to run commands in the terminal to determine the pip version running inside the virtual environment, so this PR logs that to the venv creation process.

It also logs "No output received." whenever there is an empty output from venv creation/verification commands, as otherwise we end up with empty entries in the log, which are a bit hard to follow.

### DIFF
--- a/mu/virtual_environment.py
+++ b/mu/virtual_environment.py
@@ -44,7 +44,8 @@ class VirtualEnvironmentCreateError(VirtualEnvironmentError):
 
 def compact(text):
     """Remove double line spaces and anything else which might help"""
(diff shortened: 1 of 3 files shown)


tjguk (approved):
Looks fine on paper. Haven't cloned/run but I assume you have :)

carlosperate:
Yes, I've run it in a few different scenarios to make sure it didn't interfeer with anything. Thanks for the review! I'll merge 👍

Issue #2256, opened 2022-06-03
virtualenv.py: error: no such option: --symlinks


carlosperate:

carlosperate:
## Solutions

Simplest thing might be to update the `virtualenv` version listed in `setup.py`. We tried to keep this version as low as possible to maximise compatibility with systems that might have older versions from their OS package managers.

Version 16.7.6 was picked as the oldest version packaging a new-enough version of pip (v19.3), as described in this commit message: https://github.com/mu-editor/mu/commit/7966e3d780c47c94904c93569ef7c89ce335e3b2 

`virtualenv` v20.0.0 packs pip version 20.0.2, which might contain other useful updates:
- https://github.com/pypa/virtualenv/tree/20.0.0/src/virtualenv/seed/embed/wheels

For reference both Ubuntu 18.04 and Debian Buster use v15.1, so we already require a newer version:
- https://packages.debian.org/buster/python3-virtualenv
- https://packages.ubuntu.com/bionic/python3-virtualenv

carlosperate:
It is worth noting that Linux users that have installed Mu using pip in the system Python might also encounter this issue, even if during the `pip install mu-editor` the latest version of `virtualenv` is installed.

This is because Mu uses the Isolate `-I` flag and that will exclude the packages installed by pip, and leave the ones installed by the OS packages manager, which can be v15 in Ubuntu 18.04 and Debian/Raspbian Buster.

This issue, which can manifest in different ways, is captured in:
- https://github.com/mu-editor/mu/issues/1954

PR #2266, opened 2022-06-17
Update virtualenv depedency min version to v20 for --symlinks flag.

carlosperate:
The latest AppImage updates require the user virtual env in Linux to use symlinks, so that they can be recreated when the AppImage python excutable moves to a different temp folder.

The virtualenv --symlinks flag was only added in v20.0, so that's our required new min version.

More info and fixes https://github.com/mu-editor/mu/issues/2256.


carlosperate:
Discussed and reviewed during our team call, so good to merge 👍

Issue #2268, opened 2022-06-21
ESP MicroPython device incorrectly identfied as "Arduino Clone"


carlosperate:
As the CP210x USB VID and PID have been registed in the Snek Mode.

The large majority of the ESP boards (excluding the newer versions with built-in USB) have this USB-to-UART chip, so if we don't correct this they won't be detected as MicroPython boards.

PR #2269, opened 2022-06-21
Remove ESP-related USB VIDs & PIDs from other modes.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/2268.

Unfortunately at the moment there can only be a single mode associated to a board USB VID & PID for the autotection, otherwise it will pick the first mode it matches.

This is only for autodection, once a mode is selected, the specific functionality to run/flash code should still work even if these values are not in the mode `valid_boards` list.


carlosperate:
@keith-packard as these USB VID/PID values are very common for the USB-to-UART chips used in MicroPython ESP boards we'd like to keep them only for the ESP mode.
I've left the `usb_boards` and `reset_boards` matches still there, as `valid_boards` is only used for the auto detection. So when the Snek mode is selected I believe all the specific mode functionality should still work the same.

carlosperate:
Great, thanks for checking the PR, I'll merge this 👍

PR #2276, opened 2022-06-25
Don't use our own dmgbuild fork to package Mu on macOS.

tmontes:
CONTEXT:

* Our first attempts at `pup`-packaging Mu for macOS on CI lead to `dmgbuild` failures with messages like `hdiutil: detach: timeout for DiskArbitration expired`.
* Because of that, I forked `dmgbuild` to work around the issue.
* See #1234.

WHAT's CHANGED:

* `dmgbuild`'s recent releases claim to have fixed the issue.
* `pup`'s just released 1.0.0a17 now brings in that.

THUS:

* This PR cleans up the forced usage of my own `dmgbuild` fork when packaging on macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,10 +112,6 @@ macos: check
 	# 1. Not really needed.
 	# 2. Previously active venv would be "gone" on venv-pup deactivation.
 	./venv-pup/bin/pip install pup
-	# HACK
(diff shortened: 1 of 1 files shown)


tmontes:
NOTE:

* All chekcs ok.
* In particular, `pup` 1.0.0a17 with `dmgbuild` 1.5.2 behaviour while packaging for macOS looking good.

MERGING

PR #2288, opened 2022-07-16
replace appdirs with platformdirs

tonybaloney:
appdirs is no longer actively maintained, whereas platformdirs (a fork) is actively maintained and supported by tidelift


carlosperate (approved):
Thanks @tonybaloney, we didn't realise appdirs is no longer maintained 👍

PR #2292, opened 2022-07-16
Resolve the actual data path for windows distrubutions

tonybaloney:
Fixes a bug where if the host Python is from the Windows store then DATA_DIR will be different with subproceses. This causes a crash creating virtual environments.

See https://github.com/mu-editor/crash-reports/issues/228

### DIFF
--- a/mu/config.py
+++ b/mu/config.py
@@ -1,9 +1,26 @@
 import os
+import sys
+import tempfile
 
-import appdirs
+import platformdirs
+
+
+def _resolve_data_dir():
+    path = platformdirs.user_data_dir(appname="mu", appauthor="python")
+    if sys.platform == "win32":
+        # Locate the actual path for Windows by making a temporary file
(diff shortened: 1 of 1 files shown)


tonybaloney:
Fixes https://github.com/mu-editor/crash-reports/issues/229

tjguk:
@tonybaloney would you mind running a `make tidy` over the PR and re-push, please? I'd be very keen to get this in once it's passing tests.

tonybaloney:
@tjguk done (the failing test is failing on all PRs)

carlosperate:
> @tjguk done (the failing test is failing on all PRs)

Yes, the linux sementation fault used to be a lot more sporadic, now it seems to fail once in every CI run. We need to investigate further, but we haven't been able to replicate.
- https://github.com/mu-editor/mu/issues/1454

I've restarted the failing job, should pass the second time.

PR #2297, opened 2022-07-16
WIP: Prevent Mu from opening more than one instance

tjguk:
It's quite common for Mu to be slow enough to start that users grow impatient and start a second instance. This will then cause problems as both will try to, eg, create the same virtual environment.

This PR -- as a prototype -- uses the Qt Shared Memory wrapper to effect a Mutex which will prevent any further instance from opening while one is already running


tjguk:
Current state of play: Mu will detect that another instance is blocking it and will report the PID of the blocking instance. Also, all instance now log their PID with the logging header block to assist this (and other) debugging.

If a running instance is detected, the starting instance logs the fact and exits with error code `2`

Still to do: the shared memory approach we're using to act as an interprocess Mutex operates slightly differently on *nix vs Windows. We need to add some cleanup code on *nix; on WIndows by contrast the OS owns the shared memory and will clean up after itself.

mu/app.py
```diff
@@ -256,6 +258,36 @@ def setup_modes(editor, view):
...
+    def __exit__(self, *args, **kwargs):
+        self._shared_memory.unlock()
+
+    def acquire(self):
+        if self._shared_memory.attach():
+            pid = struct.unpack("l", self._shared_memory.data()[:4])
+            raise MutexError("MUTEX: Mu is already running with pid %d" % pid)
+        else:
+            self._shared_memory.create(8)
+            self._shared_memory.data()[:4] = struct.pack("l", os.getpid())
```
ZanderBrown, line 285: Perhaps being over-cautious: Are we sure we are going to fit a pid in 4?

tjguk: Well it's a good point. I did briefly check the Win implementation of `os.getpid` but I didn't check the rest. We can easily up it, so I'll probably do that.

tjguk: I've just gone down a rabbit-hole I'd rather not have trying to work out from the Posix spec. what type is defined in terms of what. After I'd eventually reached Cloud Cuckoo Land the only thing I was sure about was that the Posix `pid_t` is a signed integer "whatever that means" [near-enough quoting the spec]. So I've gone for an 8-byte Long Long.

ZanderBrown: If I'm reading my copy of 1003.1 2008 correctly,  `pid_t` cannot exceed the width of `long`, so on that basis `ctypes.sizeof(ctypes.c_long)` is probably the ‘correct’ answer - but 8 is probably a safe bet

mu/app.py
```diff
@@ -270,9 +302,16 @@ def run():
...
+    logging.info("Process id: {}".format(os.getpid()))
     logging.info("Platform: {}".format(platform.platform()))
     logging.info("Python path: {}".format(sys.path))
     logging.info("Language code: {}".format(i18n.language_code))
+    try:
+        _shared_memory.acquire()
+    except MutexError as exc:
+        [message] = exc.args
+        logging.error(message)
+        sys.exit()
```
carlosperate, line 314: Based on the description in https://github.com/mu-editor/mu/pull/2297#issuecomment-1186488744, should this be: 
```suggestion
        sys.exit(2)
```

mu/app.py
```diff
@@ -256,6 +258,36 @@ def setup_modes(editor, view):
     }
 
 
+class MutexError(BaseException):
+    pass
+
+
+class SharedMemory(object):
```
carlosperate, line 265: The name of this class might imply this is a generic way to share memory, should we maybe rename to something more indicative to this single instance lock mechanism?

carlosperate:
In macOS I get this exception
```
2022-07-20 10:54:39,781 - root:174(excepthook) ERROR: Unrecoverable error
  File "run.py", line 6, in <module>
    run()
  File "/Users/microbit-carlos/workspace/carlos/mu/mu/app.py", line 310, in run
    _shared_memory.acquire()
  File "/Users/microbit-carlos/workspace/carlos/mu/mu/app.py", line 285, in acquire
    self._shared_memory.data()[:4] = struct.pack("l", os.getpid())
ValueError: cannot modify the size of a sip.voidptr object
```

ntoll:
Any movement on this..?

carlosperate:
I gave this a quick run in macOS and worked as expected 👍

Issue #2291, opened 2022-07-16
Utils: Create a script to convert Python stub files into Qscintilla autocompletion files


carlosperate:

Jeffrey04:
so the goal is to turn
```
def panic() -> None:
    """
    Put micro:bit in panic() mode and display an unhappy face.
    Press the reset button to exit panic() mode.
    """
    pass
```

into the following?

```
_("panic() \nPut micro:bit in panic() mode and display an unhappy face. Press the reset button to exit panic() mode.")
```

What about instance methods?

```
class _Accelerometer(object):
    def get_x(self) -> int:
        """
        Return micro:bit's tilt (X acceleration) in milli-g's.
        """
        pass
```

carlosperate:
> so the goal is to turn
> 
> ```
> def panic() -> None:
>     """
>     Put micro:bit in panic() mode and display an unhappy face.
>     Press the reset button to exit panic() mode.
>     """
>     pass
> ```
> 
> into the following?
> 
> ```
> _("panic() \nPut micro:bit in panic() mode and display an unhappy face. Press the reset button to exit panic() mode.")
> ```
> 

Yes 👍 


> What about instance methods?
> 
> ```
> class _Accelerometer(object):
>     def get_x(self) -> int:
>         """
>         Return micro:bit's tilt (X acceleration) in milli-g's.
>         """
>         pass
> ```

For the micro:bit API, classes for built in board features have a single instance by default. So the user cannot initialise a new instance of the micro:bit accelerometer class. In this case it looks like the API just follows a similar pattern a module functions:
https://github.com/mu-editor/mu/blob/9fe47acf97096649d94cd26b379247f73e74d6ee/mu/modes/api/microbit.py#L76

For thinsg like MicroPython PyBoard I2C calsses (which you can initilisate) it looks very similar:
https://github.com/mu-editor/mu/blob/9fe47acf97096649d94cd26b379247f73e74d6ee/mu/modes/api/pyboard.py#L175

Jeffrey04:
let me attempt to submit a PR then

PR #2301, opened 2022-07-17
2291 proof-of-concept to turn toy example into stub

Jeffrey04:
refs: #2291 

while it works for a toy example, but it doesn't really work with the stub linked in the original issue, am still trying to figure out a better way to do it

### DIFF
--- a/utils/completion_convert.py
+++ b/utils/completion_convert.py
@@ -1,0 +1,39 @@
+"""
+GitHub Issue #2291
+Convert python stub file into Qscintilla autocompletion files
+"""
+
+import importlib.util
+import sys
+from inspect import getmembers, isclass, isfunction, signature
+
+spec = importlib.util.spec_from_file_location("module", sys.argv[1])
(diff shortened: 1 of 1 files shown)


Jeffrey04:
I suppose I should have used `ast` to parse the stub file

Jeffrey04:
am aware of the code only work for one file for now, working on traversing the directory next
(is there a way to not trigger a build until this PR is ready?)

Jeffrey04:
for https://github.com/oivron/microbit-stubs/blob/master/microbit/lib/audio.py

script current returns a file

```
MICROBIT_APIS = [_("play(source, wait, pins) Play the source to completion where 'source' is an iterable, each element of which must be an AudioFrame instance."), _("is_playing() Return True if audio is playing, otherwise return False."), _("stop() Stops all audio playback.")]
```

Jeffrey04:
code should now generate the autocompletion guide, it is hackish, and most certainly require a lot of polishing, please let me know if I missed out anything that require fixing

PR #2311, opened 2022-07-19
Update index.rst to fix rendering error

tonybaloney:
Code lines are not rendering correctly because of invalid rST syntax


carlosperate (approved):
Thanks!

PR #2314, opened 2022-07-19
Set QT_QPA_PLATFORM=wayland env var when running on wayland.

carlosperate:
To see if that lets the AppImage launch correctly.
Not sure if this will work, will try in an Ubuntu 22.04 VM once the CI builds it.

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -256,6 +256,18 @@ def setup_modes(editor, view):
     }
 
 
+def is_linux_wayland():
+    """
+    Checks environmental variables to try to determine if Mu is running on
+    wayland.
+    """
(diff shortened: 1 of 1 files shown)


carlosperate:
Tested this in Ubuntu 22.04, and it removes the need to launch the AppImage with the `QT_QPA_PLATFORM=wayland` env var, as currently needed:
https://github.com/mu-editor/mu-editor.github.io/pull/70

Once we release v1.2.0 we can remove that step from the docs.

carlosperate:
@ntoll this should be included in the next release so that the AppImages run correctly under Wayland, so I'l merge this now.

Issue #2322, opened 2022-08-04
Test failures with updated flake8 / pyflakes


hroncok:
### What were you trying to do?

Build mu in Fedora

### What steps did you take to trigger the issue?

flake8 was updated to 5.0.3 / pyflakes to 2.5.0

### What did you expect to happen?

The tests should pass

### What actually happened?

The tests fail. They also do on the CI here, e.g. in https://github.com/mu-editor/mu/pull/2281

```
=================================== FAILURES ===================================
__________________ test_MuFlakeCodeReporter_flake_real_output __________________

    def test_MuFlakeCodeReporter_flake_real_output():
        """
        Check the reporter handles real output from flake, to catch format
        change regressions.
        """
        check = mu.logic.check
        reporter = mu.logic.MuFlakeCodeReporter()
        code = "a = 1\nb = 2\nc\n"
        check(code, "filename", reporter)
>       assert reporter.log[0]["line_no"] == 2
E       assert 0 == 2
E         +0
E         -2

tests/test_logic.py:638: AssertionError
__________________ test_check_real_flake_output_with_builtins __________________

    def test_check_real_flake_output_with_builtins():
        """
        Check that passing builtins correctly suppresses undefined name errors
        using real .check_flake() output.
        """
        ok_result = mu.logic.check_flake("foo.py", "print(foo)", builtins=["foo"])
>       assert ok_result == {}
E       assert {0: [{'column...name 'foo'"}]} == {}
E         Left contains 1 more item:
E         {0: [{'column': 0,
E               'line_no': 0,
E               'message': "foo.py:1:7: undefined name 'foo'"}]}
E         Full diff:
E           {
E         -  ,
E         +  0: [{'column': 0,
E         +       'line_no': 0,
E         +       'message': "foo.py:1:7: undefined name 'foo'"}],
E           }

tests/test_logic.py:519: AssertionError
```

### Operating System Version

all

### Mu Version

git HEAD and 1.1.1

carlosperate:
Thanks @hroncok, I've created PR https://github.com/mu-editor/mu/pull/2323 to fix this.

PR #2323, opened 2022-08-09
Update pyflakes regex to match pre and post v2.5.0 release.

carlosperate:
pyflakes 2.5.0 has unified their output fomatting:
- https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst
- https://github.com/PyCQA/pyflakes/pull/693

So, Mu's regex has to match pyflakes messages like this:
- pre-2.5.0  -> `filename:line:char message`
- post-2.5.0 -> `filename:line:char: message`

And so `:?` has been added to the regex for an optional `:` after the `char` value.

Fixes https://github.com/mu-editor/mu/issues/2322.


hroncok (approved):
The tests pass with pyflakes 2.5.0.

carlosperate:
Thanks for testing @hroncok, I'll merge this now to fix CI.

PR #2327, opened 2022-08-10
Use the usb description for CP board name

tannewt:
It appears this is the product name on Linux at least.

### DIFF
--- a/mu/modes/circuitpython.py
+++ b/mu/modes/circuitpython.py
@@ -302,7 +302,7 @@ class CircuitPythonMode(MicroPythonMode):
                     manufacturer,
                     self.name,
                     self.short_name,
-                    "CircuitPython board",
+                    port.description(),
                 )
         # No match.
         return None


carlosperate:
Thanks for the PR @tannewt!

PR https://github.com/mu-editor/mu/pull/2323 has been merged, so rebasing or merging main should resolve the CI failures.

Has this been tested in Windows and macOS?

Looks like the default value for `ListPortInfo.description` might be `n/a`:
https://github.com/pyserial/pyserial/blob/v3.5/serial/tools/list_ports_common.py#L41
Do all CircuitPython devices provide the USB data for this? In that case it doesn't matter that much, but if some don't then we might want to check for the default value and replace it with the standard `CircuitPython board` name?

Another thing to keep in mind is that `ListPortInfo.description` could also contain the `ListPortInfo.interface` string as part of it:
https://github.com/pyserial/pyserial/blob/31fa4807d73ed4eb9891a88a15817b439c4eea2d/serial/tools/list_ports_common.py#L74
https://github.com/pyserial/pyserial/blob/v3.5/serial/tools/list_ports_common.py#L55-L58
I assume `ListPortInfo.interface` is None for CircuitPython devices? But if this changes in the future it will end up adding that info here.

dhalbert:
> I assume `ListPortInfo.interface` is None for CircuitPython devices? But if this changes in the future it will end up adding that info here.

It's not `None`, it's `'CircuitPython CDC Control'`. That's how mu identifies the board as a CircuitPython board, using [Adafruit_Board_Toolkit](https://github.com/adafruit/Adafruit_Board_Toolkit). See mu #1371, where I added that capability instead of using VID/PID. But maybe you know that already.

Stock `pyserial` doesn't do this right on Windows or earlier versions of macOS. We have two unmerged PR's to pyserial to fix this:
https://github.com/pyserial/pyserial/pull/566
https://github.com/pyserial/pyserial/pull/571
Hence need for Adafruit_Board_Toolkit, which incorporates the code in those pyserial PR's.

> Has this been tested in Windows and macOS?

> Another thing to keep in mind is that ListPortInfo.description could also contain the ListPortInfo.interface string as part of it:

I tested on all three OS's. I see that `description` varies on the three OS's. Windows seems to substitute its own idea, but on the others, description includes interface. It looks like macOS 12.5 now doesn't need the special macOS workaround in Adafruit_Board_Toolkit. (@Neradoc this will be of interest to you.)

Stock `pyserial` is still wrong on Windows.

(transcripts below are a bit elided and don't show me tracking down the correct port index from multiple ports)

**Ubuntu 22.04**
```py
# adafruit_board_toolkit works
>>> import adafruit_board_toolkit.circuitpython_serial
>>> ports = adafruit_board_toolkit.circuitpython_serial.comports()
>>> ports[0].interface
'CircuitPython CDC control'
>>> ports[0].description
'Metro M4 Express - CircuitPython CDC control'

# pyserial works
>>> import serial
>>> import serial.tools.list_ports
>>> ports = serial.tools.list_ports.comports()
>>> ports[2].interface
'CircuitPython CDC control'
>>> ports[2].description
'Metro M4 Express - CircuitPython CDC control'
```

**macOS 12.5 on an Intel machine**
```py
# adafruit_board_toolkit works
>>> import adafruit_board_toolkit.circuitpython_serial
>>> ports = adafruit_board_toolkit.circuitpython_serial.comports()
>>> ports[0].interface
'CircuitPython CDC data'
>>> ports[0].description
'Metro M4 Express - CircuitPython CDC data'

# pyserial works now (didn't used to)
>>> import serial.tools.list_ports
>>> ports = serial.tools.list_ports.comports()
>>> ports[0].interface
'CircuitPython CDC data'
>>> ports[0].description
'Metro M4 Express - CircuitPython CDC data'
```

**Windows 10 21H2**
```py
# Windows substitutes its own description
>>> import adafruit_board_toolkit.circuitpython_serial
>>> ports =adafruit_board_toolkit.circuitpython_serial.comports()
>>> ports[0].interface
'CircuitPython CDC control'
>>> ports[0].description
'USB Serial Device (COM8)'

# pyserial can't get interface and gives its own description
>>> import serial
>>> import serial.tools.list_ports
>>> ports = serial.tools.list_ports.comports()
>>> ports[1].interface
>>> ports[1].description
'USB Serial Device (COM8)'
```

Neradoc:
I think it would be better to use `.product` followed by `REPL` or `data` based on the value of `.interface`.
Or fallback to `.description` if you don't get something reasonable from those.

On windows the product is missing, and the description is generic garbage (but localized !)
```
>>> ports[0].product
>>> ports[0].description
'Périphérique série USB (COM5)'
```

In discotool I found that the WMI interface truncates the USB product field (to 16 characters I think), so I use this code to dive into the low level Windows USB API and list the USB devices with the proper full name. Anyone knowing more about those APIs could probably make a simpler version... I'd take it.
https://github.com/Neradoc/discotool/blob/main/discotool/usbinfos/usb_descriptor_win32.py

Neradoc:
I believe I tried to add the product to the pyserial code by poking around with ctypes where it gets the description, and failed, but then again I don't know those APIs. My conclusion was that the APIs used by pyserial only relate to COM ports and don't have that information, so other APIs have to be used. The fact that the USB name is not available in Device Manager seemed to confirm that.

So I found that code that uses `pywin32` to list the USB devices and pull the actual fields from them and adapted it.

tannewt:
I only tested on Arch Linux.

This serial info isn't coming from pyserial directly. It is a [QSerialPortInfo](https://doc.qt.io/qt-5.15/qserialportinfo.html). The description for me and my test board is "ESP32-S3-USB-OTG-N8" and does not contain the interface name.

Neradoc:
Oh, `port` is not the comport from `adafruit_board_toolkit` 🤦 Well I misread that...
I'm really curious what it tells us on windows then.

Neradoc:
Oh well, that's what we get on windows. 😞 

mu-windows

tannewt:
@dhalbert Please finish this for me.

ntoll:
Love this.

Great delegation there @tannewt. :stuck_out_tongue: I think you missed `sudo` in your comment (context: https://xkcd.com/149/)

Big hugs to all Adafruits in this chat, and many thanks for your continued contributions and support.

aivarannamaa:
https://github.com/adafruit/Adafruit_Board_Toolkit/issues/11 explains how you could get the proper `product` value for a serial port in Windows.

PR #2346, opened 2022-10-25
Fix `None` being lowercase in checker messages

Jayman2000:
Consider the following expression:

```python
foo == None
```

Before this change, Mu’s checker would tell you “Comparison to none should be 'if cond is none:'” if it encountered that expression. The word “none” in that message should be capitalized because it refers to something that should be capitalized in Python code.

The problem was caused by an inappropriate use of `str.capitalize()`. The author probably just wanted to make the first letter of the message uppercase. [Unfortunately, `str.capitalize()` also makes every other letter in the string lowercase.][1]

[1]: <https://docs.python.org/3.8/library/stdtypes.html#str.capitalize>

### DIFF
--- a/mu/logic.py
+++ b/mu/logic.py
@@ -470,11 +470,14 @@ def check_pycodestyle(code, config_file=False):
                 description += _(" above this line")
             if line_no not in style_feedback:
                 style_feedback[line_no] = []
+            # Make sure that description starts with a capital letter.
+            if description:
+                description = description[0].upper() + description[1:]
             style_feedback[line_no].append(
                 {
                     "line_no": line_no,
                     "column": int(col) - 1,
-                    "message": description.capitalize(),
+                    "message": description,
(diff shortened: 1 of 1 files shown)


mu/logic.py
```diff
@@ -470,11 +470,14 @@ def check_pycodestyle(code, config_file=False):
                 description += _(" above this line")
             if line_no not in style_feedback:
                 style_feedback[line_no] = []
+            # Make sure that description starts with a capital letter.
```
carlosperate, line 473:
```suggestion
            # Capitalise the 1st letter keeping the rest of the str unmodified
```

carlosperate (approved):
Thanks @Jayman2000! The PR looks good to me, I've mofieid the comment to add a bit more context to clarify why `capitalize` is not a good fit.
Once the tests pass I'll merge 👍

Issue #2120, opened 2022-03-14
Packaged Numpy wheel in macOS installer is only OS X 10.14+ compatible


carlosperate:
## Problem

The main Mu dependency restricting macOS compatibility is PyQt5. As we are currently using v5.13, so it should be compatible with macOS Sierra 10.12 and newer: https://doc.qt.io/archives/qt-5.13/macos.html

Currently the Mu installers download a numpy wheel for the user venv as a dependency for pygamezero and qtconsole.

With Python 3.8 in a modern macOS pip will download the latest Numpy which is only compatible wit macOS 10.14 Mojave and newer. So users with older OS versions are seeing errors like this one:

```
2022-03-13 16:03:57,169 - root:171(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/app.py", line 159, in run
    raise ex
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/app.py", line 146, in run
    venv.ensure_and_create(self.display_text)
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 603, in ensure_and_create
    self.create()
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 772, in create
    self.install_baseline_packages()
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 885, in install_baseline_packages
    self.install_from_zipped_wheels(zipped_wheels_filepath)
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 860, in install_from_zipped_wheels
    self.pip.install(wheel, deps=False, index=False)
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 270, in install
    return self.run(
  File "/Applications/Mu Editor.app/Contents/Resources/Python/lib/python3.8/site-packages/mu/virtual_environment.py", line 246, in run
    result = self.process.run_blocking(
[...]

Issue #1952, opened 2021-12-24
Avoid installing cryptography (for esptool) module in the user venv


carlosperate:
It's been a major pain from the beginning, mostly due to needing newer version of pip to be installed correctly or to be built.
In many cases it shows as a "error installing esptool" in the crash report and can only be tracked to the `cryptography` module via the log file, which we often don't get.

It's not even used by the main esptool script we use in Mu, as it is used for encrypted firmware flashing, which we don't do.

A few possible workarounds:
- Fork esptool to remove the things we don't need.
     - This should be relatively easy because all we use is encapsulated in a single python file that only needs the standard library and pyserial
- Vendor the esptool script with Mu.
    - We should be able to copy a single file into this repo and have it working
- Lock esptool to a known version and install it without dependencies
   - We would have to also add pyserial to the venv list, but I think that should be enough as well to get it working 

Crashes due to this:
- https://github.com/mu-editor/mu/issues/1965
- https://github.com/mu-editor/mu/issues/1951
- https://github.com/mu-editor/mu/issues/1950
- https://github.com/mu-editor/crash-reports/issues/83
- https://github.com/mu-editor/mu/issues/1941
- https://github.com/mu-editor/crash-reports/issues/84
- https://github.com/mu-editor/mu/issues/1827
- https://github.com/mu-editor/mu/issues/1823
- And at this point I stopped looking as these issues are only for the last 3 months and there are quite a few more since the introduction of the crash reporter

carlosperate:
This has been fixed in  #2352 and released in v1.2.0 🎉

PR #2352, opened 2022-11-10
Move esptool from user-venv dependency to vendor it with Mu.

carlosperate:
As discussed in:
- https://github.com/mu-editor/mu/issues/2120#issuecomment-1309468274

Only a single script was needed from the entire esptool package, and the only dependency needed for that script is pyserial, which is already included with Mu.
So, vendoring esptool means we don't need to pip download, package, and install its other dependencies, which have caused trouble in the past (specially cryptography when it moved to be a rust package, as discussed in https://github.com/mu-editor/mu/issues/1952).

This changes the "category" of this dependency, from a wheel packaged with Mu to install in the user venv, to be included as a Mu app runtime dependency. As esptool is used in the Mu settings, is probably closer to a "Mu runtime" dependency, than a "user running their code in a venv" dependency.

Apart from that, esptool and some of its dependencies don't have wheels in PyPI.
To be able to use the `--platform` flag with `pip download` it needs all dependencies to have wheels, so removing this from the user wheels helps with that as well.

### DIFF
--- a/mu/interface/dialogs.py
+++ b/mu/interface/dialogs.py
@@ -18,8 +18,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import os
 import logging
-from PyQt5.QtCore import QSize, QProcess, QTimer, Qt
+import sys
 
+from PyQt5.QtCore import QSize, QProcess, QTimer, Qt
 from PyQt5.QtWidgets import (
     QHBoxLayout,
     QVBoxLayout,
@@ -427,13 +428,6 @@ class ESPFirmwareFlasherWidget(QWidget):
 
         self.mode = mode
 
-    def esptool_is_installed(self):
-        """
-        Is the 'esptool' package installed?
-        """
-        baseline, user_packages = venv.installed_packages()
-        return "esptool" in user_packages
-
     def show_folder_dialog(self):
         # open dialog and set to foldername
         filename = QFileDialog.getOpenFileName(
@@ -460,9 +454,9 @@ class ESPFirmwareFlasherWidget(QWidget):
         if device is None:
             return
 
-        esptool = "-mesptool"
+        esptool = "-mmu.contrib.esptool"
         erase_command = '"{}" "{}" --port {} erase_flash'.format(
-            venv.interpreter, esptool, device.port
+            sys.executable, esptool, device.port
         )
 
         if self.device_type.currentText() == "ESP32":
@@ -470,7 +464,7 @@ class ESPFirmwareFlasherWidget(QWidget):
                 '"{}" "{}" --chip esp32 --port {} --baud {} '
                 'write_flash -z 0x1000 "{}"'
             ).format(
-                venv.interpreter,
+                sys.executable,
                 esptool,
                 device.port,
                 baudrate,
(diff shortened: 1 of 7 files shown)


carlosperate:
I haven't tested this properly yet, but it is added here for visibility and review.

mu/interface/dialogs.py
```diff
@@ -460,17 +454,17 @@ def update_firmware(self):
...
+            sys.executable, esptool, device.port
         )
 
         if self.device_type.currentText() == "ESP32":
             write_command = (
                 '"{}" "{}" --chip esp32 --port {} --baud {} '
                 'write_flash -z 0x1000 "{}"'
             ).format(
-                venv.interpreter,
+                sys.executable,
```
ntoll, line 467: Why the change from `venv.interpreter` to `sys.executable`..?

tests/interface/test_dialogs.py
```diff
@@ -178,11 +178,7 @@ def microbit():
     return device
 
 
-@mock.patch(
-    "mu.interface.dialogs.ESPFirmwareFlasherWidget.esptool_is_installed",
-    return_value=True,
-)
-def test_ESPFirmwareFlasherWidget_setup(esptool_is_installed, microbit):
+def test_ESPFirmwareFlasherWidget_setup(microbit):
```
ntoll, line 181: Nice to see mocks are gone.

ntoll (approved):
I wonder if we can exclude `esptool.py` from LGTM's analysis too.

mu/interface/dialogs.py
carlosperate: esptool used to be run from the user venv, which would have `pyserial` installed as part of installing `esptool`.
Now as we vendor it inside the Mu source code we need to execute it in an environemnt that has `pyserial` installed (pyserial is no longer installed in the user venv), so we now have to use `sys.executable` which points to the python executable+environment that is running Mu, which will have `pyserial`.

tests/interface/test_dialogs.py
carlosperate: Yeah, `esptool_is_installed()` wasn't really used anywhere, might have been accidentally left after a refactor at some point.

carlosperate:
Tested this in a couple of ESP8266 devices and works as expected.

mu/interface/dialogs.py
ntoll: Of course. Thank you for the clarification.

carlosperate:
Since Semmle was adquire by GitHub and LGTM will be shut down in December, we might as well move to the current way to run it:
- https://github.com/mu-editor/mu/pull/2347

And how to exclude files:
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan

I'll merge the PR here and we can merge it together to the main branch.

carlosperate:
Okay, it's now configured to ignore the esptool script and the warnings went down to 27:
image

I think this PR is ready to be merged 🎉

carlosperate:
@ntoll As now we are using CodeQL directly from GitHub I guess we should deactivate LGTM, which I assume you'd have to do from their website?

ntoll:
@carlosperate yeah... lemme try to remember how to do that... :rofl:

carlosperate:
Great, thanks! I'll merge this as done then 👍

Issue #2318, opened 2022-07-23
AppImage Debian Bullseye / Lubuntu 22.04 w/ X11  "Segmentation fault"


cpina:
### What were you trying to do?

Use Mu-Editor

### What steps did you take to trigger the issue?

I followed the steps from https://codewith.mu/en/howto/1.1/install_linux

### What did you expect to happen?

mu-editor would run

### What actually happened?

I got a segmentation falt

### Operating System Version

Debian Bullseye 11.4 and Lubuntu 22.04

### Mu Version

1.1.1

### Other Info

```
carles@pinux:~$ ./Mu_Editor-1.1.1-x86_64.AppImage
No settings file found at /home/carles/.local/share/mu/venv.json; skipping
Segmentation fault
carles@pinux:~$ QT_QPA_PLATFORM=wayland ./Mu_Editor-1.1.1-x86_64.AppImage
No settings file found at /home/carles/.local/share/mu/venv.json; skipping
Failed to create wl_display (No such file or directory)
qt.qpa.plugin: Could not load the Qt platform plugin "wayland" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Aborted
carles@pinux:~$ 
```

For the Debian: `apt install mu-editor` worked. For the LUbuntu I then had to apply the patch from https://github.com/mu-editor/mu/issues/2228

I'm happy to follow up any debugging for the Debian Bullseye. I will not have (easy) access to the LUbuntu after tomorrow's (will be with the kids that will use it).

Both the Debian and LUbuntu installations use X11. Debian with lxde and LUbuntu with lxqt.

As said: no work around needed for me, reporting only so the projects know. Thanks very much for writing mu-editor! :-)

rp10007:
Same for regular Ubuntu 22.04, with same error message. This is a breaking problem for me. I will roll back to a previous version.

carlosperate:
The installation instructions were updated a couple of weeks ago:
https://codewith.mu/en/howto/1.1/install_linux

Have you tried `sudo apt install libfuse2`?

cpina:
`libfuse2` was already installed:
```
carles@pinux:~$ dpkg -l | grep -i libfuse2
[...]

PR #2357, opened 2022-11-16
packaging: Install pup from the latest commit on GitHub to use Py 3.8.13

carlosperate:
To test if that resolves https://github.com/mu-editor/mu/issues/2318 in the AppImage that CI will build.


carlosperate:
It has removed the warnings and all looks good, so I'll merge now.

PR #2359, opened 2022-11-16
CI: Update version of used Actions to remove warnings.

carlosperate:
The majority of the 50 warnings we currently get are:
- Node.js 12 actions are deprecated
- The `set-output` command is deprecated and will be disabled soon


carlosperate:
It has remove the warning and all looks good, so I'll merge now.

PR #2362, opened 2022-12-04
make: Add Makefile target to build Linux AppImage with Docker.

carlosperate:
It's beneficial to create the Linux AppImages within an old version of a linux distro, mostly to make sure it has compatibility with older glibc releases.

So this PR adds a target to create an AppImage using docker and the Mu `ghcr.io/mu-editor/mu-appimage` docker image: https://github.com/mu-editor/docker-mu-appimage
```
make linux-docker
```

Theoretically it should work in all operating systems, although I have only tried it so far in macOS.

@ntoll it'd be great if you could give this a spin as well to confirm it works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,6 +131,24 @@ linux: check
 	ls -la ./build/pup/
 	ls -la ./dist/
 
+linux-docker: clean
+	@echo "\nFetching wheels."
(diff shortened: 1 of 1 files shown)


carlosperate:
@ntoll I think you've test this and it worked for you? If so, are we good to merge this PR?

carlosperate:
Tested by @ntoll and ready to be merged.

PR #2363, opened 2022-12-05
CI: Exclude Python 3.5 and 3.6 from running in Ubuntu-latest runner.

carlosperate:
As these Python versions are not supported in 22.04 and the Action step fails to install them.
This PR only skips them for `ubuntu-latest`, the Ubuntu 18.04 runner is still running the tests with 3.5 & 3.6


carlosperate:
Merging this now to fix the Ubuntu CI failures.

PR #2364, opened 2022-12-05
Update util to retrieve GH download stats with more info.

carlosperate:
To print the releases from old to new (so that the latest release appears at the bottom of the print out, inmediately visible without having to scroll up), and to include the length of time between releases.


carlosperate:
Okay, this is a update to a small utility script that is not used by Mu itself, so as there is almost no risk I'll go ahead and merge this.

Issue #2365, opened 2022-12-06
Building macOS installer fails 


carlosperate:
pup throws the following error:
```I Step 'mac.create-dmg': starting.
I dmgbuild err: Traceback (most recent call last):
I dmgbuild err:   File "/Users/carlos/workspace/carlos/mu/venv-pup/bin/dmgbuild", line 5, in <module>
I dmgbuild err:     from dmgbuild.__main__ import main
I dmgbuild err:   File "/Users/carlos/workspace/carlos/mu/venv-pup/lib/python3.7/site-packages/dmgbuild/__init__.py", line 1, in <module>
I dmgbuild err:     from .core import build_dmg
I dmgbuild err:   File "/Users/carlos/workspace/carlos/mu/venv-pup/lib/python3.7/site-packages/dmgbuild/core.py", line 42, in <module>
I dmgbuild err:     from mac_alias import *
I dmgbuild err: AttributeError: module 'mac_alias' has no attribute 'kBookmarkVolumePathkBookmarkVolumeURL'
E [Errno 2] No such file or directory: '/Users/carlos/workspace/carlos/mu/build/pup/Mu Editor.dmgbuild/Mu Editor 1.2.0.dmg'
I Step 'mac.create-dmg': completed.
```
This comes from the `mac_alias` package, which is installed as a depedency of `dmgbuild`, which is a depedency of `pup`.

The issue is already identified, and a PR raised, but it hasn't been merged and released yet:
- https://github.com/dmgbuild/mac_alias/issues/21
- https://github.com/dmgbuild/mac_alias/pull/20

Possible workarounds right now:
a. Update pup to pin `mac_alias==2.2.0`
b. Install `mac_alias==2.2.0` in the pup virtual env, before installing pup

Option a) might take a little while, as Tiago is likely busy at the moment and might not have the bandwidth to do a release soon (@tmontes tagging you for visibility, but no need to look into this right now).

I'll prepare a PR for option b), which we can remove when the next version of `mac_alias` is released.

carlosperate:
I've merge PR https://github.com/mu-editor/mu/pull/2366 with a temporary workaround, and responded to the `mac_alias` issue to keep an eye on the resolution. So I'll leave this issue open as a reminder to remove the workaround once there is a new `mac_alias` release.

carlosperate:
The `mac-alias` pup dependency was updated and the workaround has been removed 🎉

PR #2369, opened 2022-12-07
Remove temp workaround for macOS installer builds.

carlosperate:
As discussed in https://github.com/mu-editor/mu/issues/2365.

The `mac-alias` pup dependency has been updated with the fix, so we can remove the workaround 🎉


carlosperate:
The CI passes and I've tested the built dmg, which also works correctly, so we can merge this PR to remove the workaround 👍

Issue #2375, opened 2022-12-25
New `check_only_running_once()` call failing on macOS 10.13.6; test fails also.


haikusw:
### What were you trying to do?

build and run mu on macOS 10.13.6 using python 3.8.10

### What steps did you take to trigger the issue?

// setup pyenv veritual environment & activate
// set it to use python 3.8.10 (latest 3.8)
// git clone https://github.com/mu-editor/mu.git
// do dev setup `pip install -e ".[dev]"` etc.

`make run`
fails

`make test` fails one test relating to `check_only_running_once()`

Checking out the head -1 commit results in running successfully every time and all tests passing.

### What did you expect to happen?

run successfully, all tests pass

### What actually happened?

`make run`

results in:

> (venv) haikumintthin:mu tyler$ make run
> rm -rf build
> rm -rf dist
> rm -rf .coverage
> rm -rf .eggs
> rm -rf *.egg-info
> rm -rf docs/_build
> rm -rf .pytest_cache
> rm -rf lib
> rm -rf *.mp4
> rm -rf .git/avatar/*
> rm -rf venv-pup
> find . \( -name '*.py[co]' -o -name dropin.cache \) -delete
> find . \( -name '*.bak' -o -name dropin.cache \) -delete
> find . \( -name '*.tgz' -o -name dropin.cache \) -delete
> find . | grep -E "(__pycache__)" | xargs rm -rf
> rm -f ./mu/locale/messages.pot
> rm -f ./mu/wheels/*.zip
> python run.py
> make: *** [run] Error 2
> 

----

`make test` results in:

> ... <snip tons of output>

> ======================================================================================== FAILURES =========================================================================================
> _________________________________________________________________________________ test_only_running_once __________________________________________________________________________________
> 
>     def check_only_running_once():
>         """If the application is already running log the error and exit"""
>         try:
>             with _shared_memory:
> >               _shared_memory.acquire()
> 
> mu/app.py:316: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> 
[...]

Issue #2315, opened 2022-07-21
Mu 1.1.1 on macOS 12.4 (M1):  min window height & width seems to be 50% of monitor. Way too wide on 49" wide screen monitor.


haikusw:
### What were you trying to do?

Run Mu for first time.

### What steps did you take to trigger the issue?

Launch Mu and it opens a window that is way way way wider than it needs to be.
Tried to resize it smaller - no can do!

Play with resizing height and width and determine that it appears the minimum width and height were set to be 50% of the screen height and width (an unfortunate assumption) instead of basing it on the window content (toolbar items or whatever) or just allowing the user to indicate how small _they_ want the window to be.

***Update:  comment here confirms visual idea that it was 50% https://github.com/mu-editor/mu/issues/2315#issuecomment-1191922836***

### What did you expect to happen?

I should be able to resize the window down to as small as I, the user who should be in charge, wish to.

### What actually happened?

On a 49" widescreen monitor this makes for a window with a max width of roughly 2560 pixels wide when the toolbar icons need only 1040 pixels to fit all their content.

From a UX perspective, the programmer should not decide what the minimum "usable" width is.  The user may have a use case that is different than anything you can imagine.  The only real limit should probably be: no smaller than the minimum that still allows the user to expand the window back to a more reasonable size.

One could argue that it should not go smaller than the minimum width necessary to display all the toolbar icons; I would counter by saying that this ignores use cases where the user may wish to make it smaller to the point where it's wide enough to still read the code (which they may have written at only 60 characters per line) for a situation where they want to refer to the code but don't need to interact with the toolbar icons, just as a for-example.

This is the minimum size on macOS with a Samsung 49" curved display:

![Screen Shot 2022-07-21 at 1 42 36 PM](https://user-images.githubusercontent.com/222271/180311636-05de2513-ed82-4154-b550-42d661a08e3b.png)

### Operating System Version

macOS 12.4

### Mu Version

1.1.1

### Editor Log
```
2022-07-21 13:44:55,491 - root:269(run) INFO: 

-----------------

Starting Mu 1.1.1
2022-07-21 13:44:55,491 - root:270(run) INFO: uname_result(system='Darwin', node='HaikuSilly.domain', release='21.5.0', version='Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:29 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T8101', machine='x86_64', processor='i386')
[...]

PR #2376, opened 2022-12-25
fix 2375 - this seems to fix the issue I ran into

haikusw:
Reading the documentation, it appears that `unlock()'ing the shared memory is not sufficient and that one should `detach()` it also.  Adding that seems to have resolved this issue #2375 on macOS 10.13.6 with python 3.8.16.

### DIFF
--- a/mu/app.py
+++ b/mu/app.py
@@ -284,6 +284,7 @@ class SharedMemoryMutex(object):
 
     def __exit__(self, *args, **kwargs):
         self._shared_memory.unlock()
+        self._shared_memory.detach()
 
     def acquire(self):
         if self._shared_memory.attach():


tjguk:
Well thank you. Not sure how I missed that rather obvious point! (Nor why it worked in all my tests...). Let me take your PR for a spin then hopefully I can merge in short order.

haikusw:
It appears that some operating systems may detach automatically?  Not 100% sure.  And it doesn't solve the cash of crashing not detaching, obviously.

haikusw:
I'm definitely not at all familiar with Qt nor done very much python, so please be skeptical :)

tjguk:
In short (and as expected) it runs fine on my local machine. I've just enabled the GH checks for your branch -- from experience the RPi containers will take a while to run. But assuming everything runs clear on those platforms I'm happy to merge.

haikusw:
Thanks.  I have to get some sleep (~2am!) but tomorrow I'll see about updating my branch for #2315 from master after you've merged this and then making a PR for it.

tjguk:
@haikusw - All tests cleared on live. 👍 

I had a quick look at the PR you'd recently closed against the #2315 issue. In principle it looks ok, but I'd like to get the view of the other devs as there may be experience from elsewhere which tells against the approach you're suggesting.

haikusw:
@tjguk unfortunately, this fix just disabled the check for already running because I didn't understand how `__exit__` worked.  I apologize.   I came back to see about handling the app crash hole and did more manual testing.

The automated tests didn't catch this.

I've been looking into how to fix this bug #2375 properly and I'm not finding the right way to do it.

As I understand it, the code needs to call `_shared_memory.detach()` on both normal exit and crash paths to avoid this #2375 issue.

Anyway, I apologize that my ignorance of python created a non-fix.  I'll keep looking for the right way to do it; suggestions welcome.

haikusw:
New draft PR #2386 works on this macOS 10.13.6 machine with python 3.8.16, but one test is failing in CI on a bunch of different configurations.

PR #2388, opened 2023-01-06
make: Update Linux Docker recipe to create `.tar` file from `.AppImage`.

carlosperate:
This way the `.AppImage` file will retain its executable rights and users will be able to double click it as soon as it's uncompressed, as it is currenty indicated in the installations instructions from the website.


ntoll:
Thank you..! :-)

PR #2410, opened 2023-03-24
Remove shebang from a non-executable script

hroncok:
This file is not (installed as) executable,
hence the shebang is never used and is redundant.


tjguk:
Thanks (really belatedly)

PR #2416, opened 2023-05-05
Update workflows; github dropped macOS 10-15

haikusw:
See if this gets macOS tests running again


haikusw:
Carl got this taken care of already

PR #2417, opened 2023-05-05
CI: Update deprecated Ubuntu and macOS runners.

carlosperate:
As the CI will be red otherwise.


carlosperate (commented):
All relevant tests to this PR have passed. I'll have to merge this before the rest are done to be able to unblock other contributors, but any failues on the unfinished CI jobs would be unrelated to this change.

Issue #2423, opened 2023-06-27
PyGame mode issues in macOS 10.12 Sierra


carlosperate:

carlosperate:
```
pygame 2.5.0 (SDL 2.28.0, Python 3.8.13)
Hello from the pygame community. https://www.pygame.org/contribute.html
/Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pgzero/runner.py:2: RuntimeWarning: use mixer: dlopen(/Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/mixer.cpython-38-darwin.so, 2): Library not loaded: @loader_path/libmp3lame.0.dylib
  Referenced from: /Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/.dylibs/libsndfile.1.0.34.dylib
  Reason: no suitable image found.  Did find:
	/Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/.dylibs/libmp3lame.0.dylib: cannot load 'libmp3lame.0.dylib' (load command 0x80000034 is unknown)
	/Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/.dylibs/libmp3lame.0.dylib: cannot load 'libmp3lame.0.dylib' (load command 0x80000034 is unknown)
(ImportError: dlopen(/Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/mixer.cpython-38-darwin.so, 2): Library not loaded: @loader_path/libmp3lame.0.dylib
  Referenced from: /Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/.dylibs/libsndfile.1.0.34.dylib
  Reason: no suitable image found.  Did find:
	/Users/carlos/Library/Application Support/mu/mu_venv-38-20230627-100218/lib/python3.8/site-packages/pygame/.dylibs/libmp3lame.0.dylib: cannot load 'libmp3lame.0.dylib' (load command 0x80000034 is unknown)
[...]

PR #2425, opened 2023-06-28
wheel: Temporarily pin pygame max version for user venv.

carlosperate:
As a workaround for issue:
- https://github.com/mu-editor/mu/issues/2423


carlosperate (commented):
Tested in macOs, and @tjguk tested in Windows & linuxs.

PR #2448, opened 2023-07-27
Fix asserts for called once in Python 3.12

hrnciar:


carlosperate:
Thanks for the PR @hrnciar, not sure why the tests didn't run, even if I changed the based. So I've rebased and pushed, so it looks like the CI has been finally triggered.

Once the tests pass we can merge the PR 👍

tests/test_logic.py
```diff
@@ -2524,7 +2524,7 @@ def test_change_mode_workspace_dir_exception():
         ed.change_mode("circuitpython")
         assert mock_error.call_count == 1
     assert ed.mode == "circuitpython"
-    assert python_mode.workspace_dir.called_once()
+    python_mode.workspace_dir.assert_called_once()
```
carlosperate, line 2527: Looks like `assert_called_once()` was introduced in Python 3.6, so it fails in 3.5.
We are about to update dependencies and drop support for all the legacy stuff, but in the meantime I'll update this to use assert_called_once_with() and it should pass tests.

carlosperate, line 2527:
```suggestion
    python_mode.workspace_dir.assert_called_once_with()
```

carlosperate (approved):
Okay, so the only CI failure is a known issue with the PiOS stretch environment, so this is good to be merged 👍

Thanks @hrnciar!

PR #2452, opened 2023-07-28
venv: Use Windows short paths for venv intepreter & pip as well.

carlosperate:
Attempt to fix:
- https://github.com/mu-editor/mu/issues/2451

I've been unable to replicate in my Windows environment, but could ask affected users to check if the installer from this PR might work for them.

### DIFF
--- a/mu/virtual_environment.py
+++ b/mu/virtual_environment.py
@@ -531,11 +531,11 @@ class VirtualEnvironment(object):
         #
         # Pip and the interpreter will be set up when the virtualenv is created
         #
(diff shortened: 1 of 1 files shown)


carlosperate:
@tjguk tagging for visibility in this PR and issue https://github.com/mu-editor/mu/issues/2451.
I'll update the info here in case any of the users are able to test it and confirm if it makes a difference or not.

tjguk:
Going to merge this as it's a fairly safe change and passes all tests

PR #2460, opened 2023-09-09
Reset Mutex (Issue 2429)

tjguk:
When the app hard crashes (eg via a faulty appimage) while the Mutex is held, there is no way to reset it short of writing code to do that or rebooting! This change adds a small change which will reset the Mutex if it's orphaned but will not have any effect if it's still owned by another process (its primary purpose).

It's not easy to test -- least of all automatically. I've only been able to ad hoc test it on Linux -- and that may well be the only platform on which the problem occurs, as on other platforms the shared memory which the Mutex uses is released by the OS; on Linux it has to be released by the app.

My standard approach is to use the appimage binary which crashes consistently on my Linux Mint setup. That then leaves the Mutex held but orphaned. Without this PR, Mu will never start up; with the PR in place Mu will re-own the shared memory and will acquire the Mutex. Conversely, if the Mutex is in fact held by a running Mu instance, with this PR, things happen as they should and the second instance does not run.


ntoll:
@tjguk thank you so much for all your work on this.

PR #2465, opened 2023-10-12
Bundled wheels fix: Flask 2 needs Werkzeug < 3.0.0.

tmontes:
CONTEXT:

* Running Mu on my dev intel macOS for the purposes of packaging + shipping the next release.
* Development environment created from scratch.

FAILURE:

* Bundled wheel process downloads `flask==2.0.3` and non-locked dependencies.
* Brings in `Werkzeug 3.0.0`.
* `import flask` fails with it!!!

FIX:

* Tweak wheel downloader to accept "list of pip things to install".
* For flask, specify both `flask==2.0.3` and `Werkzeug<3.0.0`.

### DIFF
--- a/mu/wheels/__init__.py
+++ b/mu/wheels/__init__.py
@@ -41,9 +41,7 @@ mode_packages = [
     # pygame versions >=2.1.3 have issues in macOS 10.x, so temporarily for
     # Mu release 1.2.1 pin the max version here
     # https://github.com/mu-editor/mu/issues/2423
-    ("pygame", ("pygame<2.1.3",)),
-    ("numpy", ("numpy",)),
-    ("pgzero", ("pgzero>=1.2.1",), "--no-dependencies"),
+    ("pgzero", ("pgzero>=1.2.1", "pygame<2.1.3")),
     # Lock Werkzeug to < 3.0.0: import flask fails, otherwise.
     ("flask", ("flask==2.0.3", "Werkzeug<3.0.0")),
     # The version of ipykernel here should match to the version used by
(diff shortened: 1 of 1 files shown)


mu/wheels/__init__.py
```diff
@@ -41,22 +41,21 @@ class WheelsBuildError(WheelsError):
...
     # https://github.com/mu-editor/mu/issues/2423
-    ("pygame", "pygame<2.1.3"),
-    ("numpy", "numpy"),
-    ("pgzero", "pgzero>=1.2.1", "--no-dependencies"),
-    # Flask v1 depends on Jinja v2, which doesn't have an upper bound limit in
-    # MarkupSafe, and v2.1 is not compatible with Jinja v2
-    ("flask", "flask==2.0.3"),
+    ("pygame", ("pygame<2.1.3",)),
+    ("numpy", ("numpy",)),
+    ("pgzero", ("pgzero>=1.2.1",), "--no-dependencies"),
```
carlosperate, line 46: The pygame entry here was added here to ensure we installed a known version (pgzero installs the latest). Numpy was only added because it's needed as the only other pgzero dependency, since `--no-dependencies` doesn't pull them.

So I believe with this PR we can also change this line to this? (haven't test it):
```suggestion
    ("pgzero", ("pgzero>=1.2.1", "pygame<2.1.3")),
```

tmontes: Thanks @carlosperate, I'll give it a go... :)

tmontes:
NOTE:

* Tested out on macOS, both from source in "pip install -e ." mode and after `pup` packaging.
* Flask/web mode working as expected.

PS:

* WRT pygame zero: also working once packaged.
* Does not work from the development venv -- needs minor fix in coming PR.

THUS:

* Will merge regardless of the two **Run tests / Test PiOS stretch-2018-03-13** check failures -- everything else looking good.

PR #2477, opened 2023-12-08
CI: Update Pi OS Stretch apt repositories URL in sources.list

carlosperate:
As raspberry pi has moved the repository URL to a different location.
It is no longer updated, but will be fine to install whatever is the latest version there of the apt packages required.


carlosperate (commented):
Had to also increase the timeout for Stretch, but all passes now, so it's good to be merged 👍

PR #2486, opened 2023-12-27
Fix test calling `called_once_with()` instead of `assert_called_once_with()`.

carlosperate:
A MagicMock object was being asserted with `called_once_with()` instead of `assert_called_once_with()`, likely an accidental mistake.

As a MagicMock always returns another instance of MagicMock on anything not previoysly set up, the `assert` was always passing, even though we updated the `Window.connect_zoom()` implementation and these asserts should be failing.

I only noticed this because in Python 3.12 this type of typo is caught and throws an exception.


carlosperate (commented):
The CI issue is not related to this PR, so I'll merge this to unblock work on Python 3.12 from other branches.

PR #2493, opened 2024-01-12
CI: Update Pi OS job to the last Buster release (35% faster).

carlosperate:
Which uses a new Qemu emulation configuration (https://github.com/carlosperate/docker-qemu-rpi-os/commit/49c922b8b9e30984767105cd8b9be960732ae095), which is about 35% faster in the CI completion time, from around 30ish minutes to 20ish minutes.

This Buster legacy release from May 2025 is also the last Buster release by Raspberry Pi. As the next OS release in October was based on Bookworm, and so Bullseye became the new "legacy" OS version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,7 +100,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


carlosperate (commented):
All test pass, so we can merge this right away 👍

Issue #2500, opened 2024-01-31
Running tests with Python 3.8 fails


carlosperate:
All CI jobs with Python 3.8+ are now failing, without any changes in this codebase to trigger this.

As an example, the latest commit in main is: https://github.com/mu-editor/mu/commit/5e4067b67a758438fdfa1977314356cd41c817e6

When it was pushed, all tests passed: https://github.com/mu-editor/mu/actions/runs/7504275586/attempts/1
I've restarted the jobs and Python 3.8 fails: https://github.com/mu-editor/mu/actions/runs/7504275586

This job for PR https://github.com/mu-editor/mu/pull/2487, which uses Python 3.7 to 3.11, and only 3.7 passes:
- https://github.com/mu-editor/mu/actions/runs/7504292209
- https://github.com/mu-editor/mu/actions/runs/7730701490

As the commit was pushed on the 12th of January, ~~something in the runners or the Pyhon action must have changed since then~~ (edit: can replicate locally with a clean installation of Mu in a Python 3.8 virtual env):
- https://github.com/actions/runner-images/issues/
- https://github.com/actions/setup-python/issues

carlosperate:
Looks like the issue is with a change introduced in pytest v8.0.0 (Python 3.8+), as setting the max pytest version to v7.4.4 works.

PR #2502, opened 2024-02-04
tests: Fix issue where gettext is not individually installed in some modules.

carlosperate:
Fixes https://github.com/mu-editor/mu/issues/2500
- https://github.com/mu-editor/mu/issues/2500

Also updates the actions/setup-python action from v4 to v5 to remove warnings about depreacted version of node used in v4.


carlosperate (commented):
As this change is relatively small, only affects the tests and fixes the CI I'll merge this right away.

PR #2503, opened 2024-02-10
Added support for CH9102 in ESP mode

sankarcheppali:
Added support for CH9102 USB-Serial chip

https://cdn-learn.adafruit.com/assets/assets/000/111/949/original/CH9102DS1.PDF?1653415979

### DIFF
--- a/mu/modes/esp.py
+++ b/mu/modes/esp.py
@@ -52,6 +52,7 @@ class ESPMode(MicroPythonMode):
         (0x1A86, 0x7523, None, "HL-340"),
         (0x10C4, 0xEA60, None, "CP210x"),
         (0x0403, 0x6001, "M5STACK Inc.", "M5Stack ESP32 device"),
+        (0x1A86, 0x55D4, None, None),  # CH9102
         (0x0403, 0x6001, None, None),  # FT232/FT245 (XinaBox CW01, CW02)
         (0x0403, 0x6010, None, None),  # FT2232C/D/L/HL/Q (ESP-WROVER-KIT)
         (0x0403, 0x6011, None, None),  # FT4232


carlosperate:
Thanks for the PR @sankarcheppali!
Where did the VID and PID come from? Are those the default values for the chip? Looks like they can also be configured with different values.
Could you provide any links showing ESP boards using this chip?

sankarcheppali:
Hi @carlosperate 
Thanks for the response. 
This is the board I have
https://github.com/Xinyuan-LilyGO/LILYGO-T8/blob/main/Schematic/t8_v1.7.1.pdf

I took the VID,PID from the windows device manager.

amaksoft:
Hello! CH2102 is also used by [M5Stamp C3](https://docs.m5stack.com/en/core/stamp_c3#driver-installation) (There's both driver and schematics that confirm this particular chip usage)

Chip info on [manufacturer website](https://www.wch-ic.com/products/CH9102.html)

To confirm chip's VID:PID
From shell:
```
$ lsusb
Bus 002 Device 001: ID 1a86:55d4 QinHeng Electronics
```
In device manufacturer [Linux driver code](https://github.com/WCHSoftGroup/ch343ser_linux/blob/main/driver/ch343.c#L2038):
```c
static const struct usb_device_id ch343_ids[] = {
        // ... other devices
	{ USB_DEVICE(0x1a86, 0x55d4) },			       /* ch9102 chip */
        // .... other devices
	{}
```

Can we please get this PR merged so that we don't have to patch it locally?

Cheppali-Reddy:
Hello @carlosperate , please let me know if there are any blockers for merging this ?

carlosperate (approved):
Thanks everyone! This is good to be merged 👍

PR #2514, opened 2024-06-25
CI: Fix current issues with Python 3.5 pip, arm macOS & pi ssh action

carlosperate:
11 CI jobs are currently failing in the main branch, this PR fixes these issues.


carlosperate (commented):
This is a CI only change and fixes the current failures, so I'll merge it now.

PR #2536, opened 2024-12-11
Fix up dependencies and CI workflows

carlosperate:
To get the CI running again.


carlosperate (commented):
Everything passes except the AppImage build, so merging this now and fixing the required docker image later.

PR #2537, opened 2024-12-11
CI: Update Docker image used for building AppImage.

carlosperate:
As running GH actions inside the Docker image (required to be able to upload the built AppImage to the artifacts) requires node 20, which requires glibc 2.28+.

Ubuntu 18.04 only has 2.27, so the Docker image base image has been moved from the previous Ubuntu 16.04 to Debian Buster.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,14 +54,15 @@ jobs:
   build-linux:
     runs-on: ubuntu-20.04
     container:
-      image: ghcr.io/mu-editor/mu-appimage:2022.05.01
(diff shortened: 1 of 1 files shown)


carlosperate:
I've tried the AppImage produced in Ubuntu 22.04 and seems to work okay.
@ntoll (or anybody else using Linux) would you mind trying the Appimage download in your environment to check things are as expected?
https://github.com/mu-editor/mu/actions/runs/12287791198

carlosperate (commented):
Looks like apart from the newest glibc everything works as expected, so I'll merge this now to get the CI fully green.

